PR #7, opened 2019-03-12
User provided page config

danielcondemarin:
## Motivation

Currently is not possible to change a specific page configuration. For example, you might want to allocate more memory for a specific page function, or  add a greater timeout etc.

## Proposal

Under the plugin custom configuration add a new optional field called `pageConfig`:

```yml
plugins:
  - serverless-nextjs-plugin

custom:
  serverless-nextjs:
    nextConfigDir: ./
    pageConfig:
     home:
      memorySize: 512MB
      timeout: 5
```

With this approach the user can have more granular control over the pages config. The override config. will be merged into the defaults provided by the plugin.

The fields available to use in the overrides are here: https://serverless.com/framework/docs/providers/aws/guide/functions/

Validation might be added to prevent overriding with incorrect values the `handler`, `name`, `runtime`, etc.


danielcondemarin:
Closing for now, waiting until there are more use cases for this

PR #37, opened 2019-04-04
Move react to peer dependencies

bushmango:
This will avoid including react twice with different versions


danielcondemarin:
@bushmango the `integration/` test packages need `react` and `react-dom`, could you add them? thanks 🙏

bushmango:
Added the package.json files -- it looks like the integration tests are failing for other reasons unfortunately

danielcondemarin:
Sorry about the int tests @bushmango. 
Got it fixed on a separate PR https://github.com/danielcondemarin/serverless-nextjs-plugin/pull/41. This makes **only next** a peer dependency. `react` and `react-dom` are next's peerDeps.

PR #66, opened 2019-05-08
feat: global page config

mekwall:
Needed a way to add config to all page functions so added a very simple way of doing so by allowing an asterisk `*` entry in pageConfig that is applied to all functions. The global config can itself be overridden by page-specific config.


mekwall:
I actually had no idea you could configure layers as global. I couldn't find that in the docs, but yeah, you're right about having other functions that I want another config for so this PR is very much relevant for that case 😊

PR #72, opened 2019-05-14
Use pageId as identifier 

Xiphe:
In order to not cause conflicts between pages with same name in different folders.

ref https://github.com/danielcondemarin/serverless-nextjs-plugin/issues/67

We might have a few pageIds with `$` in the future so while I was at it, I added a normalisation for the function name in oder to not break CloudFormation-Ressource names.

PTAL: @danielcondemarin

### DIFF
--- a/classes/NextPage.js
+++ b/classes/NextPage.js
@@ -51,15 +51,19 @@ class NextPage {
   }
 
   get functionName() {
-    if (this.pageName === "_error") {
+    if (this.pageId === "_error") {
       return "notFoundErrorPage";
     }
 
-    return this.pageName + "Page";
+    return (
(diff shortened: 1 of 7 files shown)


Xiphe:
pre-released as `@xiphe/serverless-nextjs-plugin@2.0.0-RC.1`

danielcondemarin:
Thanks for the PR 👍 . 
Can you look at the integration tests? They're checking for the Lambda resources created which changed with this PR.  https://travis-ci.com/danielcondemarin/serverless-nextjs-plugin/jobs/199991482

Xiphe:
thanks for accepting the PR <3

PR #99, opened 2019-07-09
Replacing square brackets when getting page route

vkeskinen:
Next.js 9 introduced file-system based dynamic routes (https://nextjs.org/blog/next-9#dynamic-route-segments). However AWS API Gateway doesn't allow square brackets in resource's path part.

By the way, not sure how generic this is or does it have some implications in other parts of the plugin but at least it seemed to solve our problem. If there are problems I'll be happy to consider some other solution.


edercampelo:
+1

jayarnielsen:
Brilliant! I got next 9 up and running on AWS because of this.

mattdell:
Is there anything left to do with this PR before we can get it merged?

mattdell:
And would this work with nested Dynamic Routes?

e.g.

```
GET /api/blog/[blogId]/comments/[commentId]
```

vkeskinen:
> And would this work with nested Dynamic Routes?
> 
> e.g.
> 
> ```
> GET /api/blog/[blogId]/comments/[commentId]
> ```

We've used the forked version now for a while and it solves the problem in our project so in that sense nothing to add from my side 😄 However our use case is very simple and I haven't tested anything else like the nesting - can do that though if it would be useful.

danielcondemarin:
Sorry it took me a while to look at this. I suspect there is more to do to support dynamic route segments but will merge this in considering it has been working for a few people, thanks for the PR @vkeskinen 🙏

vkeskinen:
No problem and thanks for the great plugin @danielcondemarin! That is my assumption as well; if you need any help let me know.

Issue #103, opened 2019-07-15
Incompatible with serverless-aws-alias


mekwall:
Instead of using serverless stages we instead use the https://github.com/serverless-heaven/serverless-aws-alias plugin. Sadly, the S3 part of serverless-nextjs-plugin doesn't work well together with this plugin.

We have three stages and thus need three different buckets, but serverless-nextjs-plugin only allow one bucket. Deploying with alias `dev` works just fine, but when we try to deploy to `staging` with another bucket serverless-aws-alias will first give the following error because `NextStaticAssetsS3Bucket` resource is listed twice (which isn't allowed, it has to have a unique name per alias):

```
Serverless: Preparing alias ...
Unhandled rejection Error: Resource NextStaticAssetsS3Bucket is already deployed in another alias with a different configuration. Either you change your resource to match the other definition, or you change the logical resource id to deploy your resource separately.
```

The deployment will eventually fail with the following error:

```
The assets bucket "staging-next-assets.playpilot.com" does not exist. Create it manually or delete the stack and recreate it.
```

Manually creating this bucket does't help either. You'll just get the same error.

A quick fix is to disable serverless-nextjs-plugin handling of S3 altogether and we do it manually ourselves in our config, but I haven't found a way to do that for now. Later on it would be great if these incompatibilities could be fixed so these two plugins can play well together. I'd gladly help out with that.

danielcondemarin:
@mekwall How do you solve this problem for other cf `resources` in your stack?

mekwall:
@danielcondemarin I just add the alias as part of their resource names so that they are unique.

mekwall:
With the referenced PR I can now work around this issue by disabling creation of s3 asset bucket and set it up manually in my serverless config like this:

```
const capitalizedAlias = alias[0].toUpperCase() + alias.substr(1);
const bucketResourceName = `NextStaticAssetsS3Bucket${capitalizedAlias}`;

const resources = {
  Resources: {
    [bucketResourceName]: {
      Type: 'AWS::S3::Bucket',
      Properties: {
        AccessControl: 'PublicRead',
        CorsConfiguration: {
          CorsRules: [
            {
              AllowedHeaders: ['*'],
              AllowedMethods: ['GET'],
              AllowedOrigins: ['*'],
              ExposedHeaders: ['Date'],
              Id: 'NextStaticAssetsS3BucketId1',
              MaxAge: '3600',
            },
          ],
        },
[...]

Issue #102, opened 2019-07-13
Missing required key 'Bucket' in params when updating from 2.0.3-alpha.0 to 2.2.0


me4502:
Upon updating a very simple NextJS app (Just a single page called index) from 2.0.3-alpha.0 to 2.2.0, I start getting the following error:

`Missing required key 'Bucket' in params`

Google seems to indicate this is a permissions issue, which is setup correctly here. If I revert to 2.0.3-alpha.0 it works, if I go to 2.2.0 it fails. Is something wrong with the way this plugin is handling bucket permissions?

This branch contains the site that this fails on, in the `paste` folder: https://github.com/EngineHub/enginehub-website/tree/feature/builds

```
$ SLS_DEBUG=* sls deploy -v
Serverless: Load command config
Serverless: Load command config:credentials
Serverless: Load command create
Serverless: Load command install
Serverless: Load command package
Serverless: Load command deploy
Serverless: Load command deploy:function
Serverless: Load command deploy:list
Serverless: Load command deploy:list:functions
Serverless: Load command invoke
Serverless: Load command invoke:local
Serverless: Load command info
Serverless: Load command logs
Serverless: Load command metrics
Serverless: Load command print
Serverless: Load command remove
Serverless: Load command rollback
Serverless: Load command rollback:function
Serverless: Load command slstats
Serverless: Load command plugin
Serverless: Load command plugin
Serverless: Load command plugin:install
Serverless: Load command plugin
Serverless: Load command plugin:uninstall
Serverless: Load command plugin
Serverless: Load command plugin:list
Serverless: Load command plugin
Serverless: Load command plugin:search
Serverless: Load command config
Serverless: Load command config:credentials
Serverless: Load command rollback
Serverless: Load command rollback:function
Serverless: Invoke deploy
Serverless: Invoke package
Serverless: Invoke aws:common:validate
Serverless: Invoke aws:common:cleanupTempDir
Serverless Nextjs: Started building next app ...
Creating an optimized production build ...

Compiled successfully.

Page            Size     Files  Packages
┌ ⚡ /           483 B      187         0
├   /_app       2.04 kB    187         0
├   /_document
└   /_error     7.59 kB    230         0

λ  (Lambda)       page was emitted as a lambda (i.e. getInitialProps)
⚡  (Static File)  page was prerendered as static HTML

Serverless Nextjs: Copying next pages to tmp build folder
Serverless Nextjs: Found 3 next page(s)
[...]

Issue #86, opened 2019-06-10
S3 bucket creation fails


richardwillars:
Hello. When trying to deploy for the first time I get the following error: 
"The assets bucket "foobar-qa" does not exist. Create it manually or delete the stack and recreate it."

That S3 bucket doesn't exist, and I have already deleted the cloudformation stack. 

serverless.yml:
`assetsBucketName: "foobar-${self:custom.name}"`

next.config.js
`assetPrefix: "https://foobar.blabla.com"`
(The S3 bucket has it's own subdomain pointing at it).

Any ideas?

richardwillars:
FYI for anyone else hitting this issue.. to get it working I rolled back to v1.8.1 and then commented out the line "after:aws:deploy:deploy:uploadArtifacts": this.uploadStaticAssets," in index.js, then for subsequent deployments I upgraded to the latest plugin.

mattdell:
This is what I'm trying to do, but automated.

So you've created a cloudfront to sit in front of your S3 bucket, yeah?

danielcondemarin:
@richardwillars Can you post your `create` and `update` CF templates under the `.serverless` directory (when the error happens) ?

Elyx0:
Also note that `static` and `public` have to contain at least a file or it crashed too

mattdell:
> Also note that `static` and `public` have to contain at least a file or it crashed too

That was actually raised after we published it here: https://github.com/danielcondemarin/serverless-nextjs-plugin/pull/92#issuecomment-504954593

@danielcondemarin were you able to update the package?

danielcondemarin:
> > Also note that `static` and `public` have to contain at least a file or it crashed too
> 
> That was actually raised after we published it here: [#92 (comment)](https://github.com/danielcondemarin/serverless-nextjs-plugin/pull/92#issuecomment-504954593)
> 
> @danielcondemarin were you able to update the package?

I did last night actually, https://github.com/danielcondemarin/serverless-nextjs-plugin/releases/tag/v2.0.8-alpha.0

danielcondemarin:
@richardwillars You mind trying again using v2.1.0-alpha.0? This version might have fixed this.

PR #106, opened 2019-07-17
Feature: Add option to disable bucket creation

### LINKED ISSUE #100: Found bucket "undefined" from serverless.yml
I am trying to let the plugin resolve the bucket from my `next.config.js` but instead get the following output:

```
Serverless Nextjs: Found bucket "undefined" from serverless.yml
Serverless Nextjs: Proxying NextJS assets -> https://s3-eu-west-1.amazonaws.com/undefined/_next/{proxy}
Serverless Nextjs: Proxying static files -> https://s3-eu-west-1.amazonaws.com/undefined/static/{proxy}
```

I've followed the documentation by setting a valid S3 url in `assetPrefix`, but the plugin doesn't seem to pick it up.

Edit: Turns out that `const { Bucket } = s3Urls.fromUrl(assetPrefix);` in `parseNextConfiguration.js` return `undefined` because the url was malformed. But I consider this to still be a bug since it's actually trying to use `undefined` as a value.

mekwall:
This PR will add the option `createAssetBucket` with the default value set to `true`. It also fixes some issues where it would fail to deploy if the bucketName was resolved to `null` or `undefined`.

This fixes #100 and potentially #102 and #86 as well. It's also a workaround for #103 since it allows you to manually set up your bucket with unique resource names.


danielcondemarin:
Thanks for implementing, will get it published now 👍

PR #115, opened 2019-07-30
Fix Cloudfront not being created without bucket creation enabled

Deadleg:
Using this config

```
custom:
  serverless-nextjs:
    createAssetBucket: false
    assetsBucketName: my-bucket
    cloudFront: true
```

Will result in the CloudFront distribution to not be created because
`plugin.serverless.service.resources` is undefined. Creation is at
the beginning of the method call ensures that all attempts to `merge`
with it will update that property as expected.


danielcondemarin:
I don't see a lot of value in having Cloudfront enabled without routing the static assets for you. Are you using it this way?

Deadleg:
My use case is setting up the S3 bucket externally (to apply IAM policies etc.) but still using Cloudfront to route to the bucket.

The bigger picture is I'm deploying a private website where I want it to be IP white listed. My approach has been to use WAF with IP rules associated with Cloudfront, and S3 be white listed by a Cloudfront OAI and some IAM roles.

danielcondemarin:
Merging 🚀

PR #117, opened 2019-08-08
Add support for APIs

Deadleg:
There are only a few minor issues with the current implementation to
get this going, namely:

- Allowing POST, PATCH etc. in Cloudfront.
- Checking if the NextJS page is a React component or not.

The change to `compatLayer.js` is needed so that if a request without
a body is sent to an API the Stream still has some content. Without
a `_read() is not implemented` is raised which generates an HTTP 400
response without a body.

The stream test case is removed since the stream already has EOF pushed
to it. Don't think there is a use case where the input is a stream
itself.

danielcondemarin/serverless-nextjs-plugin#116


mattdell:
Will this work for Dynamic API routes?

For example...

```
GET /api/blog/[blogId]/comments/[commentId]
```

Deadleg:
Good question, not in this change but I can't see why it wouldn't after https://github.com/danielcondemarin/serverless-nextjs-plugin/pull/99 is merged. I'll test with that pull request in the next couple of days since I've only used it with static API paths so far.

danielcondemarin:
Looks great! Thanks 🙏

PR #120, opened 2019-08-15
feat: add offline asset support

RichiCoder1:
Adds offline static support by adding a simple handler that acts a file proxy for the specific routes.

This is somewhat rough draft, but it was working well with my proof of concept app and the integration app.

Possible further changes could be moving all the logic into serveFile and consolidating the handler down to just something like `_assets.js`.

### DIFF
--- a/integration/__tests__/local-deploy.test.js
+++ b/integration/__tests__/local-deploy.test.js
@@ -1,4 +1,5 @@
 const execSync = require("child_process").execSync;
+const fse = require("fs-extra");
 const path = require("path");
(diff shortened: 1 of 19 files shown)


RichiCoder1:
Ripped out `/public` support as that actually requires some non-trivial effort to specially route and it's experimental and unsupported currently.
Also added integration tests.

danielcondemarin:
@RichiCoder1 Thanks for the PR! However, I'm not happy to add more complexity to the plugin for local development features / serverless-offline support. The direction I'm inclined to follow is for users to rely on tooling provided by the next framework, like `next dev ` for local development, and use this project for the deployment.

Issue #124, opened 2019-08-19
Error: Cannot find module 'next-aws-lambda'


barrysteyn:
# Description
When I use new version of the plugin (v 2.3.2), I get the following error when I try run on AWS (this is happening in my custom lambda)

```Error: Cannot find module 'next-aws-lambda'```

Note: I also installed `next-aws-lambda`. This is strange, because it runs perfectly locally.

**Comment by barrysteyn** [comment]
Also, my code is not the root folder... rather, it's a client folder. I have put my serverless.yml file in that (i.e. the client) folder.

**Comment by me4502** [comment]
This is due to the plugin looking for next-aws-lambda in the same folder when copying it. I noted that in another issue, and the official response was https://github.com/danielcondemarin/serverless-nextjs-plugin/issues/75#issuecomment-493856802

As forcing it to install in the leaf doesn't work reliably, I've been patching the plugin to just include the `../` in the path. Realistically this project should be updated to resolve the actual module and grab it from there, rather than just assuming it's in the same location.

**Comment by barrysteyn** [comment]
Thanks for the reply. Can you do me a favor and share some code of your
patch?
-- 

Barry Steyn
Blog <http://doctrina.org> LinkedIn <http://ca.linkedin.com/in/barrysteyn>
GitHub <https://github.com/barrysteyn>

**Comment by me4502** [comment]
My patch is just
```
-    path.posix.join("node_modules/next-aws-lambda", "**")
+    path.posix.join("../node_modules/next-aws-lambda", "**")
```
in node_modules/serverless-nextjs-plugin/lib/build.js

barrysteyn:
Would you happen to know why it works on my local deploy
(serverless-offline).

Thanks for this. I'll try it and report back here.


-- 

Barry Steyn
Blog <http://doctrina.org> LinkedIn <http://ca.linkedin.com/in/barrysteyn>
GitHub <https://github.com/barrysteyn>

barrysteyn:
Hi, yes this worked. @danielcondemarin, this definitely seems like a bug

barrysteyn:
@me4502 I submitted a PR to fix this issue. See #127. @me4502 can you do me a favor and test out my PR in your code.

franko1260:
Has/can this be extended so other packages can be included as well? I'm keen on using serverless-http instead of next-aws-lambda. I get a "Cannot find module 'serverless-http'" error when using v2.3.4

PR #118, opened 2019-08-14
Exclude unnecessary package files

jayarnielsen:
This closes #62. The test filename was upload.test..js. For some reason the extra "." was causing it to get included in the build.

Additionally I changed the include statement for the next-aws-lambda package to only include ".js" files. Thereby excluding all of the tests, license, etc.

PR #128, opened 2019-08-22
Fix: allow 'next-aws-lambda' to be found using require.resolve

### LINKED ISSUE #75: Deployments in a monorepo setup include entire path in s3 bucket
I've got a NextJS project in a monorepo setup, and when being deployed with this plugin it places the assets in the full path from the root of the git repo/yarn workspaces setup (Including the parent folder?).

Due to this, when visiting the page it cannot find the bundle. It's looking in S3/_next/static, whereas it is actually in S3/_next/parent folder/apps/project name/.next/static

Is there a way to resolve this with config options or is this a bug?

Thanks

barrysteyn:
This solution is more generalized than the previous one. It removes `findup-sync` and instead relies upon require.resolve. This fixes #75 and #124

### DIFF
--- a/packages/serverless-nextjs-plugin/lib/__tests__/build.test.js
+++ b/packages/serverless-nextjs-plugin/lib/__tests__/build.test.js
@@ -11,7 +11,6 @@ const PluginBuildDir = require("../../classes/PluginBuildDir");
 const getNextPagesFromBuildDir = require("../getNextPagesFromBuildDir");
 const NextPage = require("../../classes/NextPage");
 const ServerlessPluginBuilder = require("../../utils/test/ServerlessPluginBuilder");
-const findup = require("findup-sync");
 
 jest.mock("fs-extra");
 jest.mock("next/dist/build");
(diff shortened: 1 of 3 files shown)


barrysteyn:
@me4502 can you test this solution please and give the thumbs up. The sooner the better so Daniel can just merge.

me4502:
Yeah this appears to be working for me in all of the situations I've tested, thanks :)

barrysteyn:
Great. @danielcondemarin this is good to merge, thanks.

danielcondemarin:
Thanks folks 🙏

barrysteyn:
@jayarnielsen Can you check that this change still works for #118. I expect that it should, but nice to get your eyes no this as well.

barrysteyn:
@danielcondemarin As soon as you publish this next version, I can get my CI/CD working again ;)

danielcondemarin:
> @danielcondemarin As soon as you publish this next version, I can get my CI/CD working again ;)

Published now! v2.3.3 🎉

PR #131, opened 2019-08-27
next-aws-lambda can now be consumed using async signature

danielcondemarin:
`next-aws-lambda` now supports async promises:

```js
// using async promise

module.exports.render = async (event, context, callback) => {
  const responsePromise = compat(page)(event, context); // don't pass the callback parameter
  return responsePromise;
};
```

Also updated usage in the plugin to use this approach.

Closes https://github.com/danielcondemarin/serverless-nextjs-plugin/issues/125


barrysteyn:
Would this allow us to control header (like cache-control)?

PR #135, opened 2019-09-04
Fix api gateway methods with ANY

Deadleg:
Fixes danielcondemarin/serverless-nextjs-plugin#134 danielcondemarin/serverless-nextjs-plugin#129

### DIFF
--- a/packages/serverless-nextjs-plugin/classes/NextPage.js
+++ b/packages/serverless-nextjs-plugin/classes/NextPage.js
@@ -6,6 +6,7 @@ const PluginBuildDir = require("./PluginBuildDir");
 
 class NextPage {
   constructor(pagePath, { serverlessFunctionOverrides, routes } = {}) {
+    this.isApi = pagePath.indexOf(path.join("sls-next-build", "api")) === 0;
     this.pagePath = pagePath;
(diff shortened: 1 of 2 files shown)


packages/serverless-nextjs-plugin/classes/NextPage.js
```diff
@@ -6,6 +6,7 @@ const PluginBuildDir = require("./PluginBuildDir");
 
 class NextPage {
   constructor(pagePath, { serverlessFunctionOverrides, routes } = {}) {
+    this.isApi = pagePath.indexOf("sls-next-build/api") === 0;
```
danielcondemarin:
```suggestion
    this.isApi = pagePath.indexOf(path.join("sls-next-build", "api")) === 0;
```

danielcondemarin: I believe this is breaking the int. tests because pagePath in Windows would be `sls-next-build\\api`

Deadleg: Cheers, easy to forget about windows as a linux developer.

danielcondemarin:
Thanks!

Issue #40, opened 2019-04-05
Use Cache-Control: immutable for files in /_next/static


aheimlich:
Next.js's built in server adds a `Cache-Control: immutable` HTTP header to certain files within `/_next/static`. It would be cool if this plugin could do that as well. It would also be nice to have a config option that would apply `Cache-Control: immutable` to other files/directories within `/_next/static`.

Relevant code:

- [Cache-Control header](https://github.com/zeit/next.js/blob/ed1207f78a94cdeab47ff6ed530c4d532f26acb9/packages/next-server/server/next-server.ts#L153)
- [Which files get the header](https://github.com/zeit/next.js/blob/ed1207f78a94cdeab47ff6ed530c4d532f26acb9/packages/next-server/server/next-server.ts#L165)

danielcondemarin:
@aheimlich This makes perfect sense! I would happily accept a PR for the first part (the next cached paths). 
As for the config option I'd prefer to leave that out for now and keep the plugin options to a minimum, and get the default behaviour first closer to next's built in server.

danielcondemarin:
Closed by https://github.com/danielcondemarin/serverless-next.js/pull/144 🚀

PR #144, opened 2019-09-16
Adding cache control to build assets

barrysteyn:
Adds cache-control metadata to build assets. Infrastructure for adding cache-control for public and static folders can be used, however we can chat about how to implement configs for this in the serverless.yml config. This should address #40

### DIFF
--- a/packages/serverless-nextjs-plugin/lib/__tests__/uploadStaticAssets.test.js
+++ b/packages/serverless-nextjs-plugin/lib/__tests__/uploadStaticAssets.test.js
@@ -1,4 +1,5 @@
 const path = require("path");
+const fs = require("fs");
 const fse = require("fs-extra");
 const { when } = require("jest-when");
 const uploadStaticAssets = require("../uploadStaticAssets");
@@ -10,6 +11,7 @@ const uploadDirToS3Factory = require("../../utils/s3/upload");
 jest.mock("fs-extra");
 jest.mock("../../utils/s3/upload");
 jest.mock("../parseNextConfiguration");
+jest.mock("fs");
 
 describe("uploadStaticAssets", () => {
   let uploadDirToS3;
@@ -17,6 +19,7 @@ describe("uploadStaticAssets", () => {
   beforeEach(() => {
     uploadDirToS3 = jest.fn().mockResolvedValue();
     uploadDirToS3Factory.mockReturnValue(uploadDirToS3);
+    fs.readFileSync.mockResolvedValue("1hCeVQzuD6WJQAxuV3hwc");
   });
 
   it("does NOT upload build assets when there isn't a bucket available", () => {
--- a/packages/serverless-nextjs-plugin/lib/uploadStaticAssets.js
+++ b/packages/serverless-nextjs-plugin/lib/uploadStaticAssets.js
@@ -1,11 +1,15 @@
+const fs = require("fs");
 const fse = require("fs-extra");
 const path = require("path");
 const uploadDirToS3Factory = require("../utils/s3/upload");
 
 module.exports = async function() {
-  const uploadDirToS3 = uploadDirToS3Factory(this.providerRequest);
-
   let { nextConfiguration, staticAssetsBucket } = this.configuration;
+  const buildId = nextConfiguration.distDir // eslint-disable-next-line prettier/prettier
+    ? fs.readFileSync(path.join(this.nextConfigDir, nextConfiguration.distDir, "BUILD_ID"))
+    : null;
+
+  const uploadDirToS3 = uploadDirToS3Factory(this.providerRequest, buildId);
 
   const [bucketNameFromConfig, uploadBuildAssets] = this.getPluginConfigValues(
     "assetsBucketName",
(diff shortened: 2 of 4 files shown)


barrysteyn:
@danielcondemarin if you approve of going down this route, this is what I suggest:

 * Have a Cache headers config
 * Inside above config, you can set the headers for build assets, static assets, and public assets
 * Have the ability to apply this to a stage, meaning that for dev purposes, all caching can be turned off, while for prod, it can be turned on

What do you think?

danielcondemarin:
> @danielcondemarin if you approve of going down this route, this is what I suggest:
> 
> * Have a Cache headers config
> * Inside above config, you can set the headers for build assets, static assets, and public assets
> * Have the ability to apply this to a stage, meaning that for dev purposes, all caching can be turned off, while for prod, it can be turned on
> 
> What do you think?

@barrysteyn Thanks for the PR!

Just a few comments. I'd prefer if we sticked to next's behaviour. That is, *automatically* add the `'Cache-Control': 'public, max-age=31536000, immutable'` to:
- `chunk/*` files
- Anything inside the build folder (`build_id/*`) (build_id is in the `BUILD_ID` file)
- CLIENT_STATIC_FILES_RUNTIME folder (`runtime/*`)

as per https://github.com/zeit/next.js/blob/ed1207f78a94cdeab47ff6ed530c4d532f26acb9/packages/next-server/server/next-server.ts#L166.

I don't see much value in making it configurable as the plugin shouldn't be used for development due the lack of serverless offline support. It is recommended to use `next dev` outside of the plugin for any development. 

Also, the new component version will need the exact same change, so we can reuse what is done here 🙌

barrysteyn:
Hi, sorry it has taken so long to answer you back. I will try update this PR with your suggestions over the weekend

barrysteyn:
Hi @danielcondemarin 

I know how I want to implement this, but I need two pieces of information (help) from you:

 1. `BUILD_ID` - from looking at the next code, it seems like this file should be relative to `./next` directory, however I can't find it on my local machine.
 2. Running this locally - I know you have given instructions on how to run this locally, however I just can't get this to work locally. Can you perhaps give me an example `serverless.yml` file so I can try test things locally.

I am going to start using the new component in my new project, and caching these assets have become an absolute must for my team (It's become a blocker before releasing to prod). So I am quite keen to move on to getting this working for the new component.

danielcondemarin:
> Hi @danielcondemarin
> 
> I know how I want to implement this, but I need two pieces of information (help) from you:
> 
> 1. `BUILD_ID` - from looking at the next code, it seems like this file should be relative to `./next` directory, however I can't find it on my local machine.
> 2. Running this locally - I know you have given instructions on how to run this locally, however I just can't get this to work locally. Can you perhaps give me an example `serverless.yml` file so I can try test things locally.
> 
> I am going to start using the new component in my new project, and caching these assets have become an absolute must for my team (It's become a blocker before releasing to prod). So I am quite keen to move on to getting this working for the new component.

1. `BUILD_ID` will be in `.next/BUILD_ID`, for example:

image

2. To run locally you can follow the same approach the [dynamodb example](https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-nextjs-component/examples/dynamodb-crud/serverless.yml) follows. That is you can point at your local copy of the next component:

```yaml
nextApp:
  component: "../../" # this needs to point at a folder with a file called serverless.js
 ```

That makes sure any changes you make in the next component are picked up when deploying.

barrysteyn:
@danielcondemarin Regarding running locally, I mean for the plugin, not for the component. I need to get this working first for the old plugin, and I am struggling to run this locally

danielcondemarin:
> @danielcondemarin Regarding running locally, I mean for the plugin, not for the component. I need to get this working first for the old plugin, and I am struggling to run this locally

So for running locally the plugin I would recommend you create a special folder called `.serverless_plugins` in the app. Inside put a file called `serverless-nextjs-plugin.js` which exports the plugin index. For example:

image

This is documented here https://serverless.com/framework/docs/providers/aws/guide/plugins#service-local-plugin

barrysteyn:
Last question: My team mate tells me that in latest next.js, they are not putting the build id in the build file anymore. Your test above, what version of next are you running?

danielcondemarin:
> Last question: My team mate tells me that in latest next.js, they are not putting the build id in the build file anymore. Your test above, what version of next are you running?

I haven't heard about this change. I've tested it on next 9.0.7 and I still get a BUILD_ID

barrysteyn:
Okay. thanks.

barrysteyn:
@danielcondemarin This PR is now ready for you to review and merge. I have tested this locally with my local project and it is doing what you specified it to do.

barrysteyn:
@danielcondemarin Thanks for merging and publishing. Question: Can this code be reused for the component?

danielcondemarin:
> @danielcondemarin Thanks for merging and publishing. Question: Can this code be reused for the component?

For the most part yes, but first the [aws-s3 component](https://github.com/serverless-components/aws-s3) will need to support passing Cache Control headers.

barrysteyn:
Oh, so that means altering that component. I can look into it, but I don't have time this week. Seems strange that such a vital thing is not supported yes. Is this the official S3 component?

PR #151, opened 2019-09-22
Regex error

k2w2yut:
I tried to use the component and found an error at this line.
Fixing the regex pattern help build complete but please verify the function validity

### DIFF
--- a/packages/serverless-nextjs-component/lib/expressifyDynamicRoute.js
+++ b/packages/serverless-nextjs-component/lib/expressifyDynamicRoute.js
@@ -1,4 +1,4 @@
 // converts a nextjs dynamic route /[param]/ to express style /:param/
 module.exports = dynamicRoute => {
-  return dynamicRoute.replace(/\[(?<param>.*?)]/g, ":$<param>");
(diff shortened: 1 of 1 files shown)


danielcondemarin:
@k2w2yut Can you expand on which error you found? 

The regex seems to be work fine as is. The job of the regex is to convert next bracketed dynamic routes to express ones (using `:`). 

For example:

image

With your change it breaks:

image

k2w2yut:
image

It just simply syntax error when I try to deploy the code.
Disclaimer: I'm not a front-end dev, only taking care of bringing project up to AWS environment.

danielcondemarin:
@k2w2yut You will need at least node 10.x.x for this to work. Otherwise it might be available in earlier versions by using `--harmony-regexp-named-captures`. See https://github.com/nodejs/node/issues/19760

k2w2yut:
I see thank you to point that out

> @k2w2yut You will need at least node 10.x.x for this to work. Otherwise it might be available in earlier versions by using `--harmony-regexp-named-captures`. See [nodejs/node#19760](https://github.com/nodejs/node/issues/19760)

I see, thank you for point that out

PR #152, opened 2019-09-22
Add configuration for bucket name

### LINKED ISSUE #145: Component: specify bucket name
Hey!

Great work with the serverless components version of the plugin, it's quite frankly excellent 👏 

What would be even better is if there was an input option for `bucketName` or something similar to avoid having the serverless framework generate a random S3 bucket name. Of course this is an optional nice-to-have, but it's would be useful to know when looking at the S3 dashboard.

josephluck:
Closes #145 by exposing a configuration option for the S3 bucket name the component creates.

Looks like the input isn't documented in the `@serverless/aws-s3` component, however the input configuration for `name` is used [here](https://github.com/serverless-components/aws-s3/blob/master/serverless.js#L27). If the bucket name isn't specified, the component generates a bucket name as it does currently.

### DIFF
--- a/packages/serverless-nextjs-component/README.md
+++ b/packages/serverless-nextjs-component/README.md
@@ -95,6 +95,19 @@ myNextApplication:
     domain: ["www", "example.com"] # [ sub-domain, domain ]
 ```
 
+### Custom bucket name
+
+To override the default generated bucket name provided by Serverless, configure your `bucketName` like the example below:
+
+```yml
+# serverless.yml
+
+myNextApplication:
+  component: serverless-next.js
+  inputs:
+    bucketName: my-next-application
+```
+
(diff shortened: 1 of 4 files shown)


danielcondemarin:
@josephluck Thanks for the PR! And providing great detail on the other issues you've raised 🙌 

You mind having a look at the broken unit tests? I'm guessing they're breaking because the output of the component changed.   
Thanks!

josephluck:
> You mind having a look at the broken unit tests?

Done here: https://github.com/danielcondemarin/serverless-next.js/pull/152/commits/bb5ba653e81ebc5a711fc67c0cbde8b67acdafc6

danielcondemarin:
Merged 🚀

PR #156, opened 2019-09-22
Fixes response status code not being set properly

### LINKED ISSUE #155: Component: Bad routes returning 200 status
I'm upgrading to the new serverless component which is working great, except I noticed that invalid paths return a 200 status code instead of 404. It still shows the _error.js page but the HTTP response from CloudFront is 200. Any idea?

image

zachabney:
Resolves #155 by using the `response.status` value instead of `response.statusCode`.

Previously, setting the value of `response.statusCode` didn't have an effect since CloudFront uses `response.status`. This meant that all routes returned a 200 status code even if the route did not exist.

It looks like `response.statusCode` might have been a typo since `statusCode` isn't an actual property on `response`.

### DIFF
--- a/packages/serverless-nextjs-component/next-aws-cloudfront.js
+++ b/packages/serverless-nextjs-component/next-aws-cloudfront.js
@@ -104,10 +104,10 @@ const handler = event => {
 
   Object.defineProperty(res, "statusCode", {
     get() {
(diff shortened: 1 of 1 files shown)


danielcondemarin:
Thanks @zachabney ! Merged 🚀

Issue #154, opened 2019-09-22
Components: cookies are not passed through to edge lambda


josephluck:
It looks like cookies from the request are not passed through from CloudFront to the underlying edge lambda. After deploying a next component, the CloudFront behaviour for the edge lambda looks like this:

![image](https://user-images.githubusercontent.com/11628330/65387297-ecf91080-dd3d-11e9-9b66-a0f84d3da1e8.png)

This is a problem for apps that rely on cookies for things like authentication etc, as the cookie header is not available in `getInitialProps`.

Perhaps it would be good to have some configuration options to specify the CloudFront behaviour , though it seems like a sensible default for next apps to have this configuration turned on.

It looks as though the behaviour for cookie forwarding is not configurable by the underlying [`@serverless/aws-cloudfront` component](https://github.com/serverless-components/aws-cloudfront/blob/master/lib/getDefaultCacheBehavior.js#L11). 

There's a [PR open](https://github.com/serverless-components/aws-cloudfront/pull/8) that will enable this configuration in the underlying cloudfront component, that can be utilised by this component. If it's merged and deployed, this issue can probably be closed as the default behaviour for forwarding cookies has been changed to forward by default.

josephluck:
The PR that adds support for cookie forwarding has been merged, so I'll check whether this just works out-of-the-box or whether any additional changes are required soon 👍

danielcondemarin:
> The PR that adds support for cookie forwarding has been merged, so I'll check whether this just works out-of-the-box or whether any additional changes are required soon 👍

From looking at this line https://github.com/serverless-components/aws-cloudfront/blob/master/lib/cacheBahaviorUtils.js#L2 it seems like we will need additional changes in the component to pass in `cookies: "all"` to the cloudfront forward config.

danielcondemarin:
@josephluck This is now published on @1.2.3 🎉

josephluck:
> > The PR that adds support for cookie forwarding has been merged, so I'll check whether this just works out-of-the-box or whether any additional changes are required soon 👍
> 
> From looking at this line https://github.com/serverless-components/aws-cloudfront/blob/master/lib/cacheBahaviorUtils.js#L2 it seems like we will need additional changes in the component to pass in `cookies: "all"` to the cloudfront forward config.

[...]

PR #159, opened 2019-09-23
Update CloudFront component to latest

josephluck:
Upgrading to the latest CloudFront plugin allows for cookie forwarding and other options.

It may be worth considering exposing some inputs (with sensible defaults) to further configure CloudFront, but for now this will address the issue described in #154.

### DIFF
--- a/packages/serverless-nextjs-component/package.json
+++ b/packages/serverless-nextjs-component/package.json
@@ -22,7 +22,7 @@
     "node": ">=10.11.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


danielcondemarin:
Thanks for looking after this @josephluck ! Merging 🚀

josephluck:
> Thanks for looking after this @josephluck ! Merging 🚀

Sorry, just saw your comment on the other issue regarding having to make some updates since cookies aren't forwarded by default, another PR coming 👍

PR #160, opened 2019-09-23
Forward cookies and query strings to edge lambda

### LINKED ISSUE #154: Components: cookies are not passed through to edge lambda

josephluck:
This PR adds forwarding for cookies and query strings since aren't forwarded in the CloudFront component by default. Fixes #154.

### DIFF
--- a/packages/serverless-nextjs-component/__tests__/build.test.js
+++ b/packages/serverless-nextjs-component/__tests__/build.test.js
@@ -237,7 +237,7 @@ describe("build tests", () => {
         defaults: {
           allowedHttpMethods: expect.any(Array),
           queryString: true,
-          cookies: 'all',
+          cookies: "all",
           ttl: 5,
           "lambda@edge": {
             "origin-request":
--- a/packages/serverless-nextjs-component/__tests__/build.test.js
+++ b/packages/serverless-nextjs-component/__tests__/build.test.js
@@ -236,6 +236,8 @@ describe("build tests", () => {
       expect(mockCloudFront).toBeCalledWith({
         defaults: {
           allowedHttpMethods: expect.any(Array),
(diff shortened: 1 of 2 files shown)


packages/serverless-nextjs-component/__tests__/build.test.js
```diff
@@ -236,6 +236,8 @@ describe("build tests", () => {
       expect(mockCloudFront).toBeCalledWith({
         defaults: {
           allowedHttpMethods: expect.any(Array),
+          queryString: true,
+          cookies: 'all',
```
danielcondemarin:
```suggestion
          cookies: "all",
```
The single quotes are breaking the linter :-)

danielcondemarin:
Thanks once again @josephluck 🙌 Merging 🚀

PR #173, opened 2019-09-29
Update aws-s3 component

Gabrola:
Update aws-s3 component since any versions < 4.0.0 have a problem uploading directories on windows.

No breaking changes between 3.1.0 and 4.0.0


danielcondemarin:
Thanks! Merging 🚀

Issue #177, opened 2019-10-01
How to support multiple environments


justinwhall:
Hello, is there a way to support multiple environments with the new serverless component? We've upgraded and used to pass env vars to serverless.yml which no longer seems to be an option.

danielcondemarin:
@justinwhall Would this https://github.com/serverless/components#environment-variables help?

justinwhall:
I'm not sure. I'm familiar with passing env vars to the serverless.yml but I'm not certain where they would go with the update?

 Components are a new concept to me. Perhaps I'm missing something here but as far as the `serverless.yml` config goes, I'm not certain how both generate or conditionally deploy to different components – or perhaps that wrong terminology. Same component but different Apps.

What I'm after here is a way to build/deploy different environments something like `sls deploy --env=stage`

Does that make sense?

justinwhall:
I suppose [this](https://github.com/serverless/components/issues/481#issuecomment-527848024) is what I'm looking for. Does this package support something like:

```
myApp:
  component: serverless-next.js
  inputs:
    code: ../
```
With a directory structure like:
```
code
code/prod
code/test
```

skycremaschinin:
Hi,

how could i manage different stages, specifying different cloudfront distributions, buckets and lamba for each stages?

danielcondemarin:
Hi folks, I've just released `@1.3.0` which adds support for custom app directory paths (see [here](https://github.com/danielcondemarin/serverless-next.js#inputs) `nextConfigDir`). Credits to @jayarnielsen for implementing 🙌 

I'm closing the issue as you should now be able to do something like:

```
myApp
│   next.config.js    
│
└───pages
│   │   index.js
│   │   users.js
│   │   ...
│   
└───dev
    │   serverless.yml
└───prod
    │   serverless.yml
``` 

```yml
# dev/serverless.yml
myNextApp:
 inputs:
  domain: [ "dev", "myapp.com" ]
  nextConfigDir: "../"
```

Issue #166, opened 2019-09-26
Using Bitbucket Pipelines to deploy to AWS


shoremark:
# What I am trying to do
I would like to deploy serverless-next.js to AWS using bitbucket pipelines. I am using the component version of Serverless Next.js. I know there is a current bug #147 with using CI for deploying that requires you to store the `.serverless` the folder in the repo to get around this for multiple environments. I have created multiple versions of the folder for each environment such as `.serverless-testing`. When the CI runs I rename the folder to `.serverless` to use a specific deployment with an account-specific Key and Secret. This process works locally but in the CI I am getting an error that is hard to debug due to it gives very few specifics about the issue. Is there any way to get more information on what is causing the error below or any ideas what would cause this type of issue?

## Code used to deploy
```
- npm update -g serverless
- yarn install
- mv .serverless-testing .serverless
- export AWS_ACCESS_KEY_ID=$AWS_KEY_DEV
- export AWS_SECRET_ACCESS_KEY=$AWS_SECRET_DEV
- export AWS_REGION=us-east-1
- npx serverless --debug
```

## Error message in Pipelines
```
aNA
  36s › XXXXXXXXX › Deploying .
aNA  DEBUG ─ Deploying bucket XXXXXXXXX in region us-east-1.
  DEBUG ─ Checking if bucket XXXXXXXXX exists.
  36s › XXXXXXXXX › Deploying ..
aNA
  36s › XXXXXXXXX › Deploying ..
aNA
  error:
  { BadRequest: null
    at Request.extractError (/root/.serverless/components/registry/npm/serverless-next.js@1.2.3/node_modules/aws-sdk/lib/services/s3.js:565:35)
    at Request.callListeners (/root/.serverless/components/registry/npm/serverless-next.js@1.2.3/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/root/.serverless/components/registry/npm/serverless-next.js@1.2.3/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/root/.serverless/components/registry/npm/serverless-next.js@1.2.3/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/root/.serverless/components/registry/npm/serverless-next.js@1.2.3/node_modules/aws-sdk/lib/request.js:22:10)
[...]

PR #178, opened 2019-10-02
Multiple environment support

jayarnielsen:
This allows for multiple environments by introducing a new "code root" input. Based largely on the example provided here: https://github.com/serverless/components/issues/481#issuecomment-527848024

Essentially a project can have multiple serverless.yml files in sub folders that reference the root directory like so:

```
myNextApp:
  component: serverless-next.js
  inputs:
    code:
      root: ../
```

I left the structure like this in part to mimic what is done in [serverless/backend](https://github.com/serverless-components/backend/blob/master/serverless.js#L25) and also to potentially allow for [custom build directories](https://nextjs.org/docs#setting-a-custom-build-directory). Though for simplicity it could be renamed to inputs->code or inputs ->root to avoid the unnecessary nesting.

This solves #177 and potentially solves or at least sheds some light on #166.

This did break a bunch of tests, but I wanted to get your eyes on this @danielcondemarin. If you're happy with this approach I'll go ahead and fix the tests.

### DIFF
--- a/packages/serverless-nextjs-component/__tests__/assets.test.js
+++ b/packages/serverless-nextjs-component/__tests__/assets.test.js
@@ -42,21 +42,21 @@ describe("Assets Tests", () => {
 
     it("uploads client build assets", () => {
       expect(mockS3Upload).toBeCalledWith({
-        dir: "./.next/static",
+        dir: path.join(fixturePath, ".next/static"),
         keyPrefix: "_next/static"
       });
     });
 
     it("uploads user static directory", () => {
       expect(mockS3Upload).toBeCalledWith({
-        dir: "./static",
+        dir: path.join(fixturePath, "static"),
         keyPrefix: "static"
       });
     });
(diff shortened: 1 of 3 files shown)


danielcondemarin:
@jayarnielsen Thanks for the PR 🙏 

I think this is a good idea. Calling it `code` or `root` might be a bit too generic though. I would prefer if it was named `nextConfigDir`, i.e.

```yml
myNextApp:
  component: serverless-next.js
  inputs:
    nextConfigDir: /path/to/next.config.js
```

By default it should be `.`. This is already an option on the [next plugin](https://github.com/danielcondemarin/serverless-next.js/tree/master/packages/serverless-nextjs-plugin#all-plugin-configuration-options).

What do you think?

jayarnielsen:
@danielcondemarin I've updated this to use `nextConfigDir` and fixed the broken tests.

PR #187, opened 2019-10-07
split build and deploy phase

skycremaschinin:
As you suggested, i added `build` input and splitted the `build`and `deploy` phases.
hope it can be useful.


danielcondemarin:
@skycremaschinin This is looking great! Thank you. Just it seems there are a few linting errors, you mind having a look at those?

skycremaschinin:
@danielcondemarin just fixed.

PR #202, opened 2019-10-13
Fix cloudfront default cache behavior

macs91:
This should fix query and cookies for cloudfront default cache behaviour according to `@serverless/aws-cloudfront` [docs](https://github.com/serverless-components/aws-cloudfront#3-configure)

I need some instructions to run and test it but it should work


macs91:
I copy the component folder over my project to test it and it works, query string and cookies are correctly set

danielcondemarin:
Lgtm! Thanks @macs91 🙌

PR #219, opened 2019-10-26
example of multiple instances (stages)

barrysteyn:
A proper solution for #177. This is the preferred way of doing things (I have had lots of back and forth with both the serverless core engineers and collaborators - see https://github.com/serverless/components/issues/481)

@danielcondemarin One thing that I do not have is any domain in Route53. Could you please test this with any domain you have lying around.

### DIFF
--- a/packages/serverless-nextjs-component/examples/multiple-instance-environment/package.json
+++ b/packages/serverless-nextjs-component/examples/multiple-instance-environment/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "multiple-instance-environment",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "deployStaging": "serverless --stage=staging",
+    "deployProd": "serverless --stage=prod"
+  },
+  "dependencies": {
+    "next": "9.1.1",
+    "react": "16.11.0",
+    "react-dom": "16.11.0"
+  },
+  "devDependencies": {
+    "dotenv": "^8.1.0",
+    "@serverless/core": "^1.1.2",
+    "@serverless/template": "^1.1.3",
+    "serverless": "^1.51.0"
+  }
+}
--- a/packages/serverless-nextjs-component/examples/multiple-instance-environment/serverless.js
+++ b/packages/serverless-nextjs-component/examples/multiple-instance-environment/serverless.js
@@ -1,0 +1,35 @@
+const { Component } = require("@serverless/core");
+
+/**
+ * Will deploy the serverless component
+ */
+class Deploy extends Component {
+  async default(inputs = {}) {
+    const { stage } = inputs;
+    if (stage === "staging" || stage === "prod") {
+      // Will load env-${stage} as environment variables
+      require("dotenv").config({ path: `${__dirname}/env-${stage}` });
+
+      const template = await this.load("@serverless/template", stage);
+      const output = await template({ template: "serverless.yml" });
+      return output;
+    } else {
+      this.context.log(
+        "No environment defined... Choices are staging and prod"
+      );
+    }
+  }
+
+  // Remove will allow the stage to be set on the command line
+  // run like so: serverless remove --stage=staging
+  async remove(inputs = {}) {
+    const { stage } = inputs;
+    if (stage === "staging" || stage === "prod") {
+      const template = await this.load("@serverless/template", stage);
+      const output = await template.remove();
+      return output;
+    }
+  }
+}
+
+module.exports = Deploy;
--- a/packages/serverless-nextjs-component/examples/multiple-instance-environment/.gitignore
+++ b/packages/serverless-nextjs-component/examples/multiple-instance-environment/.gitignore
@@ -1,0 +1,25 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# testing
+/coverage
+
+# next.js
+/.next/
+/out/
+
+# production
+/build
+
+# misc
+.DS_Store
+.env*
+
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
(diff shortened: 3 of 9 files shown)


nodabladam:
Can you elaborate on how this would look in a normal root level project?  I am unclear if I only need to install serverless-next.js or if I need to have the serverless dependencies specifically called out in the package.json.  I tried replacing the "component: ../../" with serverless-next.js:
```
myFrontend:
  component: serverless-next.js
  inputs:
    bucketName: ${env.bucket}
    domain: ${env.domain}
```
...but just got 
```
  myFrontend: 
    appUrl:     https://d2eavu990jpclq.cloudfront.net
    bucketName: luu51mr-n1fvfgc
```

barrysteyn:
> Can you elaborate on how this would look in a normal root level project? I am unclear if I only need to install serverless-next.js or if I need to have the serverless dependencies specifically called out in the package.json.

The example is a normal root level project. But for your purposes, you would want a package.json file that refers to `serverless-next.js`

packages/serverless-nextjs-component/examples/multiple-instance-environment/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "multiple-instance-environment",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "deployStaging": "STAGE=staging serverless",
```
danielcondemarin: Could we change this so it is `serverless --env=staging` and same for prod below. This looks to be the official recommendation here: https://github.com/serverless/template#programmatic-usage-and-custom-environments

packages/serverless-nextjs-component/examples/multiple-instance-environment/serverless.js
```diff
@@ -0,0 +1,34 @@
...
+require('dotenv').config({ path: `${__dirname}/env-${STAGE}`})
+
+/**
+ * Will deploy the serverless component 
+ */
+class Deploy extends Component {
+    async default() {
+      if (STAGE === 'staging' || STAGE === 'prod') {
+        const template = await this.load("@serverless/template", STAGE);
+        const output = await template({ template: path.join("serverless.yml") });
```
danielcondemarin:
```suggestion
        const output = await template({ template: "serverless.yml" });
```

danielcondemarin:
@barrysteyn Thank you for looking after this 🙏 Looks great! Just a few comments I'd like to clarify first.

packages/serverless-nextjs-component/examples/multiple-instance-environment/package.json
barrysteyn: Will do.

barrysteyn:
> @barrysteyn Thank you for looking after this 🙏 Looks great! Just a few comments above I'd like your input first.

Hi Dan

So I changed the code to your suggestions. Thinking about it more, it makes sense to have the remove and deploy work the same (by passing --stage=...). One thing to point out is the README that you refer to is not really official. It is just a README provided by one of the contributors. However it was given the blessing of the Serverless author, so it's official in that sense, but it seems to me that this should not be interpreted as set in stone.

That being said, this example is rock solid in terms of creating separate instances. By passing a seconds argument to the @serverless/template component, all state (in the .serverless folder) is prefixed with that argument. And so you really do create proper instances.

The one thing I can't test is the the domains because I don't have anything in Route53.

danielcondemarin:
> > @barrysteyn Thank you for looking after this 🙏 Looks great! Just a few comments above I'd like your input first.
> 
> Hi Dan
> 
> So I changed the code to your suggestions. Thinking about it more, it makes sense to have the remove and deploy work the same (by passing --stage=...). One thing to point out is the README that you refer to is not really official. It is just a README provided by one of the contributors. However it was given the blessing of the Serverless author, so it's official in that sense, but it seems to me that this should not be interpreted as set in stone.
> 
> That being said, this example is rock solid in terms of creating separate instances. By passing a seconds argument to the @serverless/template component, all state (in the .serverless folder) is prefixed with that argument. And so you really do create proper instances.
> 
> The one thing I can't test is the the domains because I don't have anything in Route53.

Although it doesn't feel like environment configuration is a first class citizen yet, this is a good approach as it keeps separate state files per environment which is the main thing. 
I wouldn't worry much about the domains, it should work as expected. I'm happy to merge this as is, thank you very much 👍

barrysteyn:
I think it's good to merge now. One thing: I removed the STAGE env variable, and so the example page won't display anything stage specific. But that is okay since this is all about configuration, not about what get's displayed in a next page.

packages/serverless-nextjs-component/examples/multiple-instance-environment/serverless.js
```diff
@@ -0,0 +1,33 @@
+const { Component } = require("@serverless/core");
+
+/**
+ * Will deploy the serverless component 
+ */
+class Deploy extends Component {
+    async default() {
```
Enalmada: should this be `async default(inputs = {}) {`

barrysteyn: Darn. Yes, it should. Thanks for spotting. I will submit an update.

barrysteyn: @danielcondemarin I just updated. Can you get this latest version.

barrysteyn: Adam, can you test this please.

packages/serverless-nextjs-component/examples/multiple-instance-environment/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "deployProd": "serverless --stage=prod"
+  },
+  "dependencies": {
+    "next": "9.1.1",
+    "react": "16.11.0",
+    "react-dom": "16.11.0"
+  },
+  "devDependencies": {
+    "dotenv": "^8.1.0",
+    "@serverless/core": "^1.1.2",
```
Enalmada: You might consider adding `"serverless-next.js": "latest" so people can run this right away.  It looks like serverless-next.js loads `@serverless/core` so that could be removed but `@serverless/template` and `serverless` are needed.

barrysteyn: Hi,

I have not set `"serverless-next.js"` to latest since I don't want to break with the other examples that do not do this (it also allows for one to debug/test things easier). And while I agree I could probably remove `@serverless/core` (although I have not tested this), I want it in here to make this as stand alone as possible. You see, this recipe is not well known, and I suspect that other people will take it and adapt it to their needs (possibly without `"serverless-next.js"`)

Issue #220, opened 2019-10-29
Cache control headers for component


barrysteyn:
With the merging of https://github.com/serverless-components/aws-s3/pull/7, cache-control headers can now be implemented for the component in the same fashion as #144 

This issue is created to implement this feature.

danielcondemarin:
> However, the entire folder in https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-nextjs-component/serverless.js#L237 will contain only these types of files?

I believe you are right, looking at an example build:

![image](https://user-images.githubusercontent.com/1122442/68089470-fe672980-fe60-11e9-9721-28324b5b7728.png)

it only seems to contain these directories / files which are fine to be long term cached, so this should be really easy to implement 👍 No need to examine input files etc.

barrysteyn:
Yep :)

I will have a PR out sometime by the end of the day PST.

danielcondemarin:
Done in https://github.com/danielcondemarin/serverless-next.js/pull/226 🎉

PR #226, opened 2019-11-04
cache build assets for component

barrysteyn:
Solves #220


barrysteyn:
Super simple implementation. Tested things manually as well, and things work nicely :)

Issue #207, opened 2019-10-14
Next.js 9.1 public directory support


jayarnielsen:
**Describe the bug**
As of next.js 9.1, the `static` directory should be placed inside of the `public` directory. https://nextjs.org/blog/next-9-1#public-directory-support

With the current serverless component, files in the `public/static/*` folder are unreachable because of the cloudfront rule for `static/*`.

**To Reproduce**
1. place static assets in `public/static`
2. deploy
3. try to access static asset at domain.tld/static

**Expected behavior**
Static asset resolves.

Enalmada:
This would be good to prioritize as old Next users will be encouraged to migrate as console says `The static directory has been deprecated in favor of the public directory. https://err.sh/zeit/next.js/static-dir-deprecated` and new users to Next will put everything in the `public/static` directory.  Moving things to 'public/static' will work locally but not work using serverless-next.

PrimeObjects:
Not sure whether it is the same issue, I have images in /public/images folder, it works file when run locally. After deploy to AWS. I can see a public folder in my S3, However /images/file.mp4 does not work.

skycremaschinin:
opened PR #234 to solve this issue

romainquellec:
Is it closed as the PR is accepted ?

mikestopcontinues:
What's the status on this?

aldrinc:
For folks landing here wondering about why images don't show up when placed in sub folders. I found that having any assets in sub folders under /public does not work when deployed. 

Unfortunately as a workaround I had to move out all my assets from sub folders like /public/images directly into /public so they could be accessed. If anyone knows a different workaround I'm all ears.

Additionally, new assets that are added to public folder don't automatically get the correct permissions when deployed. I have to manually update them as public in S3 so the assets (e.g. images) are visible in my application.

danielcondemarin:
I have tested on the latest version of the project and this should work as expected. 

There is only one caveat, if you have both `static/` and `public/static` folders, only resources in `static/` will be found. I intend to completely remove support for `static/` in the next major version bump to avoid confusion. 

Could folks test and feedback using the latest version?

danielcondemarin:
Closing this as it should be working as expected

keisukekomeda:
I think the case "1" shown below is the only acceptable code.
I am wondering whether all the request for static image assets kick the lambda@Edge, or cloud front cache allows us to get image file without kick lambda@Edge?

[...]

PR #234, opened 2019-11-11
add public cloudfront behaviour

skycremaschinin:
Solves #207

### DIFF
--- a/packages/serverless-nextjs-component/__tests__/deploy.test.js
+++ b/packages/serverless-nextjs-component/__tests__/deploy.test.js
@@ -131,6 +131,9 @@ describe("deploy tests", () => {
               "static/*": {
                 ttl: 86400
               },
+              "public/*": {
+                ttl: 86400
+              },
               "api/*": {
                 ttl: 5,
                 "lambda@edge": {
--- a/packages/serverless-nextjs-component/serverless.js
+++ b/packages/serverless-nextjs-component/serverless.js
@@ -289,6 +289,9 @@ class NextjsComponent extends Component {
           },
           "static/*": {
             ttl: 86400
+          },
+          "public/*": {
+            ttl: 86400
           }
         }
       }


danielcondemarin:
@skycremaschinin Thank you for the PR, but I don't think this will work 🤔 

The new path pattern you've added means requests to `/public/*` will be proxied to S3 but what we need is requests to `/*` that do not match any pages or apis are proxied to S3 public folder. 

To get this working I believe we need 2 things:
1 - Add to the build manifest all files inside `public/` at any level deep. Currently it uses `readdir` which only returns the files directly under the folder. See https://github.com/danielcondemarin/serverless-next.js/blob/0a03156ecafddab71716e3df8d573a12a794d3e7/packages/serverless-nextjs-component/serverless.js#L30

2 - Copy any files inside `public/static` over to `static` folder to keep backwards compatibility. 

@jayarnielsen Since you raised the original issue how does the above sound? Hopefully I didn't miss anything.

skycremaschinin:
@danielcondemarin i was missing the real meaning of public folder:

> Any files in the public directory will be mapped to the root of the domain.

because when public folder is found in src folder a public folder is created on s3 bucket.

Why not simply copy the local public folder to bucket root level?
this would also cover:

> Because public also covers the static directory use case we have decided to deprecate the static directory in favor of creating a public/static folder with the same functionality.

static folder would be copied to s3 root level, and static/* route would work as expected.

danielcondemarin:
> Why not simply copy the local public folder to bucket root level?

That's what I meant by adding to the build manifest all files inside public/ any level deep. i.e. copy the entire folder including sub dirs etc. The Lambda@Edge function then takes care of checking whether there is a file in the public folder for the request path https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-nextjs-component/default-lambda-handler.js#L45

Basically just ensuring the entire folder along with subdirs etc. is copied over should be enough to get this working. 
The reason why is implemented as it is currently is because originally the public folder was meant only for top level file resources like `/robots` `/manifest` etc. It didn't need to support nested routes like images/*

skycremaschinin:
ok, got it.
i'll do some tests and open a new pr.

PR #245, opened 2019-11-21
Fix public

sushi30:
fixes issue [242](https://github.com/danielcondemarin/serverless-next.js/issues/242)

implementation of recursive file list from [here](https://medium.com/stackfame/get-list-of-all-files-in-a-directory-in-node-js-befd31677ec5)


barrysteyn:
There are many ways of doing this. @danielcondemarin used klaw in the plugin, which is more elegant IMHO. I think you should chat with him and enquire how he would like it done.

danielcondemarin:
Thanks folks. I'm happy with the implementation as is. I've just added a test to make sure there are no regression issues introduced later.

PR #249, opened 2019-11-29
Handle routes that may have trailing slash

fuller:
Currently we are seeing a situation where hitting a route directly like `example.com/about` will work as expected, but `example.com/about/` is hitting the `_error.js` file since the paths do not match.

This changes strips the trailing slash before doing path comparisons so that page will be found, then using canonical urls we can get search engines to understand as well.

I'm sure there are better solutions to this problem as well, but this has worked for us in the meantime.


danielcondemarin:
@fuller Thank you for the PR! 

I'd like to avoid regexes whenever possible in the Lambda Handlers, they could lead to DoS attacks, perf issues etc.

Would something like this work instead:

```js
if (path[path.length - 1] === '/') { 
  // strip last / char
  path = path.slice(0, -1)
}
```

Also a simple test case would be great!

danielcondemarin:
I will hold this off until https://github.com/zeit/next.js/pull/6421#issuecomment-472414075 is resolved! At the moment the component behaves in the same way that the next server does so will keep it like that for now.

danielcondemarin:
Closing as there is no activity in https://github.com/zeit/next.js/pull/6421#issuecomment-472414075 i

PR #259, opened 2019-12-15
214 build override

jlaramie:
@danielcondemarin If we aren't going to decouple the build then I'd like a way to pass additional ENVs to the build. As I've progressed the projects that are using this library, I've now run into needing access to serverless variables. Currently I solve this by creating my own custom component that gets fed the variables and does the NextJS builds. The outputs from the custom component get passed into the `serverless-nextjs-component` so that it doesn't run until my `build` component finishes.

I suggest something like this:

- Support `inputs.nextStaticPath`
- Support `inputs.nextBuildArguments`
- Support `inputs.buildScripts`. This defaults to running the standard build with `inputs.nextBuildArguments` but can be overridden with an array of scripts to run.

Here is a mock sample of the default configuration
```
MyNextJSProject:
  component: serverless-next.js
  inputs:
    build: true
    nextStaticDir: './'
    buildScripts:
      - cmd: 'node_modules/.bin/next'
        args: ['build'].concat(inputs.nextBuildArguments)
        cwd: './'
        env:
          envOne: true
```

Benefits

1. Allows the build to stay coupled for basic setups
2. Easily put additional dependencies into the build
3. Supports two types of being overridden: Passing `inputs.nextBuildArguments` or by replacing the `inputs.buildScripts`

Sorry for the rant but I"m just trying to figure out what will get me out of this fork and back into master. I'm going to create a pull request for this.


danielcondemarin:
@jlaramie Just a few questions from me:

- Why an array of `buildScripts`? What else would `buildScripts` be used for other than a customised next build? 

- Would it be possible to condense these options under the same `build` input? I think it will result less confusing for users if is just one input rather than several inputs `build`, `nextBuildArguments`, `buildScripts` that can also clash with each other. For example:

```yaml
# default
MyNextJSProject:
  component: serverless-next.js
  inputs:
    build:
      cmd: 'node_modules/.bin/next'
      args: ['build']
      cwd: './'
```

```yaml
# custom build opts with env etc.
MyNextJSProject:
  component: serverless-next.js
  inputs:
   build:
      cmd: '../some/other/.bin/next'
      args: ['build', 'src']
      cwd: './'
      env:
        foo: ${SomeOtherComponent.foo} 
SomeOtherComponent:
  ...
```

```yaml
# don't build
MyNextJSProject:
  component: serverless-next.js
  inputs:
    build: false
```

Presumably that covers all use cases?

jlaramie:
@danielcondemarin I can make those changes. I took this logic from my own custom serverless component which uses arrays in order to call multiple build scripts. The idea was to be able to call more than one script if you wanted to like if there are two distinct things you want to build with different arguments.

I like the idea of putting this all under `build`

danielcondemarin:
> @danielcondemarin I can make those changes. I took this logic from my own custom serverless component which uses arrays in order to call multiple build scripts. The idea was to be able to call more than one script if you wanted to like if there are two distinct things you want to build with different arguments.
> 
> I like the idea of putting this all under `build`

@jlaramie Feel free to make the changes and we'll get it merged 🙏 

Also, I'd prefer if it wasn't an array. It shouldn't be responsibility of the next component to build other things aside from the next app.

jlaramie:
@danielcondemarin Just pushed the changes you suggested. I also updated the readme and added an example

jlaramie:
@danielcondemarin Yeah really if you need multiple things built you should just modify the build script to call a wrapper command

danielcondemarin:
@jlaramie Excellent work! Thank you 🙏

PR #267, opened 2019-12-23
Properly default build.enabled

### LINKED ISSUE #266: target: "serverless", and yet "pages-manifest not found."
**Describe the bug**
When I try to deploy, I get this error:

> pages-manifest not found. Check if `next.config.js` target is set to 'serverless'

However, my target is "serverless" and the manifest is created at the right location (`.next/serverless/pages-manifest.json`).

Looking a little deeper, it seems the problem is the component isn't regenerating the next build. As in, if I delete `.next/` and run `serverless`, the directory isn't recreated. However, if I run `NODE_ENV=production next build && serverless`, I get a successful deploy.

**To Reproduce**
Just run `serverless` with the following config...

serverless.yml:
```yaml
app:
  component: serverless-next.js
  inputs:
    domain: ["subdomain", "example.com"]
```

next.config.js:
```js
module.exports = {
  target: 'serverless',
};
```

**Expected behavior**
I expect the component to handle the entire build process. Failing that, I expect the docs to describe the correct usage.

```json
{
    "next": "9.1.6",
    "serverless": "1.60.1",
    "serverless-next.js": "1.7.0",
}
```

mikestopcontinues:
I think this should account for the new case outlined in the docs.

Fixes #266

### DIFF
--- a/packages/serverless-nextjs-component/serverless.js
+++ b/packages/serverless-nextjs-component/serverless.js
@@ -238,8 +238,8 @@ class NextjsComponent extends Component {
       : process.cwd();
 
     const buildConfig = {
-      enabled: inputs.build 
-        ? inputs.build !== false && inputs.build.enabled !== false 
+      enabled: inputs.build
+        ? inputs.build !== false && inputs.build.enabled !== false
         : true,
       cmd: "node_modules/.bin/next",
       args: ["build"],
--- a/packages/serverless-nextjs-component/serverless.js
+++ b/packages/serverless-nextjs-component/serverless.js
@@ -238,7 +238,9 @@ class NextjsComponent extends Component {
       : process.cwd();
 
     const buildConfig = {
-      enabled: inputs.build ? inputs.build !== false && inputs.build.enabled !== false : true,
+      enabled: inputs.build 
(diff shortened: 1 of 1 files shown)


danielcondemarin:
@mikestopcontinues Thanks for fixing 👍 Could you have a look at the linting issue making the build fail? I'll merge once that's fixed.

packages/serverless-nextjs-component/serverless.js
```diff
@@ -238,7 +238,7 @@ class NextjsComponent extends Component {
       : process.cwd();
 
     const buildConfig = {
-      enabled: inputs.build && inputs.build.enabled !== false,
+      enabled: inputs.build ? inputs.build !== false && inputs.build.enabled !== false : true,
```
barrysteyn: !== false is the same as saying === true.

mikestopcontinues: `=== true` wouldn't allow objects to pass through to the second case, which is from the original code. So I'm trying to keep to the logic outlined in the original and in the docs. Only exactly false disables builds.

danielcondemarin:
Merged 🚀

Issue #270, opened 2019-12-29
Ability to configure all cloudfront inputs


barrysteyn:
One of the features that I always need to update is the ability to configure cloudfront. Rather than making one off input configurations for each feature (e.g. forwarding certain headers), can I suggest rather allowing one to configure any cloudfront options that are available from [@serverless/aws-cloudfront](https://github.com/serverless-components/aws-cloudfront), like so:

```
myNextApplication:
  component: serverless-next.js
  inputs:
     cloudfront:
        default:
            inputs here
        api:
            inputs here
```

The idea is that anything you have not set could be set by these inputs. For example, I need user-agent forwarding and I want to whitelist what cloudfront allows me to do regarding headers.

This would solve #240 #198 It would also make other future requests asking for cloudfront changes obsolete.

What do you think? I can get a PR for this out quite quickly.

AaronMoat:
This would potentially solve #274 too

danielcondemarin:
@barrysteyn This sounds great 🔥Feel free to PR and I'll review.

stan-sack:
+1 this would be a great feature. I'm having the same issue with user-agent headers and would also like to disable caching for some code paths which such as /user-profile because if the page is rendered on the server then currently it will be cached and shown to other users.

stan-sack:
I just did it myself: https://github.com/danielcondemarin/serverless-next.js/pull/282. Confirmed it works on my own deployment.

@danielcondemarin let me know what you think and i'll add some tests and update the docs.

stan-sack:
I've added tests and documentation. It looks like your remark markdown linter is broken. I dont think the issue its failing me for is valid

barrysteyn:
Hi, I am going to give this one a crack over the weekend. I do note that #282 is doing something similar, but it's not 100% what I had in mind.

skycremaschinin:
hi, 
i need this feature for enable http request and ipv6 disabling. When this pr would be merged?

danielcondemarin:
Sorry I forgot to close the issue. This has been merged already https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-nextjs-component/README.md#custom-cloudfront-configuration

PR #290, opened 2020-02-02
custom cloudfront configurations

barrysteyn:
Solves #270. Allows configuration for both default and api lambda cloudfront options. It will silently ignore any setting to "lambda@edge" (since this should never be changed). Also, besides default and api, no other cloudfront setting will be honored.


packages/serverless-nextjs-component/__tests__/custom-inputs.test.js
```diff
@@ -196,4 +193,89 @@ describe("Custom inputs", () => {
...
+    ],
+    [
+      { defaults: { forward: { headers: "X" } } },
+      { defaults: { forward: { headers: "X" } } }
+    ],
+    [
+      { api: { ttl: 500, "lambda@edge": "ignored value" } },
+      { api: { ttl: 500 } } // expecting lambda@edge value to be ignored
+    ]
+  ])("Lambda memory input", (inputCloudfrontConfig, expectedInConfig) => {
```
danielcondemarin:
```suggestion
  ])("Custom cloudfront configuration", (inputCloudfrontConfig, expectedInConfig) => {
```

packages/serverless-nextjs-component/serverless.js
```diff
@@ -449,14 +452,19 @@ class NextjsComponent extends Component {
 
     const defaultEdgeLambdaPublishOutputs = await defaultEdgeLambda.publishVersion();
 
+    const defaultCloudfrontInputs =
+      (inputs.cloudfront && inputs.cloudfront.defaults) || {};
     const cloudFrontOutputs = await cloudFront({
       defaults: {
         ttl: 0,
         allowedHttpMethods: ["HEAD", "GET"],
+        ...defaultCloudfrontInputs,
```
danielcondemarin: Cloud we not follow the same pattern as above and pass `...defaultCloudfrontInputs` at the end of the object so you don't need the `...defaultCloudfrontInputs.forward` bit?

packages/serverless-nextjs-component/README.md
```diff
@@ -93,6 +93,27 @@ And simply deploy:
 $ serverless
 ```
 
+### Custom Cloudfront configuration
```
danielcondemarin: Can we link this section in the top of the docs (Contents)

danielcondemarin:
Cheers for the PR Barry. Overall LGTM 🚀 Just a few comments as above.

packages/serverless-nextjs-component/serverless.js
barrysteyn: No. Unfortunately, the spread operator is not that clever in JS.  If we did this, and no *forward* key was defined, it would result in no forward section being present since it would get deleted by `... defaultCloudfrontInputs`

packages/serverless-nextjs-component/README.md
barrysteyn: Sure thing...

packages/serverless-nextjs-component/serverless.js
barrysteyn: Hmmmm - you are making me doubt this now... if no forward section is defined, then I think you are correct, it will work... but this is easy to test... I will try it, and if the unit tests pass, then it works like you say ;)

barrysteyn: Ah, okay, I know why I did it now...

So if you have this config:

```
cloudfront:
  defaults:
    forward:
       headers: [ blah ]
```
Then, the forward section would be totally replaced (and not augmented) by headers. I thought that this would become a nightmare with people making mistakes and then creating issues. So the way I have it now, it always tries to augment the forward section, never replaces it.

danielcondemarin: Makes sense! Let's keep it as you have it then.

PR #293, opened 2020-02-04
Add option to use naked domain

### LINKED ISSUE #192: Redirect non-www to www and vice-versa
Add an option to enable redirect from `example.com` to `www.example.com` and vice-versa.
For example, add option `redirect` to force redirect to www:

```
myApp:
  component: serverless-next.js
  inputs:
    domain: ["www", "example.com"]
    redirect: true
```

endigo:
I need to use only the naked domain. 

If `useNakedDomain == true` then prevent to pass option subdomains to `@serverless/domain` component.

Close #192


danielcondemarin:
@endigo Is this working fine for you? I was under the impression this isn't supported by the (domain component)[https://github.com/serverless-components/domain] but I might be wrong. cc. @eahefnawy

endigo:
@danielcondemarin  yeah, it's not working as I expected. 

If I don't pass subdomains to domain component, it uses a default empty object.
https://github.com/serverless-components/domain/blob/master/utils.js#L46

But in reality, it doesn't work as I expected.

barrysteyn:
> @endigo Is this working fine for you? I was under the impression this isn't supported by the [domain component](https://github.com/serverless-components/domain) but I might be wrong. cc. @eahefnawy

Hi, sorry this took a long time. Rather can you alter the underlying the component that handles domain (@serverless/domain) to provide logic that will do what you want it to do. I assume this would be a switch. Once it does this, then this project can just pass the setting to that component. This component really should not be involved with business logic regarding domains - it's not what it does.

PR #294, opened 2020-02-05
useNakedDomain option added

### LINKED ISSUE #192: Redirect non-www to www and vice-versa

endigo:
This time it definitely works as I expected.

Ref #293 

Close #192


barrysteyn:
The domain component has this feature already built in?

endigo:
@barrysteyn But serverless-next.js component doesn't expose that feature.

endigo:
@danielcondemarin Hey, please review this PR.

SarKurd:
@danielcondemarin could you confirm if we still need this? we are about to deploy a V2 of our app to a naked domain which uses this plugin, we don't have a way to test if it works as we don't have a spare domain to test on it as the V1 of our app uses it.

owlas:
Hey @endigo thanks for adding this. What exactly would I use in my config to have naked domain AND www subdomain point to my application?

owlas:
Actually this already works for me:
```yaml
myWhackyApp:
  component: serverless-next.js
  inputs:
    domain: "mydomain.com"
```

endigo:
@owlas It creates an only naked domain. Then you can redirect `www` to the naked one.

owboateng:
Sorry but can someone please show how to use the naked domain option. Currently I get two aliases created in Route53 for www and non www with both pointing to cloudfront. This causes them to be seen as two different sites which is affecting my SEO

endigo:
@owboateng Unfortunately, this PR is not merged yet.

You can create custom Cloudfront distribution which redirects to the other one.

When you deploy your web to AWS using serverless command it will fail because of a duplicated record of Cloudfront, but it'll work fine.

I tried to solve this issue with this PR.

Thanks

owboateng:
@endigo Thanks for the quick response. Any one here knows when this PR will be integrated? Thanks

lone-cloud:
do you mind adding a test and fixing the conflicts. I'll test this out after.

lone-cloud:
I'd like to handle this differently. Closing but I'll be addressing it in #465

PR #295, opened 2020-02-05
feat: add omitErrorPage setting

oridyn:


danielcondemarin:
@oridyn Sorry but this would mean going out of parity with next.js default behaviour which is not a good idea.

PR #298, opened 2020-02-16
Catch all dynamic routes

danielcondemarin:
Implements https://nextjs.org/docs/routing/dynamic-routes#catch-all-routes

Closes https://github.com/danielcondemarin/serverless-next.js/issues/286


paleite:
Looks really great, but I can't seem to get it to work for API-routes: https://nextjs.org/docs/api-routes/dynamic-api-routes#catch-all-api-routes 

My file is `pages/api/[...path].js` locally, but it's missing in the API-lambda. Other API-paths still work though.

yanv1991:
@danielcondemarin I created a demo project taken from next docs about catch all routes https://d1z3sbtml48wer.cloudfront.net/post/2020/first-post/with/catch/all/routes but when you access the url from ssr it throws an 503 error from cloudfront

PR #325, opened 2020-03-20
Feature/resource region support

a14m:
Add region support (mainly to be able to provision s3 buckets in certain regions)


a14m:
There seems to be a problem with the configuration that caused the cloudfront distribution to return access denied

PR #329, opened 2020-03-25
Fix the issue with TypeError hasHeader is not a function

a14m:
When deploying to netlify or AWS (both depend on AWS-lambda)
The SSR components (using `getServerSideProps` fails with the following
error:
```
  ERROR	TypeError: res.hasHeader is not a function
    at sendPayload (/var/task/pages/test.js:5787:41)
    at renderReqToHTML (/var/task/pages/test.js:1773:13)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Module.render (/var/task/pages/test.js:1816:22)
```
This is caused by nextjs trying to use the default node `hasHeader`
method here: https://github.com/zeit/next.js/blob/v9.3.1/packages/next/next-server/server/send-payload.ts#L25

This commit fix the issue reported here: https://github.com/zeit/next.js/issues/11223
which I could verify also happening on our stack by adding the method
that mimics the nodeJS method (using `res.getHeader(name)` internally wrapping it
in boolean response)

### DIFF
--- a/packages/next-aws-lambda/lib/__tests__/compatLayer.response.test.js
+++ b/packages/next-aws-lambda/lib/__tests__/compatLayer.response.test.js
@@ -265,6 +265,19 @@ describe("compatLayer.response", () => {
     });
   });
 
+  it("hasHeader", () => {
+    const { res } = create({
+      requestContext: {
+        path: "/"
+      },
+      headers: {}
+    });
+    res.setHeader("x-custom-1", "1");
+
+    expect(res.hasHeader("x-custom-1")).toBe(true);
+    expect(res.hasHeader("x-custom-2")).toBe(false);
+  });
+
   it(`res.write('ok')`, done => {
     const { res } = create(
       {
--- a/packages/next-aws-lambda/lib/compatLayer.js
+++ b/packages/next-aws-lambda/lib/compatLayer.js
@@ -97,6 +97,9 @@ const reqResMapper = (event, callback) => {
   res.getHeaders = () => {
     return res.headers;
   };
+  res.hasHeader = name => {
+    return !!res.getHeader(name);
+  };
 
   const onResEnd = (callback, resolve) => text => {
     if (text) res.write(text);


danielcondemarin:
Thanks for fixing! Do you mind adding a quick test for `hasHeader` similar to https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/next-aws-lambda/lib/__tests__/compatLayer.response.test.js#L251 ?

a14m:
also quick question, is there a reason why `getHeader` lower cases the header name, I think this can cause unintended behaviour... for example if
```
res.setHeader("X-CUSTOM", "VAL")
res.getHeader("X-CUSTOM") => null
res.hasHeader("X-CUSTOM") => false

res.setHeader("x-custom", "val")
res.getHeader("X-CUSTOM") => "val"
res.hasHeader("X-CUSTOM") => true
```

a14m:
Is it possible to give a time frame when this is supposed to be released (we depend on this to deploy to AWS and also to deploy to netlify through https://github.com/FinnWoelm/next-on-netlify/blob/master/package.json#L27)?

joseomar10:
Thanks for fixing! 
Is it possible to know when this fix could be published? I am depending on this to deploy aws. Please :)

iamgbayer:
Thanks for fixing @a14m, I also would like to know when this cloud be published.

danielcondemarin:
Hi folks, sorry for the delay on publishing. 
It has now been released next-aws-lambda@2.3.7, serverless-nextjs-plugin@2.4.8 🎉

iamgbayer:
Hi @danielcondemarin,

I'm still receiving an error with `hasHeader` in a page that uses `getServerSideProps`, even updating the version, I'm using `serverless-next.js 1.9.7`

```
2020-03-29T20:11:19.446Z b6935fb7-e719-49f8-a684-d85e65904937 ERROR TypeError: res.hasHeader is not a function at sendPayload (/var/task/pages/index.js:15193:41) at renderReqToHTML (/var/task/pages/index.js:22480:13)
```

a14m:
You need to update `next-aws-lambda` to at least `^2.3.7`

iamgbayer:
@a14m, why I need to update it? It isn't a dependency from `serverless-next.js`? Already has updated https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-nextjs-component/package.json#L32

a14m:
In theory that should fix the issue, but due to the complexity in package management (and JavaScript in general 😅) that’s not the case...
Can you verify that the version of `next-aws-lambda` is at least `2.3.7` In `package-lock.json` or `yarn.lock`

iamgbayer:
In `yarn.lock`, I found `next-aws-lambda` with version `2.3.8`, this should fix the problem, but I think it isn't happening in another part of the code.

joseomar10:
@danielcondemarin I think this fix also needs to be added to `next-aws-cloudfront.js` of `serverless-nextjs-component`
I'm using serverless-next.js on version 1.9.7 but have same error

zorzysty:
I have "serverless-next.js": "^1.9.7" with "next-aws-lambda": "^2.3.8"
And the error still occurs.

a14m:
I've added a fix for the issue (and the inconsistent header cases there too) in the 👆 PR

PR #342, opened 2020-04-03
Fix the issue with TypeError hasHeader is not a function

a14m:
When deploying to netlify or AWS (both depend on AWS-lambda)
The SSR components (using `getServerSideProps` fails with the following
error:
```
  ERROR	TypeError: res.hasHeader is not a function
    at sendPayload (/var/task/pages/test.js:5787:41)
    at renderReqToHTML (/var/task/pages/test.js:1773:13)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Module.render (/var/task/pages/test.js:1816:22)
```
This is caused by nextjs trying to use the default node `hasHeader`
method here: https://github.com/zeit/next.js/blob/v9.3.1/packages/next/next-server/server/send-payload.ts#L25

This commit fix the issue reported here: zeit/next.js#11223
which I could verify also happening on our stack by adding the method
that mimics the nodeJS method (using `res.getHeader(name)` internally wrapping it
in boolean response)

Previously fixed for next-aws-lambda in #329
also fix the header inconsistencies previously fixed for next-aws-lambda in #331

### DIFF
--- a/packages/serverless-nextjs-component/__tests__/next-aws-cloudfront.response.test.js
+++ b/packages/serverless-nextjs-component/__tests__/next-aws-cloudfront.response.test.js
@@ -223,6 +223,19 @@ describe("Response Tests", () => {
     });
   });
 
+  it("hasHeader", () => {
+    const { res } = create({
+      request: {
+        path: "/",
+        headers: {}
+      }
+    });
+    res.setHeader("x-custom-1", "1");
+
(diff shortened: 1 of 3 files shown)


danielcondemarin:
Thanks again 🙏

zorzysty:
What's the ETA of this getting into release?

danielcondemarin:
I've just published it. Should be available in `serverless-next.js@1.9.8` 🎉

Issue #271, opened 2019-12-30
Custom domains don't work w/ alternate DNS provider (ie. Cloudflare)


wilsonpage:
**Describe the bug**

I'm attempting to use my own domain which is managed by Cloudflare. I'm able to set this up manually by:

1. Deploying the app without a `domain` in the config
2. Creating a CNAME record (in Cloudflare) pointing to the app's `xxx.cloudfront.net` endpoint
3. Creating and validating a certificate in ACM for `batch.songwhip.com`
4. Changing the Cloudfront's domain/CNAME to accept request from my custom domain (batch.songwhip.com)

<img width="911" alt="Screenshot 2019-12-30 at 10 56 55" src="https://user-images.githubusercontent.com/1020551/71579160-4322e000-2af3-11ea-818d-ecc69630c33e.png">

But using the `domain` config in `serverless.yml` always fails with:

`Error: Domain songwhip.com was not found in your AWS account. Please purchase it from Route53 first then try again.`

## Questions

1. Does this plugin only support domains registered/managed by Route53?
2. If so are there any workarounds you can think of?

barrysteyn:
I am using cloudflare to manage my domains, and I pretty much do what you are doing. However, if you want the domain to be automatically setup in cloudfront upon deploy, you will need to use Route53. This requirement does not come from this component, rather, it comes from the fact that this component is using another component that sets things up which insists upon this.

Enalmada:
Note that you don't actually have to change your nameservers to Route53...you just have to create a host entry there so the deploy process wont fail.

wilsonpage:
@Enalmada interesting! Would you be able to offer a little more detail here?

wilsonpage:
@barrysteyn thanks, glad I'm not the only one in this boat! It seems once the domains are manually wired up preferences stick, so it's probably just a one time config job for new projects. Either way, it's nice to have 'config as code' wherever possible.

Enalmada:
@wilsonpage If you try and setup a serverless-next without any domain set, you will get a response like this:
```
  myApp: 
    appUrl:     https://dabcdefghijk.cloudfront.com
    bucketName: my-bucket
```
Add the appUrl as a cname in your CloudFlare.  Then create a hosted zone in route53 for your domain.  You don't actually need to purchase a domain in route53 to created a hosted zone.  After that, you should be able to get serverless-next working.

danielcondemarin:
[...]

Issue #251, opened 2019-12-04
Custom domain does not work


yankouskia:
Create simple configuration for deploying nextjs application:
```yml
app:
  component: serverless-next.js
  inputs:
    domain: www.app.com # tried with app.com as well
    bucketName: my-app-bucket
```

After running:
```sh
serverless --aws-profile profile --debug
```

Always getting: 
```
  Error: Domain www.app.com was not found in your AWS account. Please purchase it from Route53 first then try again.
    at getDomainHostedZoneId (/Users/me/.serverless/components/registry/npm/serverless-next.js@1.6.1/node_modules/@serverless/domain/utils.js:120:11)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Domain.default (/Users/me/.serverless/components/registry/npm/serverless-next.js@1.6.1/node_modules/@serverless/domain/serverless.js:58:32)
    at async NextjsComponent.deploy (/Users/me/.serverless/components/registry/npm/serverless-next.js@1.6.1/node_modules/serverless-next.js/serverless.js:440:29)
    at async fn (/Users/me/projects/app/node_modules/@serverless/template/utils.js:280:41)
    at async Promise.all (index 0)
    at async executeGraph (/Users/me/projects/app/node_modules/@serverless/template/utils.js:294:3)
    at async Template.default (/Users/me/projects/app/node_modules/@serverless/template/serverless.js:67:38)
    at async Object.runComponents (/Users/me/projects/app/node_modules/@serverless/cli/src/index.js:211:17)

  11s › app › Error: Domain www.app.com was not found in your AWS account. Please purchase it from Route53 first then try again.
```

I have purchased corresponding domain in Route53:
```sh
AWS_PROFILE=profile aws route53domains list-domains

{
    "Domains": [
        {
            "DomainName": "app.com",
            "AutoRenew": true,
            "TransferLock": false,
            "Expiry": 123
        }
    ]
}
```

**Versions:**
 - OS: MacOS Mojave 10.14.6 
 - serverless: 1.58.0
 - serverless-next.js: 1.6.1
 - next: 9.1.4

nodabladam:
I got that same error when I didn't actually have my app.com domain purchased due to using another registrar.  When adding it to Route 53 it then worked.  I believe the format is:
```
    domain:
      - ${subdomain}
      - ${domain}
```

yankouskia:
@nodabladam getting almost the same result...
`Domain www was not found`
[...]

PR #366, opened 2020-04-22
update readme with custom domain name instructions

owlas:
solves #251 #271 

Currently the README says you must purchase a domain through AWS Route53. Actually you can use any domain name registrar but you must use Route53 for DNS hosting. I've added a link to the AWS docs for migrating your DNS hosting to Route53.

The good news is that any existing domain can be used as long as the user is happy to have their DNS records managed in the same AWS account as the deployed application.


owlas:
Thanks!

PR #377, opened 2020-05-02
Allow setting of additional lambda triggers for default and api

gehan:
Allow passing of additional lambda triggers to default/api. Note it still overwrites any `origin-request` sent through.

```yaml
    cloudfront:
      defaults: # options for lambda that handle SSR
        lambda@edge:
          viewer-response: lambdaarn:version
      api:
        lambda@edge:
          viewer-response: lambdaarn:version
```

### DIFF
--- a/packages/serverless-component/README.md
+++ b/packages/serverless-component/README.md
@@ -111,7 +111,11 @@ myNextApplication:
           headers: [CloudFront-Is-Desktop-Viewer, CloudFront-Is-Mobile-Viewer, CloudFront-Is-Tablet-Viewer]
         lambda@edge:
           viewer-response: lambdaarn:version
-      api: # options for lambdas that handle API request
+      api: # options for lambdas that handle API
+        ttl: 10
+      static: # options for user static assets
+        ttl: 10
+      next: # options for build assets
         ttl: 10
       origins: # options for custom origins and behaviors
         - url: /static
--- a/packages/serverless-component/README.md
+++ b/packages/serverless-component/README.md
@@ -109,6 +109,8 @@ myNextApplication:
       defaults: # options for lambda that handle SSR
         forward:
           headers: [CloudFront-Is-Desktop-Viewer, CloudFront-Is-Mobile-Viewer, CloudFront-Is-Tablet-Viewer]
+        lambda@edge:
+          viewer-response: lambdaarn:version
       api: # options for lambdas that handle API request
         ttl: 10
       origins: # options for custom origins and behaviors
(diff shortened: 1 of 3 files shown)


danielcondemarin:
@gehan Could you explain why you need this? I'm skeptical of allowing `viewer-*` triggers unless there is a very good reason for it.

gehan:
We add `viewer-response` lambdas to alter the headers - nothing major.

I'd say it's just a matter of exposing more AWS capabilities to the user which are available anyway really. Also the docs say that you can set custom cloudfront configuration using options listed in the aws-cloudfront component - so it's also making it more consistent with that. I had originally expected to be able to pass an origins object too since it said this - which is why I ended up submitted that pull request!

danielcondemarin:
@gehan @stan-sack We need to agree on an approach that doesn't make this PR conflict with https://github.com/danielcondemarin/serverless-next.js/pull/282.

gehan:
I guess we'll end up being able to do this right?

```yaml
next:
  component: serverless-next.js
  inputs:
    cloudfront:
      # basic 4 caches   
      defaults:
        ttl: 10
        lambda@edge:
          viewer-response: lambdaarn:version
      /api/*:
        ttl: 10
        lambda@edge:
          viewer-response: lambdaarn:version
      /static/*:
        ttl: 10
        lambda@edge:
          viewer-response: lambdaarn:version
      /_next/*:
        ttl: 10
        lambda@edge:
          viewer-response: lambdaarn:version
    # extra caches  
      /another-page/*:
        ttl: 10
        lambda@edge:
          viewer-response: lambdaarn:version
```

Should be a simple modification to #282 I would have thought to do this - probably around here from looking very briefly. https://github.com/danielcondemarin/serverless-next.js/blob/4a3880de76d00ed0d8d36cab95e957f8fb57874c/packages/serverless-component/serverless.js#L295

stan-sack:
I'll update my PR in the next 24 hours

gehan:
@stan-sack you'll need to reinstate this code though - https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-component/serverless.js#L147

i.e. allowing extra origins that was merged the other day

stan-sack:
my PR is updated PTAL

danielcondemarin:
I've already mentioned in https://github.com/danielcondemarin/serverless-next.js/pull/282#issuecomment-626057952 but will leave here for reference as well, we need to prevent overriding `origin-response`. For now let's allow it only for viewer-request and viewer-response. 
There is some upcoming work that will require the project to use the `origin-response` trigger.

danielcondemarin:
@gehan Can I close this now considering https://github.com/danielcondemarin/serverless-next.js/pull/282 was merged?

oran1248:
@danielcondemarin 
I think its related to this issue: every time I deploy my app, my CloudFront Viewer-Request trigger is removed. DO you know how can I solve it?

PR #381, opened 2020-05-04
Ability to set a custom description from yml file.

Qubica:
Makes it easier to read the setup in AWS console.


danielcondemarin:
Thanks for the PR @Qubica ! Could you add a test to https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/serverless-component/__tests__/custom-inputs.test.js please 🙏 ?

Issue #387, opened 2020-05-06
Handle errors from @serverless/domain


jlaramie:
**Describe the bug**
Despite being the last and optional step of deploying serverless-nextjs, an error during this process results in missing output data from the serverless component even though it succeeded otherwise.

**To Reproduce**
1. Setup a serverless-nextjs component with a domain that will fail
2. Deploy
4. See error and missing final outputs even though the cloudfront URL will work just fine

**Expected behavior**
Outputs should be displayed since the build succeeded. 

An error about the domain should display.

PR #388, opened 2020-05-06
Added error handling for domain component failing.

jlaramie:
Not sure how we want to output the errors these days but this is what I've been doing internally to handle deployments where setting the domain throws an error

Fix for #387


danielcondemarin:
I'm going to be closing this .. there will be a few upcoming changes that will improve error handling.

PR #394, opened 2020-05-08
Added support for public/static folder cache control

### LINKED ISSUE #392: Missing Cache-Control header on /public assets
**Describe the bug**
Deployed with latest serverless-nextjs and checked my pagespeed results, and I'm seeing none of the public assets being cached(the last one is actually the recaptcha image from google, not one of my own): 
PageSpeed_Insights 

I've checked the S3 bucket, and Cache-Control is indeed missing from assets: 

S3_Management_Console (1)

However, everything under `_next` is being correctly cached: 

S3_Management_Console (2)

**To Reproduce**
1. deploy app with at least one asset in /public
2. check the assets metadata in S3

**Expected behavior**
Public assets set the Cache-Control header. 

 - OS: same thing from happens from local Mac and Ubuntu CI
 - Version: `serverless-next.js@^1.11.3`

DanielOrtel:
Fixes #392 

Adds support for setting the `Cache-Control` headers on `/public` or `/static` folder assets. 

Format: 
```yaml
myNextApplication:
  component: serverless-next.js
  inputs:
    publicAssetCache: true
```

This uses the default of `Cache-Control=public, max-age=31536000, must-revalidate`(not sure if `must-revalidate` is even needed) for most common image formats: `/\.(gif|jpe?g|jp2|tiff|png|webp|bmp|svg)$/i`. 

This can be customized by specifying `value` and/or `test`. If one is missing, it will use the default value:
```yaml
myNextApplication:
  component: serverless-next.js
  inputs:
    publicAssetCache:
      value: public, max-age=604800
      test: /\.(gif|jpe?g|png|txt|xml)$/i
```

**Update**: 
Based on feedback, `Cache-Control` is applied by default to the public folder. Setting it to false disables this, using an object allows for configuration: 
```yaml
myNextApplication:
  component: serverless-next.js
  inputs:
    publicDirectoryCache:
      value: public, max-age=604800
      test: /\.(gif|jpe?g|png|txt|xml)$/i
```

```yaml
myNextApplication:
  component: serverless-next.js
  inputs:
    publicDirectoryCache: false
```

Tested with both the default settings: 
default_cache

And using custom settings: 
custom_cache_policy


packages/serverless-component/serverless.js
```diff
@@ -253,8 +256,6 @@ class NextjsComponent extends Component {
         allowedHttpMethods: ["HEAD", "GET"],
         ...defaultCloudfrontInputs,
         forward: {
-          cookies: "all",
```
DanielOrtel, line 256: Not part of the PR but these keys were duplicated.

danielcondemarin:
@DanielOrtel Great work thank you 🙏 

One minor thing, rather than calling it publicAssetCache could you call it `publicDirectoryCache`. I think that will be more familiar to next.js devs as is officially named the public directory.

packages/s3-static-assets/tests/upload-assets.test.ts
```diff
@@ -180,3 +192,35 @@ describe.each`
     });
   }
 );
+
+describe("Content upload options", () => {
+  describe.each([
```
danielcondemarin, line 197: Could you use [this way ](https://github.com/danielcondemarin/serverless-next.js/blob/master/packages/s3-static-assets/tests/upload-assets.test.ts#L95) of testing with `describe.each` to keep it consistent with other packages?

packages/s3-static-assets/src/lib/constants.ts
```diff
@@ -1,2 +1,5 @@
 export const IMMUTABLE_CACHE_CONTROL_HEADER =
   "public, max-age=31536000, immutable";
+
+export const DEFAULT_ASSET_CACHE_CONTROL_HEADER = "public, max-age=31536000, must-revalidate";
```
danielcondemarin, line 4:
```suggestion
export const DEFAULT_PUBLIC_DIR_CACHE_CONTROL = "public, max-age=31536000, must-revalidate";
```

packages/s3-static-assets/src/lib/constants.ts
```diff
@@ -1,2 +1,5 @@
...
+export const DEFAULT_ASSET_CACHE_REGEX = /\.(gif|jpe?g|jp2|tiff|png|webp|bmp|svg)$/i;
```
danielcondemarin, line 5:
```suggestion
export const DEFAULT_PUBLIC_DIR_CACHE_REGEX = /\.(gif|jpe?g|jp2|tiff|png|webp|bmp|svg)$/i;
```

packages/serverless-component/README.md
```diff
@@ -219,6 +226,31 @@ myNextApplication:
       apiLambda: fooApiLambda
 ```
 
+### Public asset configuration
+
+By default, assets under `/public` or `/static` don't have cache control applied. However, you can overwrite this by setting `publicAssetCache`.
```
danielcondemarin, line 231: Can we change the default to be caching the `public/` directory for 1 year? That should be more common than not wanting to cache it. Also, users can opt out by setting it to `false`.

danielcondemarin:
Also, may be worth a test at the `serverless-component/` package level to test the integration with `s3-static-assets`. https://github.com/danielcondemarin/serverless-next.js/tree/master/packages/serverless-component/__tests__

packages/s3-static-assets/tests/upload-assets.test.ts
DanielOrtel: Meaning remove the nesting and simply keep at the root?

packages/serverless-component/README.md
DanielOrtel: yeah, sure, I implemented it with `false` as the default, since that's the current behavior, so this wouldn't be a breaking change. But sure, I can add true as the default.

packages/s3-static-assets/tests/upload-assets.test.ts
danielcondemarin: I mean not to use the square bracket notation and instead use string interpolation 
```
describe.each`
...
`
```

DanielOrtel: Right, I don't know why I thought it was about the nesting 😄

DanielOrtel:
@danielcondemarin I've pushed a new commit addressing the feedback. I'll squash that one before merging.

danielcondemarin:
@DanielOrtel Thanks for addressing all the feedback and quick turnaround.

I'll merge and publish in a bit 🚀

DanielOrtel:
One minor thing: the Readme of `packages/serverless-component` was updated but I failed to notice that this is mirrored in the main Readme, that should probably get updated as well.

danielcondemarin:
> One minor thing: the Readme of `packages/serverless-component` was updated but I failed to notice that this is mirrored in the main Readme, that should probably get updated as well.

Thanks for the heads up. I'll remove the one from `packages/serverless-component` I think and just keep the main one going forwards. Keeping them in sync is a bit of a pain.

PR #405, opened 2020-05-17
Serverless trace target support

danielcondemarin:
### Problem

Next.js "serverless" target creates one bundle per SSR page in your application. The bundle contains everything the page needs for it to be server side rendered in Lambda. The problem with this is it leads to a lot of duplicated code / dependencies across page bundles. I started a discussion on this a while back on the Next repo: https://github.com/zeit/next.js/discussions/10456.
This is particularly an issue for applications with a large number of server side rendered pages that need deploying to Lambda (@Edge or not), where the artefact size limit set by AWS is 250MB (uncompressed). 

### Solution

Next.js core team recently introduced a new target: `experimental-serverless-trace` which rather than bundling all dependencies / node_modules into each page bundle, it marks node_modules as `externals` in webpack thus creating thin serverless wrappers around the pages. This is also the target being used by [the Now platform since next 9.0.4](https://github.com/zeit/now/blob/master/packages/now-next/src/index.ts#L806).

With this PR serverless-next.js will copy the dependencies of the next pages into the lambda artefacts, using [@zeit/node-file-trace](https://github.com/zeit/node-file-trace) to track down which files are needed, same as `now-next` builder does.

This change will be non breaking, the "serverless" target will continue to be used for a while but that may change in future. 

To benefit from the new target, make sure when you are running `$ serverless` to deploy, your `next.config.js` is set to:

```js
module.exports = {
  target: "experimental-serverless-trace",
  ...
}
```

Even though is prefixed with `experimental-` it is the target the Now platform is using behind the scenes so it should be safe for `serverless-next.js` to use. 

Related issues: https://github.com/danielcondemarin/serverless-next.js/issues/236


kartikag01:
i am getting error during build, 
`Error: ENOENT: no such file or directory, stat '/home/project/.next/prerender-manifest.json'`

"next": "9.1.4",
"react": "16.12.0",
"serverless-next.js": "^1.12.0-alpha.4"

danielcondemarin:
@kartikag01 Could you try bumping your next version to 9.3.0 or greater?

TuuZzee:
I am trying to use this feature to deploy my projects which include Firestore, but when I do I get the following error during the build:
```
error:
  Error: Source and destination must not be the same.
    at /Users/myNextApplication/.serverless/components/registry/npm/serverless-next.js@1.13.0/node_modules/@sls-next/lambda-at-edge/node_modules/fs-extra/lib/util/stat.js:39:17
    at cb (util.js:196:39)
    at processTicksAndRejections (internal/process/task_queues.js:81:21)

  90s › myNextApplication › Error: Source and destination must not be the same.
```

I am using the `useServerlessTraceTarget: true` and I set `target: 'experimental-serverless-trace'` in the `next.config.js` file.

The config I am using:
```
    "firebase": "^7.15.1",
    "next": "^9.4.4",
    "redux-firestore": "^0.13.0",
    "serverless": "^1.73.1",
    "serverless-next.js": "^1.14.0",
```

PR #407, opened 2020-05-18
Custom cloudfront bug fixes

stan-sack:
Hey, I just found a few bugs when I went to use the custom cloudfront configs.

1. custom configs for `static/*` may fail validation because there wont be any pages at these paths (the same for `_next/*`)
2. The nodejs AWS SDK will throw an error if you have an undefined value for a key, meaning we need to preprocess the values

correct me if im wrong, but we have tests for the second case, i think the mock API behaves differently

### DIFF
--- a/packages/serverless-component/__tests__/custom-inputs.test.js
+++ b/packages/serverless-component/__tests__/custom-inputs.test.js
@@ -520,10 +520,10 @@ describe("Custom inputs", () => {
         "PATCH"
       ],
       "lambda@edge": {
-        "origin-request":
-          "arn:aws:lambda:us-east-1:123456789012:function:my-func:v1",
         ...(expectedInConfig["api/*"] &&
-          expectedInConfig["api/*"]["lambda@edge"])
+          expectedInConfig["api/*"]["lambda@edge"]),
+        "origin-request":
+          "arn:aws:lambda:us-east-1:123456789012:function:my-func:v1"
       }
     };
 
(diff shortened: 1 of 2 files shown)


danielcondemarin:
Thanks @stan-sack ! Could you add a test for this? There is a lot of logic around the CloudFront custom config. I'd rather have it tested to ensure there is no regression in future releases. 

Also, if you could take a look at **3.** in https://github.com/danielcondemarin/serverless-next.js/issues/236#issuecomment-630234932 that would be ace. I suspect setting any values on the `api:` cloud front configuration is broken. Maybe `stillToMatch.delete("api/*")` should be `stillToMatch.delete("api")` ?

stan-sack:
internally you have defined the key for api cloudfront behaviour as "api/*" so i think this should stay the same. Theres a bug in some logic that is overriding the api lambda ARN. I just pushed a fix, you'll see what I mean. I will add some better tests tonight.

It gets more complicated when deciding how to pass overrides for `api`, `static` and `_next` behaviour. Because internally the keys are `api/*`, `static/*` and `_next/*` we would need custom logic to look for keys such as `api`, `static` etc... and mutate them. Right now if you want to override the api behaviour you need to override the 'api/*' key in yaml. I think that it should continue to behave like this and it should just be documented. Please let me know what you think

stan-sack:
the issue with the test is that the expected value is derived from the input value. thats why it didnt pick it up. I cant see a non-verbose way to test this.

PR #412, opened 2020-05-21
Explicitly set host header for s3 origin

stan-sack:
## Background
- I ran into some issues when I started forwarding the `Host` header from cloudfront. It causes signature invalidation on the request because the host is used to generate the signature
- https://github.com/danielcondemarin/serverless-next.js/issues/408

## Change
- Explicitly set the host header when forwarding requests to S3 origin

## Testing
- Tested on my own website. Confirmed the fix


danielcondemarin:
Thanks @stan-sack for PR'ing. I've added a test and also set the host header for dynamic HTML pages which was missing.

Issue #236, opened 2019-11-12
RequestEntityTooLargeException: Request must be smaller than 69905067 bytes for the CreateFunction operation


jschimmoeller:
Hey daniel, 

I finally had some time to check the latest code against my large next.js project and I am get this error; 

  error:
  RequestEntityTooLargeException: Request must be smaller than 69905067 bytes for the CreateFunction operation
    at Object.extractError (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\protocol\json.js:51:27)
    at Request.extractError (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\protocol\rest_json.js:55:8)
    at Request.callListeners (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\sequential_executor.js:106:20)
    at Request.emit (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\sequential_executor.js:78:10)
    at Request.emit (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\request.js:683:14)
    at Request.transition (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\request.js:22:10)
    at AcceptorStateMachine.runTo (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\state_machine.js:14:12)
    at C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\state_machine.js:26:10
    at Request.<anonymous> (C:\Users\james\.serverless\components\registry\npm\serverless-next.js@1.5.0\node_modules\serverless-next.js\node_modules\aws-sdk\lib\request.js:38:9)
[...]

PR #425, opened 2020-06-03
Resolve dependencies using .next/serverless as base path

thiagozf:
### Problem
When using the new `experimental-serverless-trace` build target, the Next.js Serverless Component is copying Lambda dependencies to wrong destinations in two identified scenarios (see issue #236):

- Dynamic chunks are currently going to `"<root>/.next/serverless"`, when they should be placed directly into `"<root>"`.
- In monorepo setups, `node_modules` should always be copied to `<root>`

_PS.: `<root>` is `"<next config path>/.serverless_nextjs/<lambda handler>"`_

### Solution
The component now uses `<next config path>/.next/serverless` as the relative base path to calculate Lambda handler dependencies destinations. Also, all `node_modules` files are now copied to `<root>/node_modules`, ignoring relative parent paths. These changes should resolve the issues in both scenarios.


danielcondemarin:
Getting this merged despite the test failures. They seem to be Travis specific issues. I'll take a look at that separately.

Issue #357, opened 2020-04-13
The root level resource request of `service-worker.js` does not forward to S3 bucket.


krish-dev:
**Describe the bug**
- nextjs PWA not working.
- the root level request of `service-worker.js` does not forward to `static/service-worker.js`

**To Reproduce**
I have used [next-offline](https://github.com/hanford/next-offline#service-worker-path) for building pwa application. This plugin generates `service-worker.js` under `static/service-worker.js`

**Expected behavior**
Public resources. Requests to root level resources like `/robots.txt`, `/favicon.ico`, `/manifest.json`, etc. These are forwarded to S3. `service-worker.js` this is also a root level resources. So, this should forward to S3 as well.

krish-dev:
> **Expected behavior**
> Public resources. Requests to root level resources like `/robots.txt`, `/favicon.ico`, `/manifest.json`, etc. These are forwarded to S3. `service-worker.js` this is also a root level resources. So, this should forward to S3 as well.

I think the best approach should be if we allow adding this kind of custom strategy from the configuration (serverless.yml) file.

ashbuilds:
For now I found the work around by changing default path of server-worker in `next.config.js` config to `public`

```
 workboxOpts: {
    swDest: path.resolve(__dirname, './public/service-worker.js'),
  }
```

krish-dev:
@ashbuilds absolutly we can use `public` dir,  In that case, service-worker might get installed successfully. But as the request not get routed through lambda so service worker will not get valid permission (`AccessDenied`) while accessing the internal js file and it will lead to following issue

```javascript
service-worker.js:1 Uncaught (in promise) bad-precaching-response: bad-precaching-response :: [{"url":"https://xxxxxxxxxx.cloudfront.net/_next/static/chunks/6b2a94c785d9e2be08deadc3973900fa4d342531.60755c98fab970be950f.js","status":403}]
    at E.K (https://xxxxxxxxxx.cloudfront.net/service-worker.js:1:11535)
    at async Promise.all (index 5)
    at async E.install (https://xxxxxxxxxx.cloudfront.net/service-worker.js:1:10954)
```
![Screenshot 2020-06-02 at 10 16 21 AM](https://user-images.githubusercontent.com/7298557/83482655-6d236880-a4be-11ea-9aac-e78257133d0b.png)

I'm still using this configuration 
```javascript
workBoxConfig.workboxOpts['swDest'] = 'static/service-worker.js';
```
[...]

PR #426, opened 2020-06-03
fix support for next/dynamic import

krish-dev:
#236 
Just added one condition into `copyLambdaHandlerDependencies` method to coping chunks into the valid directory. [Ref](https://github.com/danielcondemarin/serverless-next.js/issues/236#issuecomment-637912442) 

#357 
also added service-worker support.

### DIFF
--- a/packages/lambda-at-edge/src/build.ts
+++ b/packages/lambda-at-edge/src/build.ts
@@ -14,6 +14,8 @@ import pathToPosix from "./lib/pathToPosix";
 import expressifyDynamicRoute from "./lib/expressifyDynamicRoute";
 import pathToRegexStr from "./lib/pathToRegexStr";
 import createServerlessConfig from "./lib/createServerlessConfig";
+import isChunk from "./lib/isChunk";
+import pathToFilename from "./lib/pathToFilename";
 
 export const DEFAULT_LAMBDA_CODE_DIR = "default-lambda";
 export const API_LAMBDA_CODE_DIR = "api-lambda";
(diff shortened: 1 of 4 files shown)


danielcondemarin:
@krish-dev Thanks for the PR and being on top of this issue! However I've already approved and merged https://github.com/danielcondemarin/serverless-next.js/pull/425. Sorry that this wasn't communicated properly on the issue on you both ended up working on the same thing. 

With regards to the `service-worker` fix, I'd like to have a solution that works for any filename rather than hardcoding `service-worker`. 
My preferred solution would be to allow users to [add a custom cache behaviour](https://github.com/danielcondemarin/serverless-next.js#custom-cloudfront-configuration). Most of the work is already done, however I suspect there is a bit more to do to support your use case. Let me know if you want to take a stab at this and I'm happy to assist where I can.

krish-dev:
> @krish-dev Thanks for the PR and being on top of this issue! However I've already approved and merged #425. Sorry that this wasn't communicated properly and you both ended up working on the same thing.

No worries, Your current release (`serverless-next.js@1.12.0-alpha.8`) works perfectly.



> With regards to the `service-worker` fix, I'd like to have a solution that works for any filename rather than hardcoding `service-worker`.
> My preferred solution would be to allow users to [add a custom cache behaviour](https://github.com/danielcondemarin/serverless-next.js#custom-cloudfront-configuration). Most of the work is already done, however I suspect there is a bit more to do to support your use case. Let me know if you want to take a stab at this and I'm happy to assist where I can.

I'm agreed with you and I'm also against hardcoding, It was just a quick fix by considering `service-worker` is a special case.  Yes, I would like to take a stab at this, I will appreciate if you just give me a hint that what you suspect. If you wish you can create a separate communication thread by opening an RFC or you continue with the issue #357

PR #431, opened 2020-06-07
update next.js version to 9.4.4, minor typo fixes

### LINKED ISSUE #401: Should follow the Next.js environment variable file pattern
https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables

The .env file has the least precedence and should have its variables be overriddable by .local and environment-specific files. The build produced by running `npx serverless` with this component currently ignores all .env* files other than .env

I'm currently keeping my secrets for dev in .env.local. Before I want to deploy I copy the variables from .env.production.local (gitignored by nextjs by default) into .env and revert the change after the deployment is done. This is not the best experience. 

Environment variables should be be retrievable from other .env* files, not just .env.

lone-cloud:
fixes #401 

**The goal**
Improve the development experience when working on local and deployment environments that use a different set of environment variables. As of Next.js 9.4, the recommended way to handle this is: https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables
The direction seems to be to have the non-sensitive (ex. public) keys committed with .env(.development|.production)? files and have the sensitive keys stored locally in .env*.local files.

**Impact**
The examples might no longer work well because they import dotenv which is now a naughty thing to do: https://github.com/vercel/next.js/blob/2ad351511ed6fb44161f42ae401a19b37cf131e3/packages/next/lib/load-env-config.ts#L27
My personal opinion is that the "multiple-instance-environment" example might be an anti-pattern as anyone with such a setup should be strongly considering continuous deployment instead of locally running deployments to multiple envs.

@danielcondemarin you might get some value out of:
- https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker
- setting up husky with pre-commit hooks for prettier. I like to use it with the following package.json config:
```
"husky": {
    "hooks": {
      "pre-commit": "lint-staged"
    }
  },
  "lint-staged": {
    "*.{js,jsx,ts,tsx,md}": "prettier --write"
  }


lone-cloud:
> 
> 
> Looks good thanks 🙏
> 
> Feel free to add the lint-staged section to the package.json btw 👍🏻 I should’ve added it before.

Added using: https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged

PR #432, opened 2020-06-07
return a static 404 page when a route cannot be found

### LINKED ISSUE #378: 404 Error Page not working
**Describe the bug**
When using `next@9.3.5`, navigating to an invalid URL does not show the 404 error page. Instead, you see 200 | An unexpected error has occurred.

**To Reproduce**
1. Clone this repository: [https://github.com/sidarthar/serverless-nextjs-404-error](https://github.com/sidarthar/serverless-nextjs-404-error)
2. Run `npx serverless` to deploy 
3. After deployed, visit the appUrl/pageThatDoesNotExist

**Expected behavior**
Should see a 404 | Page Not Found error.

**Screenshots**
image

 - OS: N/A
 - Version next@9.3.5, serverless-next.js@1.11.0

**Additional context**
When using `next@9.3.4`, the 404 page works as expected.

lone-cloud:
fixes #378 

According to https://nextjs.org/docs/advanced-features/custom-error-page#404-page the 404 page is "statically generated at build time". As such, it cannot be SSR'ed and is expected to be returned when a route cannot be found. Whereas _error.js can only be SSR'ed and is expected to handle other kinds of errors.

**Impact**
This approach will be a slight regression for those looking to localize their 404 pages. On the server side, one could previously read the language cookie (or something similar) and set the page locale based on that. Always returning a static 404.html page means that a client-side implementation will be required which is going to feel bad.... or maybe don't even worry about localizing 404 pages. Next.js does not currently provide a good way to localize apps. :(


danielcondemarin:
Thanks again 🙌

I wonder if we could make this change backward compatible the same way [next announced ](https://nextjs.org/blog/next-9-3#automatic-static-optimization-for-404) it,


> This feature is fully backwards compatible so if you currently have a custom pages/_error.js it will continue to be used for the 404 page until you add pages/404.js.

Also, do you know if 404.html is always produced on build, even when users are still using `_error.js` for 404s?

lone-cloud:
> 
> 
> Thanks again 🙌
> 
> I wonder if we could make this change backward compatible the same way [next announced ](https://nextjs.org/blog/next-9-3#automatic-static-optimization-for-404) it,
> 
> > This feature is fully backwards compatible so if you currently have a custom pages/_error.js it will continue to be used for the 404 page until you add pages/404.js.
> 
> Also, do you know if 404.html is always produced on build, even when users are still using `_error.js` for 404s?

Looks like it is not produced in that case. The build output suggests that for that scenario (_error defined, but no 404) the 404 page should be SSR'ed from a lambda. damn 
I think the low-hanging fruit solution is what you suggested in the PR: https://github.com/lone-cloud/serverless-next.js/commit/834886a41324500a5c29831d3614366b08725c5f#diff-867646f333d11e3bbfc6d2da7ffb967f
The more idiomatic Next.js way seems to be to conditionally return error.js if 404.html was not created, but I'm not 100% how that implementation would look like here.
Thoughts?

danielcondemarin:
> > Thanks again 🙌
> > I wonder if we could make this change backward compatible the same way [next announced ](https://nextjs.org/blog/next-9-3#automatic-static-optimization-for-404) it,
> > > This feature is fully backwards compatible so if you currently have a custom pages/_error.js it will continue to be used for the 404 page until you add pages/404.js.
> > 
> > 
> > Also, do you know if 404.html is always produced on build, even when users are still using `_error.js` for 404s?
> 
> Looks like it is not produced in that case. The build output suggests that for that scenario (_error defined, but no 404) the 404 page should be SSR'ed from a lambda. damn
> I think the low-hanging fruit solution is what you suggested in the PR: [lone-cloud@834886a#diff-867646f333d11e3bbfc6d2da7ffb967f](https://github.com/lone-cloud/serverless-next.js/commit/834886a41324500a5c29831d3614366b08725c5f#diff-867646f333d11e3bbfc6d2da7ffb967f)
> The more idiomatic Next.js way seems to be to conditionally return error.js if 404.html was not created, but I'm not 100% how that implementation would look like here.
> Thoughts?

In the default-handler you could check the `manifest.json` for a 404.html page entry, if there is one use it, otherwise fall back to _error.js. 

This is an example `.serverless_nextjs/default-lambda/manifest.json` where 404.html is available:
{
  "buildId": "Sl-G_UYjxcrPfbonpj6nd",
  "pages": {
    "ssr": {
      "dynamic": {},
      "nonDynamic": {
        "/_error": "pages/_error.js",
        "/todos": "pages/todos.js",
        "/_app": "pages/_app.js",
        "/_document": "pages/_document.js"
      }
    },
    "html": {
      "dynamic": {},
      "nonDynamic": {
        "/todos/new": "pages/todos/new.html",
        "/404": "pages/404.html"
      }
    }
  },
  "publicFiles": {}
}

PR #435, opened 2020-06-08
Cache static pages and invalidate distribution after deploy

### LINKED ISSUE #430: Static HTML Pre-rendered pages are not cached in Cloudfront
**Describe the bug**
Whenever I request for html file inside static-pages, cloudfront hits the origin and returns Miss-From-Cloudfront header. Not sure it was a expected behaviour, hitting origin each time for a static html and having higher TTFB.

**To Reproduce**
Request for any html inside static-pages, inspect the response header for the html. I have been using out of box solution provided

**Expected behavior**
Request should Hit cache from cloudfront. These are static pages should be served from edge, shouldn't send each requests to s3 origin.

**Screenshots**
Screenshot 2020-06-07 at 4 01 41 PM

thiagozf:
### Problem
As reported by #430, static pages are not cached by CloudFront and client requests are always hitting the origin.

### Expected behaviour
Static pages should be automatically cached at the edge after the first request for an arbitrary long period of time. When a new deployment occurs, the CloudFront cache should be invalidated, ensuring that the latest version is served to clients. This gives us the most flexibility, as users get the latest file from CDN immediately after deploying.

### Solution
- Instead of using `Cache-Control: no-store` for static pages, the component now uses `public, max-age=0, s-maxage=86400, must-revalidate`. This prevents the page from being cached in the browser, but `s-maxage` allows the page to be cached at edge locations for 1 day (86400 seconds). 
- An invalidation request for all paths (`/*`) is created after each deployment, thus ensuring that the latest version is always served to clients while using the CDN effectively.

### Implementation details
- A helper package, called `@sls-next/cloudfront`, was created to handle cache invalidation. We could also expose an invalidation method on `@serverless/aws-cloudfront` component, similar to `@serverless/aws-s3` [upload](https://github.com/serverless-components/aws-s3/blob/master/serverless.js#L109) helper method to upload files, but Components `v1` is no longer maintained and we would need to get it merged.
- Timestamp was used as [`CallerReference`](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_InvalidationBatch.html) for the cache invalidation request. We could also use the Next.js BUILD_ID.
- An arbitrary cache period of 1 day was used. We could increase this for a longer time ([Vercel uses 31 days](https://vercel.com/docs/v2/edge-network/caching#static-files)).

resolves #430

### DIFF
--- a/packages/serverless-component/__mocks__/@sls-next/cloudfront.js
+++ b/packages/serverless-component/__mocks__/@sls-next/cloudfront.js
@@ -1,0 +1,3 @@
+module.exports = {
+  default: jest.fn()
+};
--- a/packages/cloudfront/CHANGELOG.md
+++ b/packages/cloudfront/CHANGELOG.md
@@ -1,0 +1,4 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
--- a/packages/cloudfront/README.md
+++ b/packages/cloudfront/README.md
@@ -1,0 +1,1 @@
+### AWS CloudFront helper library
--- a/packages/cloudfront/package-lock.json
+++ b/packages/cloudfront/package-lock.json
@@ -1,0 +1,108 @@
+{
+  "name": "@sls-next/cloudfront",
+  "version": "1.0.0-alpha.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "aws-sdk": {
+      "version": "2.691.0",
+      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.691.0.tgz",
+      "integrity": "sha512-HV/iANH5PJvexubWr/oDmWMKtV/n1shtrACrLIUa5vTXIT6O7CzUouExNOvOtFMZw8zJkLmyEpa/0bDpMmo0Zg==",
+      "requires": {
+        "buffer": "4.9.2",
+        "events": "1.1.1",
+        "ieee754": "1.1.13",
+        "jmespath": "0.15.0",
+        "querystring": "0.2.0",
+        "sax": "1.2.1",
+        "url": "0.10.3",
+        "uuid": "3.3.2",
+        "xml2js": "0.4.19"
+      }
+    },
+    "base64-js": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.3.1.tgz",
+      "integrity": "sha512-mLQ4i2QO1ytvGWFWmcngKO//JXAQueZvwEKtjgQFM4jIK0kU+ytMfplL8j+n5mspOfjHwoAg+9yhb7BwAHm36g=="
+    },
+    "buffer": {
+      "version": "4.9.2",
+      "resolved": "https://registry.npmjs.org/buffer/-/buffer-4.9.2.tgz",
+      "integrity": "sha512-xq+q3SRMOxGivLhBNaUdC64hDTQwejJ+H0T/NB1XMtTVEwNTrfFF3gAxiyW0Bu/xWEGhjVKgUcMhCrUy2+uCWg==",
+      "requires": {
+        "base64-js": "^1.0.2",
+        "ieee754": "^1.1.4",
+        "isarray": "^1.0.0"
+      }
+    },
+    "events": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/events/-/events-1.1.1.tgz",
+      "integrity": "sha1-nr23Y1rQmccNzEwqH1AEKI6L2SQ="
+    },
+    "ieee754": {
+      "version": "1.1.13",
+      "resolved": "https://registry.npmjs.org/ieee754/-/ieee754-1.1.13.tgz",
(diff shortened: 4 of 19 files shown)


packages/serverless-component/__mocks__/@sls-next/cloudfront.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
danielcondemarin, line 1: Could we avoid mocking `@sls-next/cloudfront`. I think this is implementation detail that we shouldn't be mocking even if is a different package.

danielcondemarin:
@thiagozf I really like the approach you've taken. Just one comment regarding the tests but the rest look good. I'll be testing it shortly!

danielcondemarin:
@thiagozf Is there anything else you wanted to do in this PR? From my pov looks good. I've tested it today and is working perfectly. I want to release it asap in an alpha release 🚀

thiagozf:
@danielcondemarin Not really, just wanted you to check if I was missing something.

packages/serverless-component/__mocks__/@sls-next/cloudfront.js
thiagozf: We cover AWS API calls with tests for the package ([`cache-invalidation.test.ts`](https://github.com/danielcondemarin/serverless-next.js/blob/ac809ee970981c9b9e3229cb29cd0885566b74e9/packages/cloudfront/tests/cache-invalidation.test.ts)). Do you think we should also mock the `aws-sdk` directly inside `serverless-component`? If so, should I just add a CloudFront mock inside `serverless-plugin/__mocks__/aws-sdk.js` or use a custom mock?

danielcondemarin: Ah I see aws-sdk isn't mocked in the serveless-component. Nevermind, this is good for now, I'll look at tidying the tests up a bit later 👍

lucas-varela:
Just as a heads up: this change broke my build process because my CI user did not have permission to create invalidations.

Fixing was easy, though, I simply added the "cloudfront:CreateInvalidation" permission to the CI user on IAM.

danielcondemarin:
> Just as a heads up: this change broke my build process because my CI user did not have permission to create invalidations.
> 
> Fixing was easy, though, I simply added the "cloudfront:CreateInvalidation" permission to the CI user on IAM.

I missed this. Would've been better to publish as a breaking change. What error did you get? I think is worth documenting this in the README somewhere.

lucas-varela:
The error:

```
DEBUG ─ Updating CloudFront distribution of ID <distribution-id>.
DEBUG ─ CloudFront deployed successfully with URL: https://<distribution>.cloudfront.net.
error:
  AccessDenied: User: arn:aws:iam::<account-id>:user/bitbucket is not authorized to perform: cloudfront:CreateInvalidation on resource: arn:aws:cloudfront::<account-id>:distribution/<distribution-id>
    at Request.extractError (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/protocol/rest_xml.js:53:29)
    at Request.callListeners (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/request.js:685:12)
    at Request.callListeners (/root/.serverless/components/registry/npm/@starlightcms/trailblazer@1.2.0/node_modules/aws-sdk/lib/sequential_executor.js:116:18)
{
  message: 'User: arn:aws:iam::<account-id>:user/bitbucket is not authorized to perform: cloudfront:CreateInvalidation on resource: arn:aws:cloudfront::<account-id>:distribution/<distribution-id>',
  code: 'AccessDenied',
  time: 2020-06-16T20:31:46.463Z,
  requestId: '<id>',
  statusCode: 403,
  retryable: false,
  retryDelay: 11.447611865861518
}
```

Redacted some info for obvious reasons.

kartikag01:
An invalidation request for all paths (/*) currently invalidating all CloudFront behaviors cache,
I have `/images`, `/json`, `/api`, and many more behaviors, which I don't want to invalidate but everything is removed from the cache once `/*` is run.

I don't know how not to invalidate other behavior, but it's better to give control to the user when he is not using SSG
I have created the PR. 
https://github.com/serverless-nextjs/serverless-next.js/pull/603

PR #451, opened 2020-06-17
Add support for custom s3 region

danielcondemarin:
In order to minimise latency when CloudFront proxies a request to S3, a new input will be added (`bucketRegion`) to set which region you want your bucket to be hosted. 

Example:

```yml
# serverless.yml
component: serverless-next.js@...
inputs:
 bucketRegion: "eu-west-1" // defaults to us-east-1
```


packages/serverless-component/serverless.js
```diff
@@ -180,7 +180,8 @@ class NextjsComponent extends Component {
 
     const bucketOutputs = await bucket({
       accelerated: true,
-      name: inputs.bucketName
+      name: inputs.bucketName,
+      region: inputs.s3BucketRegion || "us-east-1"
```
lone-cloud, line 184: needs a README update?

danielcondemarin: Good catch. Done that now

PR #452, opened 2020-06-18
document the full list of AWS permissions required for a deployment

### LINKED ISSUE #280: What permissions are required to deploy?
When I deploy using this component, I get an access denied error from AWS. This was expected because I'm running it as a user with very restrictive permissions. Do you have any information regarding what permissions are required?

These are the permissions that I've determined are necessary so far:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": "s3:HeadBucket",
            "Resource": "*"
        },
        {
            "Sid": "VisualEditor1",
            "Effect": "Allow",
            "Action": [
                "s3:PutAccelerateConfiguration",
                "s3:CreateBucket",
                "s3:ListBucket"
            ],
            "Resource": "arn:aws:s3:::bucket-name"
        },
        {
            "Sid": "VisualEditor2",
            "Effect": "Allow",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::bucket-name/*"
        }
    ]
}
```

However, then I come across the following error:

```
AccessDenied: User: <user_arn> is not authorized to perform: iam:GetRole on resource: role mzppeul-0a646kc
```

Because this is an auto-generated name with no way of configuring the name, I can't give very fine-grained permissions for this.

lone-cloud:
fixes #280 

I think it's invaluable to keep this info in the README. At least where I work, full admin access to AWS isn't given out to everybody and I had to come up with a more granular list listed here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,7 @@ The exhaustive list of AWS actions required for a deployment:
   "route53:ListResourceRecordSets",   // only for custom domains
   "s3:CreateBucket",
   "s3:GetAccelerateConfiguration",
+  "s3:GetObject",                     // only if persisting state to S3 for CI/CD
   "s3:HeadBucket",
   "s3:ListBucket",
   "s3:PutAccelerateConfiguration",
--- a/README.md
+++ b/README.md
@@ -217,6 +217,51 @@ myNextApplication:
 
 Make sure you add CloudWatch log permissions to your custom policy.
 
+The exhaustive list of AWS actions required for a deployment:
+
+```
+  "acm:DescribeCertificate", // only for custom domains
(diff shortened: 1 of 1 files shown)


danielcondemarin:
This is awesome @lone-cloud ! It may be even better to provide a yaml policy users can just copy and use, what do you think?

lone-cloud:
I'm not primarily a dev-ops engineer, so I don't know the most popular way to configure things. All I can say is that I put JSON (array) here because that's the way the IAM in AWS console accepts it. Also the dev-ops guys at my current gig are defining the user policies in terraform which also works with JSON.

PR #460, opened 2020-06-22
update all deps, don't overwrite the cloudfront default.forward config

### LINKED ISSUE #397: queryString forwarding stopped - recent "allow custom configuration for Cloudfront" PR broke our Cloudfront distribution
Our cloudfront distribution recently broke due to queryStrings suddenly not being forwarded.  (Resulting in getInitialProps `ctx.query` always being empty object). This forwarding behaviour was previously configured through aws GUI when we initially set it up and wasnt changed. However a deploy today suddenly switched the forwarding off due to it not being specified in our `serverless.yml`. (we were also not pinning the serverless-next.js version which we have now done)

I believe this was caused by the update here: 
https://github.com/danielcondemarin/serverless-next.js/commit/e7bed6aa8d0a8f7fd332e1d0c37d8dc67ec3aa65#diff-04c6e90faac2675aa89e2176d2eec7d8R128

Our package json was:
` "serverless-next.js": "^1.9.10",` 

Our serverless.yml was:

```

myNextApplication:
  component: serverless-next.js
  inputs:
    domain: [**hidden**]
    cloudfront:
      defaults: # options for lambda that handle SSR
        forward:
          headers: [**hidden**]
    env:
      NODE_ENV: production
```

In order to fix updated to:
(with version now pinned)
```

myNextApplication:
  component: serverless-next.js@1.9.10
  inputs:
    domain: [**hidden**]
    cloudfront:
      defaults: # options for lambda that handle SSR
        forward:
          headers: [**hidden**]
          queryString: true
  build:
    env:
      NODE_ENV: production
```

lone-cloud:
better prerenderedManifest typing, one more AWS permission that's used when trying to rename lambdas

fixes #397


packages/serverless-component/serverless.js
```diff
@@ -406,12 +403,12 @@ class NextjsComponent extends Component {
     const cloudFrontOutputs = await cloudFront({
       defaults: {
         ttl: 0,
+        ...defaultCloudfrontInputs,
```
lone-cloud, line 406: I believe this is the fix for #397

PR #462, opened 2020-06-24
enforce compression for the default cloudfront behaviour

lone-cloud:
Without this the cloudfront default behaviour doesn't set compression to true which causes all the responses to the initial page requests to not be gziped. This wastes bandwidth, but worst of all, it brings the Lighthouse performance score down.

Not sure why there's such a large change to the lock file. All I did was run 'npm i'.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6819,6 +6819,30 @@
             "path-to-regexp": "*"
           }
         },
+        "@zeit/node-file-trace": {
+          "version": "0.6.5",
+          "resolved": "https://registry.npmjs.org/@zeit/node-file-trace/-/node-file-trace-0.6.5.tgz",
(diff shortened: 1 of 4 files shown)


lone-cloud:
I'm running two environments that are basically identical. One is being deployed with the latest alpha and the other is deploying my dev branches (this one). Making this change absolutely solved the problem of the initial page not being gziped, but I cannot explain why. @danielcondemarin do you know if it's possible to debug the lambda handler code locally and with an interactive debugger?
I'm much more used to working with Next where I debug using `cross-env NODE_OPTIONS='--inspect' next dev`. Wonder if I can do something similar with serverless components.

danielcondemarin:
@danielcondemarin do you know if it's possible to debug the lambda handler code locally and with an interactive debugger?

I don't think is possible to attach an inspector to a remote aws lambda, nevermind lambda@edge. However the newer version of serverless components has a `serverless dev` (https://www.serverless.com/blog/serverless-components-ga/) command which should help here by streaming console.logs from CloudWatch to your CLI in real time. Not sure if that will work out of the box with Lambda@Edge but worth keeping in mind for when we upgrade.

PR #466, opened 2020-06-25
default the security policy to TLSv1.2, add support for naked domains, allow verbose logging on deployment

### LINKED ISSUE #465: feature: pull in @serverless/domain into serverless-next.js to be able to configure it better
That library is hiding various important configuration that should be exposed to the users of serverless-nextjs. Note that it doesn't appear that, that @serverless/domain is being maintained anymore: https://github.com/serverless-components/domain

Proposal:
- pull in @serverless/domain. Note that this will obliterate the code coverage % as that library comes with no tests.
- update the MinimumProtocolVersion default to the latest, recommended TLSv1.1_2018
- ~~add new "domain" inputs:  MinimumProtocolVersion, PriceClass and DomainType~~ can already be done with cloudfront input but non-default MinimumProtocolVersion will be configurable
- DomainType input can be one of 3: 
'apex' - if only a domain is provided, don't create a www subdomain.
'www' - if a 'www' subdomain is provided, don't create an apex subdomain.
'both' (default) - current behaviour

For the 'apex' and 'www' cases I'm thinking of creating CNAME DNS records to correct the routing for the unsupported case. Not sure how well that will work.

Let me know how this sounds @danielcondemarin

lone-cloud:
fixes #465 

new stuff:
- bringing in the @serverless/domain package into the monorepo to be able to support and extend its functionality. Only the relevant (cloudfront) parts have been copied over to this project to reduce confusion and code coverage drop.
- CloudFront will now default to the latest TLSv1.2_2018 security policy instead of TLSv1.1_2016 when using a custom domain
- adding a new "domainType" input to allow for "apex" (aka "naked"), "www" or "both" types of domains. This input will not implement the 301 redirection from the unsupported domain when using apex/www.
- skipping a flaky test in lambda-at-edge

Problems and TODOs:
- feelsbadman about the test coverage, but what to do?
- there's no easy way to redirect requests to the unsupported subdomain when in "www" or "apex" mode. The suggested solution that I've been reading about is to configure a new S3 bucket as described: https://aws.amazon.com/premiumsupport/knowledge-center/redirect-domain-route-53/


danielcondemarin:
@lone-cloud LGTM. Let me know if you need help 👍

lone-cloud:
@danielcondemarin this is ready for review

danielcondemarin:
@lone-cloud There is quite a lot going on in this PR. Could some of those TODOs be split onto separate PRs using an incremental approach?

PR #468, opened 2020-06-26
don't forward cookies, headers and query parameters for requests to static paths

### LINKED ISSUE #459: Cookies shouldn't be forwarded for static files
_Edit_: I'm completely rewriting the original issue.

Static file routes on CloudFront get created with the Query String and Cookies being forwarded to the S3 origin. This renders caching inefficient. Let's change it to not forward anything. I'd be happy to send a PR if you point me to where the config is in the code.

lone-cloud:
fixes #459


katspaugh:
Wow, many thanks!

Issue #469, opened 2020-06-27
access denied on CloudFront website after setting bucketRegion 


yellowtailfan:
**Describe the bug**
When I set the bucketRegion to ap-southeast-2, I got an access denied page when I visited the CloudFront URL.

**To Reproduce**

1. Run `npx create-next-app` and change directory into the project folder
2. Add `serverless.yml` to the project folder
3. Run `sls`
4. Wait for the AWS CloudFront console to show the new distribution has finished deploying
5. Visit appUrl printed in terminal
6. Web browser shows a CloudFront access denied error page

`serverless.yml` contents:
```
nextTest:
  component: serverless-next.js
  inputs:
    bucketRegion: ap-southeast-2
```

**Expected behavior**
The NextJS sample website should appear in the browser.

**Work around**
I used the AWS CloudFront console to change the distribution origin from:
```
<bucket-name>.s3.amazonaws.com
```
to:
```
<bucket-name>.s3-ap-southeast-2.amazonaws.com
```
then waited for the distribution to deploy, and the sample website now comes up fine in the browser. The [AWS docs ](https://aws.amazon.com/premiumsupport/knowledge-center/s3-http-307-response/) also say this error clears by itself if you wait 24 hours after creating a new distribution pointing to an S3 bucket in a custom region.

The serverless-next.js README says 307 redirects may happen when you use a bucket with a different region, but the README doesn't say anything about access denied errors happening in the first 24 hours.

yellowtailfan:
I made a simple change to the sample website text and ran `sls` to deploy the change and got access denied again. The deployment had stripped the region name back out of the CloudFront origin. I put it back in and the sample website works again including the text change. 

So it seems the work around has to be reapplied after each `sls`.

lone-cloud:
Hey, thanks for the report and your observations are on point. https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-component/serverless.js#L197 needs to be fixed and the bucketUrl below it. Right now it defaults it to s3 (us-east) instead of being region-specific.

yellowtailfan:
It looks like this is still an issue in 1.15.1. And according to your PR comment (https://github.com/serverless-nextjs/serverless-next.js/pull/474#issuecomment-652077380), it depends on an upstream fix in the cloudfront component?

danielcondemarin:
[...]

PR #474, opened 2020-06-29
Use S3 regional endpoint when not in us-east-1

danielcondemarin:
Fixes https://github.com/serverless-nextjs/serverless-next.js/issues/469


packages/lambda-at-edge/src/default-handler.ts
```diff
@@ -67,6 +67,20 @@ const router = (
...
 const normaliseUri = (uri: string): string => (uri === "/" ? "/index" : uri);
+const normaliseS3OriginDomain = (s3Origin: CloudFrontS3Origin): string => {
+  if (s3Origin.region === "us-east-1" || !s3Origin.region) {
+    return s3Origin.domainName;
+  }
+
+  if (!s3Origin.domainName.includes(s3Origin.region)) {
+    return s3Origin.domainName.replace(
+      "s3.amazonaws.com",
+      `s3-${s3Origin.region}.amazonaws.com`
```
lone-cloud, line 78: We shouldn't use this pattern according to: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro
`s3.${s3Origin.region}.amazonaws.com`

danielcondemarin: good catch, updated!

danielcondemarin:
I was hoping as part of this PR to use the regional endpoint in for the CloudFront S3 origin but the aws-cloudfront component [strips the region out](https://github.com/serverless-components/aws-cloudfront/blob/master/lib/getOriginConfig.js#L21) :/ I'll revisit once aws-cloudfront is pulled into the repo. 

In the meantime will get this merged which allows lambda@edge to use regional endpoints when needed.

PR #485, opened 2020-07-02
Update README.md to document stable npm package

Gabrola:
Anyone landing on the github repo package will be directed to use the unusable npm package @sls-next/serverless-component which has no stable release yet.

Should update the README to point to that package only when a stable release is made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Add your next application to the serverless.yml:
 # serverless.yml
 
 myNextApplication:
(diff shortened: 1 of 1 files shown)


lone-cloud:
I don't think this will be necessary. Ideally we want everyone to start testing and using the latest version. I believe the current alpha is stable and I run daily deploys with it: 
`component: "@sls-next/serverless-component@1.15.0-alpha.4"`

danielcondemarin:
Yeah I second @lone-cloud. Also the alpha is not far from being merged to stable now :)

Issue #496, opened 2020-07-11
Public asset throws 403 SignatureDoesNotMatch when bucket region is not N. Virginia


botond-veress:
**Describe the bug**
I use this lib for a while now, and it always worked like a charm. Good job guys!

But lately, I noticed on new projects that public assets (favicon, robots.txt, basically anything in the public folder) cannot be loaded because I get a Cloudfront/S3 SignatureDoesNotMatch error. I couldn't find anybody reporting the same issue so I started digging deeper.

**To Reproduce**

1. Set a bucket region other than N. Virginia
2. Deploy
3. Any file that was placed in the /public directory cannot be loaded.

P.S I could reproduce the issue in "@sls-next/serverless-component@1.15.0-alpha.5" and "serverless-next.js@1.9.10" so most likely more versions are affected.

**Expected behavior**
I should receive the file I requested instead of the SignatureDoesNotMatch error.

**Screenshots**
```
<Error>
<Code>SignatureDoesNotMatch</Code>
<Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
<AWSAccessKeyId>{REDACTED}</AWSAccessKeyId>
<StringToSign>AWS4-HMAC-SHA256 20200711T085322Z 20200711/eu-central-1/s3/aws4_request ae2db368616c3ee5907fb7d3b5851ebaff0bc2ea7f4f14f07126b5668b62c302</StringToSign>
<SignatureProvided>9416a27e3980602a06eb7f92c1928f7884ce4fbfdbcaad72bf998f7906e07bc1</SignatureProvided>
<StringToSignBytes>41 57 53 34 2d 48 4d 41 43 2d 53 48 41 32 35 36 0a 32 30 32 30 30 37 31 31 54 30 38 35 33 32 32 5a 0a 32 30 32 30 30 37 31 31 2f 75 73 2d 65 61 73 74 2d 31 2f 73 33 2f 61 77 73 34 5f 72 65 71 75 65 73 74 0a 61 65 32 64 62 33 36 38 36 31 36 63 33 65 65 35 39 30 37 66 62 37 64 33 62 35 38 35 31 65 62 61 66 66 30 62 63 32 65 61 37 66 34 66 31 34 66 30 37 31 32 36 62 35 36 36 38 62 36 32 63 33 30 32</StringToSignBytes>
[...]

PR #497, opened 2020-07-11
chore: fix s3 bucket not being normalized for public assets

botond-veress:
Fixes https://github.com/serverless-nextjs/serverless-next.js/issues/496


danielcondemarin:
Thanks! You mind adding a quick test [here](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/lambda-at-edge/tests/default-handler/default-handler.test.ts)

PR #498, opened 2020-07-11
chore(all): bump aws-sdk versions across all packages so it won't fail to install randomly

botond-veress:
Fixes lerna exec mentioned in https://github.com/serverless-nextjs/serverless-next.js/issues/496


danielcondemarin:
Closing as https://github.com/serverless-nextjs/serverless-next.js/pull/509 should fix this already.

Issue #494, opened 2020-07-10
BasePath support


rrooding:
I'm trying to use the Next.js basePath option (https://github.com/vercel/next.js/issues/4998) with serverless-next.js but I'm having some problems.

Please let me know if I'm missing something but as far as I can see this is not supported yet.

I am happy to contribute a PR but would like to discuss the implementation first. My suggestion is to read the basePath setting from the next.config.js file during deployment and, when configured, use that for the following:

* Change uploadAssetsToS3 to upload to `<basePath>/` subdirectory in the bucket
* Change the Cloudfront origins to `<basePath>/_next/static/*`, `<basePath>/static/*`, etc..

Anything else?

danielcondemarin:
@rrooding A couple of thoughts from me on the approach:

- Rather than reading from the next.config.js file its easier to read from `.next/routes-manifest.json` which contains the `basePath` property. 

- I'm not sure you actually need to a new directory in the s3 bucket. As long as the CloudFront cache behaviour maps `<basePath>/_next/static/*` -> s3 _next/static/* should be OK I think. Only saying because it may be simpler to implement. 

FYI - I've raised an RFC for multi-zones like support [here](https://github.com/serverless-nextjs/serverless-next.js/issues/477) that is sort of related. 

Feel free to start a PR and I'll help out with approach and testing.

rrooding:
I was on holiday, but just got to fixing the last failure, if you could have a look at it again 👍

rrooding:
PR is merged

Issue #479, opened 2020-07-01
Trailing slashes on root pages return 404


medv:
**Is your feature request related to a problem? Please describe.**
Accessing root pages with a trailing slash (often how links get formatted in applications and sites we push them out to) returns the 404 page.

**Describe the solution you'd like**
I can see that Next@9.4.5, which is currently in canary, solves or at least addresses this problem, so hopefully this is something that can be maintained here as well, if the next update brings with it some additional resources to manage by this component.

**Describe alternatives you've considered**
We have tried supplying a set of hard redirects in next.config.js using "exportPathMap", which works locally but not when deployed using the serverless-next.js component. We have considered adding another lambda@edge to the CloudFront inputs that can trim the slash before landing into the component's routing logic, however that would increase latency. We have also considered just re-uploading the routing logic and altering the `normaliseUri` function to trim the trailing slash before the page manifest is searched by the uri, which is a wooden solution obviously (second to forking this project for altering that one function).

Thank you for any insight on this! Any stopgap ideas would be very appreciated as well.

Relevant pull request here https://github.com/serverless-nextjs/serverless-next.js/pull/448

lone-cloud:
The immediate solution would indeed be to fork this project and merge #448 in. You can then deploy from local serverless-components (look at CONTRIBUTING.md for reference).
I see a PR was merged for Next 9.4.5 to add ```experimental: {
    trailingSlash: true,
  },```. I see this was only implemented to be read on the next-server and not something that would be reflected in the manifests (AFAIK) which is how this lib determines routing. If @danielcondemarin is good with it, I think we can prematurely start implementing this feature now. We can check if the experimental flag was set in the user's next.config.js and do the stuff from #448 if it's set to true.

danielcondemarin:
I'd normally recommend holding off merging until it becomes stable. Having said that if someone wants to take a stab and start putting together a PR go for it 👍

medv:
Ah of course! That bit about the separation between server and manifest routing makes sense. Thanks @lone-cloud 

[...]

PR #510, opened 2020-07-15
BasePath support

rrooding:
Add basePath support as per #494


packages/serverless-component/__tests__/basepath.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    consoleWarnSpy.mockRestore();
+    fse.remove.mockRestore();
+    process.chdir(tmpCwd);
+  });
+
+  afterAll(cleanupFixtureDirectory(fixturePath));
+
+  describe("cloudfront", () => {
+    it("creates distribution", () => {
+      expect(mockCloudFront).toBeCalledWith({
```
danielcondemarin, line 73: Could we change the assertion so we only test things affected by basePath? e.g. 

```
const cloudFrontCreateArgs = mockCreateCloudFront.mock.calls[0][0];
const pathPatterns = cloudFrontCreateArgs.origins[0].pathPatterns;
expect(Object.keys(pathPatterns)).toEqual("custom_base/_next/static/*", "custom_base/_next/data/*", ...)
```

packages/serverless-component/src/component.ts
```diff
@@ -42,9 +43,26 @@ class NextjsComponent extends Component {
...
+  }
+
+  pathPattern(pattern: string, routesManifest: RoutesManifest): string {
+    const basePath = routesManifest.basePath;
+    if (basePath.length === 0) {
+      return pattern;
+    }
+
+    const basePathPattern =
+      basePath.charAt(0) === "/" ? basePath.slice(1) : basePath;
```
danielcondemarin, line 57: Out of curiosity why do we need this? Does next.js allow for base paths with and without `/`?

rrooding: It doesn't :) Fixed

README.md
```diff
@@ -354,30 +354,30 @@ The fourth cache behaviour handles next API requests `api/*`.
...
+| bucketName               | `string`          | `null`                                                             | Custom bucket name where static assets are stored. By default is autogenerated.                                                                                                                                                                                                                                                                 |
+| bucketRegion             | `string`          | `null`                                                             | Region where you want to host your s3 bucket. Make sure this is geographically closer to the majority of your end users to reduce latency when CloudFront proxies a request to S3. Set this if you are experiencing 307 temporary redirects. See https://aws.amazon.com/premiumsupport/knowledge-center/s3-http-307-response/ for more details. |
```
danielcondemarin, line 361: The addition of "Set this if you are experiencing ... etc." should be deleted. That has been solved now.

packages/lambda-at-edge/src/default-handler.ts
```diff
@@ -24,6 +26,11 @@ const addS3HostHeader = (
 
 const isDataRequest = (uri: string): boolean => uri.startsWith("/_next/data");
 
+const normaliseUri = (uri: string): string => {
+  if (basePath && uri.startsWith(basePath)) uri = uri.slice(basePath.length);
+  return uri === "/" || uri === "" ? "/index" : uri;
```
danielcondemarin, line 31: The `/` to `/index` normalisation isn't necessary anymore. I think you can just return `uri` directly after the basePath normalisation;

```suggestion
  return uri;
```

packages/serverless-component/src/component.ts
```diff
@@ -42,9 +43,19 @@ class NextjsComponent extends Component {
     );
   }
 
+  readRoutesManifest(nextConfigPath: string): Promise<RoutesManifest> {
+    return readJSON(join(nextConfigPath, ".next/routes-manifest.json"));
+  }
+
+  pathPattern(pattern: string, routesManifest: RoutesManifest): string {
+    const basePath = routesManifest.basePath;
+    return basePath.length === 0 ? pattern : `${basePath.slice(1)}/${pattern}`;
```
danielcondemarin, line 52: Is this backward compatible with an older route manifest that doesn't include basePath? i.e. what if basePath is undefined?

packages/lambda-at-edge/src/default-handler.ts
rrooding: Removing `uri === "/"` works. If I remove `uri === ""` then this specific test fails: https://github.com/serverless-nextjs/serverless-next.js/pull/510/files#diff-1e2c3ac8a5020d5e147d22a7d7df9029R49

packages/serverless-component/src/component.ts
rrooding: Added an extra check for this :+1:

README.md
rrooding: Fixed

rrooding:
Not sure about the CI failures, seem to happen sometimes? Are they related to this PR? Locally everything is green.

packages/lambda-at-edge/src/default-handler.ts
danielcondemarin: What if you do `return uri === "" ? "/" : uri;` ?

danielcondemarin:
@rrooding I've pushed a fix for returning a 404 when the path doesn't start with the base path. It was matching both with and without the basepath. Also fixed `public/` files not using the `basepath/`

There seems to be an outstanding issue with trailing slashes. `exampledistrib.com/basepath/` is matched but `exampledistrib.com/basepath` isn't. Probably related to https://github.com/serverless-nextjs/serverless-next.js/issues/479 so I'll merge this PR anyway. 

Thanks for implementing 🙏

danielcondemarin:
> Not sure about the CI failures, seem to happen sometimes? Are they related to this PR? Locally everything is green.

The CI issues seemed to be due to an attempt to cache node_modules https://github.com/serverless-nextjs/serverless-next.js/pull/510/commits/28396ffe6b5ec6937849501509d77bb64c7df89a. Sorted now

PR #519, opened 2020-07-21
Change in serverless-component entry location

maddygoround:
Entry location changed from dist/compoenent.js to serverless,js


danielcondemarin:
Thanks!

PR #533, opened 2020-08-07
IncludeBody option only with viewer-request or origin-request

Enalmada:
I need to set headers for a specific public file so I wanted to setup origin-response or viewer-response for that.

Adding a viewer-response or origin-response causes the following error message:
  `InvalidLambdaFunctionAssociation: The IncludeBody option can only be used with viewer-request or origin-request events.`

**To Reproduce**
Add to serverless.yml
```
    cloudfront:
      defaults:
        lambda@edge:
          viewer-response: <arn>
```

- OS: Windows 10
- Version @sls-next/serverless-component@1.16.0

**Additional context**

Doing something like this...
```
    cacheBehavior.LambdaFunctionAssociations.Items.push({
      EventType: eventType,
      LambdaFunctionARN: lambdaAtEdgeConfig[eventType],
      IncludeBody: eventType.contains("request") ? true : undefined
    });
```
in /packages/aws-cloudfront/lib/addLambdaAtEdgeToCacheBehavior.js works for viewer-response.

Note: If I try to do it for origin-response, the deployment is successful but origin-response still doesn't get set.  I see these lines of comments in component.ts.html so perhaps serverless-next.js is intentionally filtering.
    // make sure that origin-response is not set.
    // this is reserved for serverless-next.js usage

### DIFF
--- a/packages/aws-cloudfront/__tests__/__snapshots__/lambda-at-edge.test.js.snap
+++ b/packages/aws-cloudfront/__tests__/__snapshots__/lambda-at-edge.test.js.snap
@@ -55,12 +55,12 @@ Object {
               },
               Object {
                 "EventType": "origin-response",
-                "IncludeBody": true,
+                "IncludeBody": undefined,
                 "LambdaFunctionARN": "arn:aws:lambda:us-east-1:123:function:originResponseFunction",
               },
               Object {
                 "EventType": "viewer-response",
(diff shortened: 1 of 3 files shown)


danielcondemarin:
Published now under `@sls-next/serverless-component@1.17.0-alpha.1` 🥳

Issue #547, opened 2020-08-15
Very high cold start times


asterikx:
**Describe the bug**
I'm experiencing `Invocation` times of more than 7 seconds for cold lambdas. Subsequent invocations take ~250ms.

I'm using the `@sls-next/lambda-at-edge@1.6.0-alpha.0` package with serverless trace mode and deploy the lambdas/public resources using CDK. My (zipped) bundle size is 6.9MB.

First invocation:
<img width="1076" alt="Screenshot 2020-08-15 at 13 57 21" src="https://user-images.githubusercontent.com/16797665/90311931-6377a180-df00-11ea-9325-a078e8a6660e.png">

Second invocation (within a few minutes after the first invocation):
<img width="1097" alt="Screenshot 2020-08-15 at 13 57 46" src="https://user-images.githubusercontent.com/16797665/90311933-712d2700-df00-11ea-9756-d9570c7fc5ee.png">

**Expected behavior**
`Invocation` duration for cold lambdas should be similar to hot lambdas. It shouldn't be 30 times more.

**Question**
How to measure/debug the performance of the `default-handler`? The performance problem seems related to the handler code rather than resource initialization during cold starts.

danielcondemarin:
@asterikx Could you share the code for the next page you're seeing the high cold start time? i.e. Are you using getServerSideProps, getStaticProps etc. that may be adding time during server side rendering?

asterikx:
@danielcondemarin sure, I'm using next-i18next, that needs `getInitialProps`:

```tsx
import * as React from 'react';
import { ChakraProvider, CSSReset, Global } from '@chakra-ui/core';
import { track } from 'react-tracking';
import type { AppProps, AppContext } from 'next/app';
import { RecoilRoot } from 'recoil';

import { theme } from '../theme';
import { appWithTranslation } from '../i18n';
import { styles } from '../styles/styles';
import { UserProvider } from '../contexts/user';
import { configureAuth } from '../utils/auth';
import { HeaderProvider } from '../contexts/header';
import { MDXProvider } from '../components/MDXProvider';
import { useTrackPages } from '../hooks/useTrackPages';
import App from 'next/app';

configureAuth();

function MyApp({ Component, pageProps }: AppProps) {
  useTrackPages();

  return (
    <ChakraProvider theme={theme}>
      <CSSReset />
      <Global styles={styles} />
      {/*<Suspense fallback={<SplashScreen />}>*/}
      <MDXProvider>
[...]

PR #549, opened 2020-08-16
Feature/log execution times

asterikx:
Add execution time logging to `default-handler` to debug high response times.

#547


danielcondemarin (approved):
Thanks again for PR'ing @asterikx 🙏

PR #562, opened 2020-08-26
Support Preview Mode

thchia:
#355 
#### lambda-at-edge-compat

- NextJS internally calls `res.writeHead(...).end()`, so we have to return the res in `writeHead`.

#### lambda-at-edge

- For all HTML/JSON data requests (including potential fallbacks), only forward the response to the origin if it is not a preview request.
- When we reach the SSR portion, we are potentially still handling a JSON data request (due to changes in the previous point). Check and render/return the appropriate content.

Handling a client side transition in preview mode is not really supported beyond the first transition. After this, NextJS caches the JSON response in memory and does not even issue a new request if you navigate back to that page in the same session (without a refresh). Consider the following steps:
1. Enter preview mode for `mysite.com/blog/1` (an SSG page)
2. `mysite.com/blog/1` is SSR'd because we are in preview mode
3. Navigate to `mysite.com/blog/2` using `next/link`
4. Navigate back to `mysite.com/blog/1` using `next/link`
5. JSON data for Blog 1 is generated at request time
6. Repeat steps 3-4
7. JSON data is **not** regenerated, because we already have it in memory and NextJS does not issue a new request


packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -162,10 +180,16 @@ const handleOriginRequest = async ({
   const normalisedS3DomainName = normaliseS3OriginDomain(s3Origin);
   const hasFallback = hasFallbackForUri(uri, prerenderManifest);
   const { now, log } = perfLogger(manifest.logLambdaExecutionTimes);
+  const isPreviewRequest = checkPreviewRequest(request);
```
danielcondemarin, line 183: Could we check for "preview request" only if previewMode is enabled? There is a perf impact on looking for the preview mode cookie on every req that hopefully we can do only if necessary.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -21,6 +21,24 @@ import {
 import S3 from "aws-sdk/clients/s3";
 import { performance } from "perf_hooks";
 
+const parseCookieValue = (cookie: string) => {
```
danielcondemarin, line 24: I was hoping we could use express cookie-parser (https://github.com/jshttp/cookie) here. It's benchmarked and battle tested. Even if it means to [borrow the code parser](https://github.com/jshttp/cookie/blob/243fa7edab7f213a6af8adb339e9530daa23b445/index.js#L49).

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -21,6 +21,24 @@ import {
...
+    return acc;
+  }, {});
+};
+
+const checkPreviewRequest = (request: CloudFrontRequest) => {
+  const targetCookie = request.headers.cookie || [];
+  return targetCookie.some((cookieObj) => {
+    const cookieValue = parseCookieValue(cookieObj.value);
+    return (
+      cookieValue["__next_preview_data"] && cookieValue["__prerender_bypass"]
```
danielcondemarin, line 37: We'll need to also validate that the signature of the _next_preview_data is correct using json web tokens. [Here is](https://github.com/vercel/next.js/blob/42e8dcf281526b5dde27448010e96edcd5ceb196/packages/next/next-server/server/api-utils.ts#L374) next-server implementation of it.

packages/libs/lambda-at-edge/src/default-handler.ts
thchia: I think I'm missing a trick - I thought the only way to determine preview mode _is_ due to the presence of the cookies. There isn't a build config or any static info that we can check I think.

packages/libs/lambda-at-edge/src/default-handler.ts
thchia: Any guidance on how we will include these extra dependencies? For the cookie parser it's not a big deal because it doesn't have dependencies of its own, so we can do the same thing that is being done for `next-aws-cloudfront` and copy it to the lambda bundle.

For `jsonwebtoken`, that won't be possible (without copying over everything in `node_modules`). We could inline it by copying over the contents from [here](https://github.com/vercel/next.js/blob/42e8dcf281526b5dde27448010e96edcd5ceb196/packages/next/compiled/jsonwebtoken/index.js), but not sure if that's what you had in mind.

packages/libs/lambda-at-edge/src/default-handler.ts
danielcondemarin: I think you're right, seems like the only way to determine this is looking if the cookies exist :/ Next.js is generating in the `prerender-manifest.json` a "preview" object that will always be there on each build.

packages/libs/lambda-at-edge/src/default-handler.ts
danielcondemarin: Yeah I was going to suggest a bundler. Last time I checked `rollup` was the best option for libraries.

thchia: I'll proceed with rollup, it's a stalwart of the bundlers.

thchia:
@danielcondemarin marking this as draft for now.

I have added rollup and am bundling the default handler. As a result of this, the build step for this package is getting fragmented, with one thing for default lambda and another for the api lambda and the build function (also, the bundler needs a different tsconfig).

Should we bundle the api lambda too? For now I am just telling the TS process for the existing build to ignore the default lambda file, and letting it get handled by rollup. We can have 2 sub-builds, one for the lambdas and one for the build function, with both going into the `dist` folder.

danielcondemarin:
> Should we bundle the api lambda too? 
If it isn't much extra work go for it. There is refactor that could be made once the handler is bundled like splitting logic into other files etc. 

> For now I am just telling the TS process for the existing build to ignore the default lambda file, and letting it get handled by rollup. We can have 2 sub-builds, one for the lambdas and one for the build function, with both going into the dist folder.

Sounds good to me 👍

thchia:
@danielcondemarin this is good for review. I will squash the commits before we merge and close this though.

- For some reason, rollup didn't like using a tsconfig that extends another tsconfig. Hence why `tsconfig.bundle.json` is standalone 🤷‍♀️
- If the preview token verification fails, it's a "hard fail" rather than sending back a nice 403 page (like `pages/_error.js`). I'm neither here nor there on this, but I don't think it's necessary to go through the SSR for this.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -272,7 +313,17 @@ const handleOriginRequest = async ({
     }
```
dphang, line 313: I think this can be moved into the else block below just before the `await page.render(req, res)`?

packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
```diff
@@ -186,6 +199,35 @@ describe("Lambda@Edge", () => {
 
         expect(response.status).toEqual("200");
       });
+      
+      it("handles preview mode", async () => {
+        const event = createCloudFrontEvent({
+          uri: "/preview",
```
dphang, line 205: I think you would need to append trailing slash if `trailingSlash` is true to the URI

packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
```diff
@@ -186,6 +199,35 @@ describe("Lambda@Edge", () => {
...
+            cookie: [
+              {
+                key: "Cookie",
+                value: "__next_preview_data=abc; __prerender_bypass=def"
+              }
+            ]
+          }
+        });
+
+        if (trailingSlash) {
```
dphang, line 217: You shouldn't need the `trailingSlash` block here, I added a redirect test where you can add `/preview` as one of the tests, it will test both kinds of redirects

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -21,6 +22,27 @@ import {
...
+      cookieValue[NEXT_PREVIEW_DATA_COOKIE] &&
+      cookieValue[NEXT_PRERENDER_BYPASS_COOKIE]
+    );
+  });
+  if (previewCookie) {
+    return cookie.parse(previewCookie.value) as {
+      [NEXT_PREVIEW_DATA_COOKIE]: string;
+      [NEXT_PRERENDER_BYPASS_COOKIE]: string;
+    };
+  }
```
dphang, line 44: Add an explicit return value here?

packages/libs/lambda-at-edge/src/default-handler.ts
thchia: Does that mean we know for a fact that data requests will not result in a possible error page?

packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
thchia: You mean changing the URI in `createCloudFrontEvent` to be either `/preview` or `/preview/` depending on whether `trailingSlash` is true, and then just testing for the expected response?

packages/libs/lambda-at-edge/src/default-handler.ts
dphang: Ah, I guess unmatched data request can still go to `pages/_error.js`, maybe still this at the start

packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
dphang: Yep, basically if trailingSlash is true then you can pass `/preview/` as the URI, otherwise it is `/preview`.

Then you can add the same path `/preview` as a test case in redirect test, it will auto append trailing slash and test the redirects

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -211,11 +233,30 @@ const handleOriginRequest = async ({
...
+  const isPreviewRequest = getPreviewCookies(request);
+
+  if (isPreviewRequest) {
+    try {
+      jsonwebtoken.verify(
+        isPreviewRequest[NEXT_PREVIEW_DATA_COOKIE],
+        prerenderManifest.preview.previewModeSigningKey
+      );
+    } catch (e) {
+      return {
```
dphang, line 245: Maybe some logging is useful here, I added it for 500 page as well, not sure if it's needed for this though

packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
thchia: Done.

packages/libs/lambda-at-edge/src/default-handler.ts
thchia: Done.

packages/libs/lambda-at-edge/src/default-handler.ts
thchia: I added a `console.error`, similar to the logging for the 500 response.

packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
dphang: Also i think you can add `/preview` to redirect test (around line 158).

thchia: Done.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -21,6 +22,32 @@ import {
...
 import { performance } from "perf_hooks";
 import { ServerResponse } from "http";
+import jsonwebtoken from "jsonwebtoken";
+
+const NEXT_PREVIEW_DATA_COOKIE = "__next_preview_data";
+const NEXT_PRERENDER_BYPASS_COOKIE = "__prerender_bypass";
+
+const getPreviewCookies = (request: CloudFrontRequest) => {
+  const targetCookie = request.headers.cookie || [];
+  const previewCookie = targetCookie.find((cookieObj) => {
```
danielcondemarin, line 32: I think using a `reduce` here instead of a map will save 2 the extra call to `cookie.parse` below.

danielcondemarin (approved):
🚀

danielcondemarin:
I'll do some testing and get an alpha published tonight 👍

Issue #566, opened 2020-08-27
CloudFront showing generic 503 error page when SSR page fails to render (e.g in getInitialProps()) instead of Next.js error page


dphang:
**Describe the bug**
When SSR page fails to render (e.g a server-side API call fails, etc.), then CloudFront error page (503 Bad Gateway) is shown instead of the Next.js error page.

Searched and I don't believe anyone else reported this issue

**To Reproduce**
Steps to reproduce the behavior

1. Create an SSR page that throws an error in `getInitialProps()`
2. Deploy with serverless-next.js (1.16, 1.17)
3. Go to that page
4. You will get 503 error as below.

```
503 ERROR
The request could not be satisfied.
The Lambda function associated with the CloudFront distribution is invalid or doesn't have the required permissions. We can't connect to the server for this app or website at this time. There might be too much traffic or a configuration error. Try again later, or contact the app or website owner.
If you provide content to customers through CloudFront, you can find steps to troubleshoot and help prevent this error by reviewing the CloudFront documentation.
Generated by cloudfront (CloudFront)
Request ID: t-JGXBL9ipU4vhfWAEKtcwPreZ5zWn3Ujwf-U2ynduRlP90VdiZ2AQ==
```

In Lambda logs, we get an unhandled promise rejection:

```
2020-08-27T04:15:23.768Z	dc177b96-2a0c-4094-afea-eebd8899fcbe	ERROR	Unhandled Promise Rejection 	
{
    "errorType": "Runtime.UnhandledPromiseRejection",
    "errorMessage": "Error: error",
    "reason": {
        "errorType": "Error",
        "errorMessage": "error",
        "stack": [
            "Error: error",
            "    at Function.module.exports.2HMD.AnotherPage.getInitialProps (/var/task/pages/anotherSSR.js:5691:9)",
            "    at loadGetInitialProps (/var/task/pages/anotherSSR.js:32196:27)",
            "    at Function.appGetInitialProps (/var/task/pages/anotherSSR.js:13867:58)",
            "    at loadGetInitialProps (/var/task/pages/anotherSSR.js:32196:27)",
            "    at renderToHTML (/var/task/pages/anotherSSR.js:1738:1221)",
            "    at async renderReqToHTML (/var/task/pages/anotherSSR.js:21856:22)",
            "    at async Module.render (/var/task/pages/anotherSSR.js:21929:22)"
        ]
    },
    "promise": {},
    "stack": [
[...]

PR #570, opened 2020-08-28
fix(lambda-at-edge): render Next 500 page when SSR render fails, and ensure 404 pages return 404 status codes

dphang:
## Description

This fixes https://github.com/serverless-nextjs/serverless-next.js/issues/566 and few other improvements:

* 500 page that is part of Next.js (`_error.js`) is rendered instead of CloudFront's 503 when SSR page has an error. It also sets 500 status code so that the message in the page actually says there is a 500 error.
* 404 pages were returning 200 status codes incorrectly, and now return 404 status codes. Similarly, ensured that we passed in 404 status into page render (previously we weren't, though Next.js by default renders 404 page anyway)

FYI: CloudFront will cache 4xx/5xx responses for 1 minute: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages-expiration.html. We could make it less if needed, but I think this is sensible.

## Tests

Added and updated unit tests for `default-handler` for all the above cases.

You can also see a sample app:

500 page (page throws error in `getInitialProps()`: https://dpjozqtv88tg7.cloudfront.net/anotherSSR/ -> 500 status code
404 page (statically rendered): https://dpjozqtv88tg7.cloudfront.net/notfound/ -> 404 status code. Try it with any unmatched path and it should work.

404 page (SSR): will add another app for this.

## Other Notes

I believe this should have minimal perf impact:

* In case of a 500, the Lambda is already initialized from trying to render the initial page. So doing another `require` on the `_error.js` page (which would normally have few dependencies, maybe just the app layout) should be pretty fast.
* In case of a 404 that's based on `_error.js`, i.e it is SSR (user did not add a static 404 page), the rendering is the same, I am just setting the status code to be 404 (instead of 200) to render the page as a 404 and have the correct response code.
* In case of a 404 that's statically rendered (user added a static 404 page), in origin response handler, after S3 returns success, we check the `router` for whether it's the `404.html` page, and set the 404 status code in the response as well. Future calls will be cached in CloudFront. This router check would need to be added for all such origin responses. I believe the router call should be pretty fast.

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -20,6 +20,7 @@ import {
 } from "../types";
 import S3 from "aws-sdk/clients/s3";
 import { performance } from "perf_hooks";
+import { ServerResponse } from "http";
 
 const perfLogger = (logLambdaExecutionTimes: boolean): PerfLogger => {
   if (logLambdaExecutionTimes) {
@@ -221,22 +222,41 @@ const handleOriginRequest = async ({
     s3Origin.path = `${basePath}/static-pages`;
     request.uri = pagePath.replace("pages", "");
     addS3HostHeader(request, normalisedS3DomainName);
+
     return request;
   }
 
   const tBeforePageRequire = now();
-  const page = require(`./${pagePath}`); // eslint-disable-line
+  let page = require(`./${pagePath}`); // eslint-disable-line
   const tAfterPageRequire = now();
 
   log("require JS execution time", tBeforePageRequire, tAfterPageRequire);
 
   const tBeforeSSR = now();
   const { req, res, responsePromise } = lambdaAtEdgeCompat(event.Records[0].cf);
-  page.render(req, res);
+  try {
+    // If page is _error.js, set status to 404 so _error.js will render a 404 page
+    if (pagePath === "pages/_error.js") {
+      res.statusCode = 404;
+    }
+
+    // Render page
+    await page.render(req, res);
+  } catch (error) {
+    // Set status to 500 so _error.js will render a 500 page
+    console.error(
+      `Error rendering page: ${pagePath}. Error:\n${error}\nRendering Next.js error page.`
+    );
+    res.statusCode = 500;
+    page = require("./pages/_error.js"); // eslint-disable-line
+    await page.render(req, res);
+  }
   const response = await responsePromise;
   const tAfterSSR = now();
 
   log("SSR execution time", tBeforeSSR, tAfterSSR);
+
+  setCloudFrontResponseStatus(response, res);
 
   return response;
 };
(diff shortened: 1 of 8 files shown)


packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -221,23 +222,42 @@ const handleOriginRequest = async ({
...
 
   log("require JS execution time", tBeforePageRequire, tAfterPageRequire);
 
   const tBeforeSSR = now();
   const { req, res, responsePromise } = lambdaAtEdgeCompat(event.Records[0].cf);
-  page.render(req, res);
+  try {
+    // If page is _error.js, set status to 404 so _error.js will render a 404 page
+    if (pagePath === "pages/_error.js") {
+      res.statusCode = 404;
```
dphang, line 240: Not completely sure why this works, I think the params might be slightly named incorrectly in case of `_error.js` page? Will need to look into the code more for the `render()` function in `_error.js`.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -221,23 +222,42 @@ const handleOriginRequest = async ({
...
+    }
+
+    // Render page
+    await page.render(req, res);
+  } catch (error) {
+    // Set status to 500 so _error.js will render a 500 page
+    console.error(
+      `Error rendering page: ${pagePath}. Error:\n${error}\nRendering Next.js error page.`
+    );
+    res.statusCode = 500;
```
dphang, line 250: Not completely sure why this works, I think the params might be slightly named incorrectly in case of _error.js page? Will need to look into the code more for the render() function in _error.js.

packages/libs/lambda-at-edge/src/default-handler.ts
thchia: `statusCode` follows the property name from NodeJS's HTTP module, which is followed by express, which in turn I think is followed by NextJS.

dphang: I see, makes sense, thanks.

packages/libs/lambda-at-edge/tests/default-handler/default-handler-with-basepath.test.ts
```diff
@@ -445,5 +445,26 @@ describe("Lambda@Edge", () => {
         await runRedirectTest(path, expectedRedirect);
       });
     });
+
+    describe("500 page", () => {
+      xit("renders 500 page if page render has an error", async () => {
```
dphang, line 450: woops, need to change this from xit -> it

packages/libs/lambda-at-edge/src/default-handler.ts
danielcondemarin: Lgtm. I hadn't realised before that Next.js stopped serving 500 error page on `page.render` uncaught errors. https://github.com/vercel/next.js/commit/15cdb4f4089713b7f1e3079522b15bdd424ac27c

danielcondemarin (approved):
Lgtm 🚀

packages/libs/lambda-at-edge/src/default-handler.ts
dphang: Thanks for merging! Yeah, it looks like it was changed in 9.5? Might be worth to add some end-to-end tests on CloudFront in the future to verify with the latest Next.js versions (e.g 9.4.x and 9.5.x) as the unit and integration tests won't always catch these kinds of issues :)

Issue #571, opened 2020-08-28
Root page with basePath doesn't work without trailingSlash


marcin-piela-sf:
**Describe the bug**

I see one problem after deploy, I have a basePath "test/dwa" and app returns 404 on:

https://d1g4p9lwi1bqdw.cloudfront.net/test/dwa

But it works on:
https://d1g4p9lwi1bqdw.cloudfront.net/test/dwa/

I have `trailingSlash:false` in next.js config.

```
sls-next/lambda-at-edge@1.6.0-alpha.5
```

**Expected behavior**
Root url with subPath should work with and without trailingSlash

dphang:
Thanks, let me take a look!

dphang:
Just tried out. I have a pages structure of the following (tested having `a/b` route to `b` page and `a/c` with index page under `c`:

```
pages
├── a
│   ├── b.tsx
│   └── c
│       └── index.tsx
└── index.tsx
```

Next config:

```js
module.exports = {
  target: 'serverless',
  trailingSlash: false
}

```

I'm deploying with `@sls-next/serverless-component@1.17.0-alpha.7`

The repro app is here: https://github.com/dphang/nextjs-repros/tree/nested-pages

The subpath `a/b` and `a/c` returns 200 and `a/b/` and `a/c/` returns a 308 redirect.

https://d113kzwzk7s27f.cloudfront.net/a/b (200)
https://d113kzwzk7s27f.cloudfront.net/a/b/ (308 redirect to a/b)
https://d113kzwzk7s27f.cloudfront.net/a/c (200)
https://d113kzwzk7s27f.cloudfront.net/a/c/ (308 redirect to a/c)

Could not reproduce any issue with this and subpaths should be covered by the tests too. Could you please share your repro code?

marcin-piela-sf:
Sorry I wrote "subPath" but I meant "basePath" in next.config.js:
```
  basePath: '/test/dwa',
```

dphang:
Ah, thanks for clarifying, I can reproduce the issue with `basePath`. Trying to find the bug now.

dphang:
Found the problem. It seems to be in the normalization here:

https://github.com/serverless-nextjs/serverless-next.js/blob/ca63b80d4bf784ebfdc5a32352a53dde85b4b4d9/packages/libs/lambda-at-edge/src/default-handler.ts#L49-L58

[...]

PR #572, opened 2020-08-28
fix(lambda-at-edge): proper routing for pages with basePath

dphang:
## Description

This fixes routing for basePath URLs and fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/571.

1. Routing/redirects for basePath with index page (and others) did not work correctly (e.g with default settings `/basePath` returned  `/index` which is not a valid path so it returned 404 page and `/basePath/` returned index page).
2. If basePath is expected but unnormalized URL doesn't have it as prefix, serve static 404 page. No redirects are done for these if they have trailing slash or not (same as Next.js)

## Cause

It looks to be due to the normalization logic where `basePath` becomes empty after removing `basePath`, then returns `/index` after normalization. Then it tries to find a page `/index`, but it doesn't exist, so the SSR 404 page is rendered. With `basePath/`, it becomes `/` which routes to index page correctly.

This wasn't caught in tests before since `/index` was added as a route, but in actual Next.js apps only `/` exists (I saw the `/index` route was added in tests in Sept 2019, probably Next no longer supports this and only provided the `/` route).

As an aside, `basePath` stripping logic was also slightly wrong as it was just removing the length of `basePath`. If you have basePath of `/basepath`, then `/abc` becomes empty, and normalized URI is `/index` (which doesn't exist, so it goes to 404 page, which happens to be correct).  Updated this to check for whether the URI actually starts with `basePath`, if not, return static 404 page.

## Tests

Updated `default-handler` tests. Removed `/index` route as above, in a real Next.js app, `index` page is never mapped to `/index`, only `/`.

Test app with basepath of /test/dwa: https://d113kzwzk7s27f.cloudfront.net/test/dwa (with trailing slash it redirects to this). I will test with various pages and with/without basepath.

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -47,14 +47,18 @@ const addS3HostHeader = (
 const isDataRequest = (uri: string): boolean => uri.startsWith("/_next/data");
 
 const normaliseUri = (uri: string): string => {
-  if (basePath) uri = uri.slice(basePath.length);
+  // Remove basePath prefix
+  if (basePath && uri.startsWith(basePath)) {
+    uri = uri.slice(basePath.length);
+  }
 
-  // Remove trailing slash for all paths except "/"
-  if (uri.length > 1 && uri.endsWith("/")) {
+  // Remove trailing slash for all paths
+  if (uri.endsWith("/")) {
     uri = uri.slice(0, -1);
   }
 
-  return uri === "" ? "/index" : uri;
+  // Empty path should be normalised to "/" as there is no Next.js route for ""
+  return uri === "" ? "/" : uri;
 };
 
 const normaliseS3OriginDomain = (s3Origin: CloudFrontS3Origin): string => {
@@ -169,8 +173,10 @@ const handleOriginRequest = async ({
     if (newUri.endsWith("/")) {
       newUri = newUri.slice(0, -1);
     }
-  } else if (uri !== "/index" && uri !== "/") {
-    // HTML/SSR pages get redirected based on trailingSlash in next.config.js, except for index page
+  } else if (request.uri !== "/" && request.uri !== "") {
+    // HTML/SSR pages get redirected based on trailingSlash in next.config.js
+    // We should never redirect unnormalised URI "/" or "" as this could cause a redirect loop due to browsers appending trailing slash
+    // to root page requests
     const trailingSlash = manifest.trailingSlash;
 
     if (!trailingSlash && newUri.endsWith("/")) {
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -47,9 +47,13 @@ const addS3HostHeader = (
 const isDataRequest = (uri: string): boolean => uri.startsWith("/_next/data");
 
 const normaliseUri = (uri: string): string => {
-  // Remove basePath prefix
-  if (basePath && uri.startsWith(basePath)) {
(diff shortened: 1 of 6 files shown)


packages/libs/lambda-at-edge/tests/default-handler/default-build-manifest-with-404.json
```diff
@@ -40,7 +40,6 @@
     "html": {
       "nonDynamic": {
         "/": "pages/index.html",
-        "/index": "pages/index.html",
```
dphang, line 43: "/index" is not a valid route for the current Next.js app builds (only "/" is). It was inadvertently causing us to pass the basepath tests as "/basePath" redirected to /index.html due to this. So I removed it.

packages/libs/lambda-at-edge/tests/default-handler/default-handler-with-basepath.test.ts
```diff
@@ -109,8 +109,6 @@ describe("Lambda@Edge", () => {
       it.each`
         path                                                        | expectedPage
         ${"/basepath"}                                              | ${"/index.html"}
-        ${"/basepath/"}                                             | ${"/index.html"}
```
dphang, line 112: Trailing slash should be redirected to non-trailing slash instead of going to index.html page

packages/libs/lambda-at-edge/tests/default-handler/default-handler-with-basepath.test.ts
```diff
@@ -150,6 +148,7 @@ describe("Lambda@Edge", () => {
 
       it.each`
         path
+        ${"/basepath"}
```
dphang, line 151: This is the 308 redirect test

packages/libs/lambda-at-edge/tests/default-handler/default-handler-with-basepath.test.ts
```diff
@@ -109,8 +109,6 @@ describe("Lambda@Edge", () => {
       it.each`
         path                                                        | expectedPage
         ${"/basepath"}                                              | ${"/index.html"}
-        ${"/basepath/"}                                             | ${"/index.html"}
-        ${"/basepath/index"}                                        | ${"/index.html"}
```
dphang, line 113: This is not a valid page route for real Next.js apps (only /, not /index)

packages/libs/lambda-at-edge/tests/default-handler/default-handler-with-basepath.test.ts
```diff
@@ -434,7 +433,7 @@ describe("Lambda@Edge", () => {
       });
 
       it("redirects unmatched request path", async () => {
-        let path = "/page/does/not/exist";
+        let path = "/basepath/page/does/not/exist";
```
dphang, line 436: This is just a minor fix on a test (forgot to add basepath)

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -47,14 +47,22 @@ const addS3HostHeader = (
 const isDataRequest = (uri: string): boolean => uri.startsWith("/_next/data");
 
 const normaliseUri = (uri: string): string => {
-  if (basePath) uri = uri.slice(basePath.length);
+  if (basePath) {
+    if (uri.startsWith(basePath)) {
+      uri = uri.slice(basePath.length);
+    } else {
+      // basePath set but URI does not start with basePath, return 404
+      return "/404";
```
dphang, line 55: This is a reserved route that will let us serve the 404.html (static page) (generated by Next.js), won't conflict with any user pages.

danielcondemarin:
LGTM 🚀

PR #575, opened 2020-08-30
Allow passing an inline policy

SarKurd:
This will make adding extra permissions easier than creating a policy then passing an ARN


SarKurd:
Not sure why the test failing :/

SarKurd:
Tested on our website and it works great, appreciate review.

danielcondemarin (approved):

Issue #576, opened 2020-08-30
All public folder files are 404ing


tyler-ground:
Update: All public folder files are 404ing in 1.17.0.  DPhang created a fix linked below.

Old Post Body:
The /.well-known/apple-app-site-association path in public folder is returning 404.  Seeing it on alpha-7 but also appears to be happening on alpha-6.  Oddly, the file is still returned with the 404 status code.  Problem was detected by a sudden Cloudfront 4XX error rate increase a few hours ago.

I have temporarily worked around this by adding the /.well-known path as a behaviour and putting a file there in S3.  Deployments delete the behaviour of course so it's a short term resolution.

**To Reproduce**
1. Put an apple-site-association file in your public .well-known folder
2. Deploy
3. Try to reach it

**Expected behavior**
File returns

Version:
1.17.0-alpha.7 but likely also earlier

Update: This affects all 1.17 builds since the latest lambda-at-edge package is used for all of them.

dphang:
Thanks for reporting this bug! Yeah, I think it's because of this code: https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/libs/lambda-at-edge/src/default-handler.ts#L310-L313. It's supposed to set the 404 status code properly for 404.html page, but looks like it's also setting public files' status code to 404. Will take a look at this.

tyler-ground:
You're right!  I see that it's actually happening for other public files as well.

tyler-ground:
This might be an even older issue actually.  I just rolled back even further to alpha.5 and it's still happening (That said, perhaps Cloudfront caching is causing an issue here).

tyler-ground:
@dphang I think it may be a problem with how the package.json is setup for the project.  It appears to get the latest version of lambda-at-edge irregardless of which version of this package you install.  I wonder if that means every deployment is broken for every version of serverless-next.js right now?

dphang:
Yeah, I saw the same issue before. @danielcondemarin not sure how to fix this?

tyler-ground:
Can I override the serverless deploy to only use my local node_modules ?  I can overwrite all the @sls-next packages with versions from before yesterday but when I call serverless it seems to fetch its own versions.

tyler-ground:
Alternatively @dphang , do you have release access?  You could release a revert :)

dphang:
[...]

PR #577, opened 2020-08-30
fix(lambda-at-edge): fix for 404s on public files

dphang:
## Description

This fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/576

* Fixes issue in latest alphas (1.17 alpha.8) where public files are served correctly but with 404 status code. This is because in origin response we should only set 404 code on unnormalised uri `/404.html` (these are file URIs returned by the S3 origin).

## Tests

Added some test cases as well to verify this.

Verified with my app: https://d17ujsvie9w0tw.cloudfront.net/app-store-badge.png (returns 200)
https://d17ujsvie9w0tw.cloudfront.net/notfound/ (returns 404)
https://d17ujsvie9w0tw.cloudfront.net/anotherSSG/ (returns 200)
https://d17ujsvie9w0tw.cloudfront.net/anotherSSR/ (returns 200)
https://d17ujsvie9w0tw.cloudfront.net/ (returns 200)
https://d17ujsvie9w0tw.cloudfront.net/_next/data/jRDzHTftCxGo7bHQ1tybv/index.json (returns 200)
https://d17ujsvie9w0tw.cloudfront.net/_next/data/notfound (returns 404)

## Additional context

Seems something like this is not easily caught in unit tests (or hard to write, due to different inputs to request vs. response handlers). Apologies since I was confusing the input to origin response handler (should be "/404.html" for 404 page). Probably if we have e2e tests it would be easier to catch (which I'm working on).

Also, as mentioned in the issue, there seems to be some versioning issue in the latest alpha..seems the 1.17 alpha.5 - alpha.8 all point to latest commit of lambda-at-edge package? So this issue was impacting all those versions. Not sure why this is so?

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -306,8 +306,8 @@ const handleOriginResponse = async ({
   const uri = normaliseUri(request.uri);
   const { status } = response;
   if (status !== "403") {
-    const pagePath = router(manifest)(uri);
-    if (pagePath === "pages/404.html") {
+    const { publicFiles } = manifest;
+    if (!publicFiles[uri] && router(manifest)(uri) === "pages/404.html") {
       response.status = "404";
       response.statusDescription = "Not Found";
     }
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -303,16 +303,16 @@ const handleOriginResponse = async ({
 }) => {
   const response = event.Records[0].cf.response;
   const request = event.Records[0].cf.request;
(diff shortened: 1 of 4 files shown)


danielcondemarin:
Also, as mentioned in the issue, there seems to be some versioning issue in the latest alpha..seems the 1.17 alpha.5 - alpha.8 all point to latest commit of lambda-at-edge package? So this issue was impacting all those versions. Not sure why this is so?

That's really strange. nextjs-component package [adds as a dependency](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-components/nextjs-component/package.json#L41) next-aws-cloudfront. On publish lerna should then replace `link:../` with the actual version number, alpha or whatever that is.

danielcondemarin (approved):

Issue #573, opened 2020-08-29
Add e2e tests (e.g app deployed on CloudFront) as part of CI/CD workflow.


dphang:
**Is your feature request related to a problem? Please describe.**
Not directly, but it would help the developer experience and reduce # regressions. This request is to add end-to-end test automation to the CI/CD workflows. Basically, as part of the workflow, actually deploy a test app to CloudFront and run an e2e test suite.

This would help in a few ways (among others):
* Regressions can be caught earlier by testing across a few recent versions of Next.js in the CI/CD workflow. Fewer GH issues would be created as well.
* We could then tag a release as tested with certain versions of Next.js (e.g 9.5.2). This would give more confidence to users, and less time is needed to validate alpha releases?
* Developers who are making PRs can use the test app + e2e test suite to locally verify their changes on their own AWS account.

**Describe the solution you'd like**

Basically, there are two parts:

* Test app: this should be a simple app, ideally without any UI libraries. This would test common scenarios like SSR/SSG pages, client-side navigation, trailing slash redirects, experimental-serverless-trace, error pages, etc. I guess the test app can live in the same monorepo.
* e2e test suite: we could use a browser test framework like [Cypress](https://www.cypress.io/) and API testing frameworks like [Supertest](https://github.com/visionmedia/supertest), or just use libraries like Axios, or node-fetch. (I'm using these myself, apart from Axios).

Bonus:

* Once this automation is in place and has sufficient coverage, on successful tests + e2e tests, we could auto-publish each commit as alpha releases on npm, so users can try out the latest features soon after the commits are merged. Only manual action is needed to publish stable releases.

**Describe alternatives you've considered**
N/A

**Additional context**
I thought of this since I was recently making a few changes. Right now I am manually testing with my own test app + my production app with these changes, but I can't cover all scenarios by myself or may miss some accidentally. I'm happy to help with some of these (e.g e2e test suite, test app) if you are willing to provide the AWS account (not sure if AWS provides free accounts for open source projects? It probably shouldn't cost more than a few cents to run per month).

danielcondemarin:
@dphang This is a great summary of something I have been wanting to do for a while now, but unfortunately because of time haven't been able to. 

I am 100% on-board with the idea. To be honest I don't have much availability in the next month due personal circumstances but more than happy to help out in the little time I have if you want to take a stab at this.

dphang:
Sounds good! I can try to setup a basic Next.js app in the next week or so.

[...]

PR #578, opened 2020-08-31
test(e2e-tests): add initial test next app

dphang:
## Description

This is the first PR for: https://github.com/serverless-nextjs/serverless-next.js/issues/573

This adds a simple Next.js app with an SSG page, SSR page 1 (getServerSideProps) and SSR page 2 (getInitialProps).

I also added the dependencies needed for testing using Cypress framework, but there are no tests yet

Note: I may change the structure and add/remove dependencies later as needed.


packages/e2e-tests/next-app/next.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
danielcondemarin, line 1: Maybe best to leave next.config.js out. That would test serverless-next.js ability to generate a next.config.js for users.

dphang: Yeah, I was planning to have different configurations in the future, this app would just be one of a few different apps (or we could also programmatically update the next.config.js for a single app).

For the first app, I could remove this.

danielcondemarin:
@dphang This looks good to me, I can merge unless there is anything else you wanted to do?

dphang:
> @dphang This looks good to me, I can merge unless there is anything else you wanted to do? 

Please go ahead :) I'll work on the actual tests in the meanwhile.

danielcondemarin (approved):

Issue #580, opened 2020-09-02
S3 import in default-handler likely causing higher cold start time


dphang:
**Describe the bug**
Not really a bug, more a slight optimization we can do. So I saw the S3 import was added for `getStaticPaths` support:

https://github.com/serverless-nextjs/serverless-next.js/blob/a854139f4344530de1a42268828231a4d38c7c91/packages/libs/lambda-at-edge/src/default-handler.ts#L21

Based on some quick perf tests, this `require` seems to be adding up to 80-100 ms to init times on cold start (1024 MB function, Node 12.x). Total init time is ~250 ms for my app.

I think s3 is not needed all the time, just if `getStaticPaths` is being used, so maybe this should be `required` just before  where it's actually needed, i.e: https://github.com/serverless-nextjs/serverless-next.js/blame/a854139f4344530de1a42268828231a4d38c7c91/packages/libs/lambda-at-edge/src/default-handler.ts#L319 instead?

**To Reproduce**

1. Add performance logging to the `default-handler` before and after the require of S3. Then build and deploy with the new handler.
2. Load any page and ensure it hits the Lambda.
3. Look into Lambda logs, you should see that S3 require takes ~100 ms (1024 MB function).

**Expected behavior**
Maybe lazily `require` this module for better overall performance.

**Additional context**
Also, maybe consider using the `aws-sdk-js-v3` (https://github.com/aws/aws-sdk-js-v3), as it is more modularized and lightweight? Although it is in gamma release now, so maybe we can wait for a more stable release.

I also did do some initial tests by bundling this new version: `@aws-sdk/client-s3-node` (via Rollup.js) and it seems to be `required` somewhat faster (by around 20-30 ms) than the built-in `aws-sdk` one, which I thought would have been fairly optimized already...

Related article about `aws-sdk` `require` times: https://theburningmonk.com/2019/03/just-how-expensive-is-the-full-aws-sdk/

thchia:
Sounds good - also this is the first I've heard of the `@aws-sdk` scope... it looks official on npm but is there any release documentation on it (just out of interest)?

danielcondemarin:
I like the idea of lazy requiring stuff that is only needed in certain code paths, sort of what Next.js [does here with jsonwebtoken](https://github.com/vercel/next.js/blob/fa153bcface44c19e3cb80832d429e39c1e24cc7/packages/next/next-server/server/api-utils.ts#L374).

dphang:
[...]

PR #583, opened 2020-09-04
feat(lambda-at-edge): use new aws s3 client for faster require time

dphang:
## Description

This closes https://github.com/serverless-nextjs/serverless-next.js/issues/580.

* Uses the client here in AWS SDK JS v3, or `@aws-sdk/client-s3` (https://github.com/aws/aws-sdk-js-v3/tree/master/clients/client-s3) which is a more modularized client. It will allow us to import only the commands needed (GetObject, PutObject) instead of heavy S3 client in `aws-sdk`, and do it dynamically instead of in every function invocation (i.e before the handler). E.g this will only need it for `getStaticPaths` cases.
* Tests in the linked Issue have shown it takes < 3 ms to load all of these. Unfortunately it does increase handler size from ~150 kB to ~500 kB because there are still a lot of other types for other S3 APIs (but the command themselves only import 2) (but code size has very little impact on cold start times compared to require size, i.e requiring `aws-sdk/S3` will load all APIs. In another PR, we could use terser with Rollup to reduce this to ~180 kB (and we can provide both minified and normal handlers).

Note, I had to use `rollup-plugin-typescript2` (a fork of original plugin) as I was getting these errors and could not resolve it:

```
[!] (plugin typescript) Error: Could not load /Users/danielphang/IdeaProjects/serverless-next.js/packages/libs/lambda-at-edge/node_modules/@aws-sdk/client-s3/S3Client.ts (imported by src/default-handler.ts): Debug Failure. False expression: Expected fileName to be present in command line
```

Though this is better in the sense that it prints more informative error messages. I also added `json` rollup plugin as it is needed to bundle the AWS client correctly.

## Tests

Updated the unit tests in `origin-response` handler test.

Currently in progress of testing end-to-end.


dphang:
This has been tested on this test app:

https://d2fn5f8m5hw83u.cloudfront.net/tags/128128218 (try replacing the end number with a random number, it will load fallback until it gets the data)

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -19,10 +19,10 @@ import {
   OriginResponseEvent,
   PerfLogger
 } from "../types";
-import S3 from "aws-sdk/clients/s3";
 import { performance } from "perf_hooks";
 import { ServerResponse } from "http";
 import jsonwebtoken from "jsonwebtoken";
+import { Readable } from "stream";
```
dphang, line 25: This is just a type, should be fine to import here.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -422,7 +429,14 @@ const handleOriginResponse = async ({
...
       Key: `static-pages${hasFallback.fallback}`
     };
-    const { Body } = await s3.getObject(s3Params).promise();
+    const { GetObjectCommand } = await import(
+      "@aws-sdk/client-s3/commands/GetObjectCommand"
+    );
+    const { Body } = await s3.send(new GetObjectCommand(s3Params));
+
+    // Body is stream per: https://github.com/aws/aws-sdk-js-v3/issues/1096
+    const getStream = await import("get-stream");
```
dphang, line 438: I am using get-stream package: https://bundlephobia.com/result?p=get-stream@6.0.0 which is small

packages/libs/lambda-at-edge/src/default-handler.ts
danielcondemarin:
```suggestion
import type { Readable } from "stream";
```

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -156,8 +156,10 @@ const router = (
 export const handler = async (
   event: OriginRequestEvent | OriginResponseEvent
 ): Promise<CloudFrontResultResponse | CloudFrontRequest> => {
+  // @ts-ignore
```
danielcondemarin, line 159: do we need these @ts-ignores ?

dphang: Hm, I saw errors when building at one point due to this being from json file, maybe this one is not needed anymore, let me try to remove it

packages/libs/lambda-at-edge/src/default-handler.ts
dphang: Oh, I thought Rollup would optimize this but I guess explicitly importing just type is good practice, too.

danielcondemarin (dismissed):
Just a few minor things but overall lgtm 👍

danielcondemarin (approved):

dphang:
Thanks for merging!

PR #590, opened 2020-09-07
Cloudfront custom error responses

rrooding:
Added support for https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages.html


danielcondemarin:
Interesting addition. I wonder if we could hook this up to [Next.js statically generated error pages](https://nextjs.org/docs/advanced-features/custom-error-page) rather than defining your own HTML error page. If that was achievable, there would be some performance benefits of doing so. For example, on a 404, CloudFront [wouldn't need to do a roundtrip](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/libs/lambda-at-edge/src/default-handler.ts#L149) to S3 to fetch the Next.js static error page. 

What do you think? cc. @dphang

rrooding:
I am not sure about that because the app we're developing doesn't have statically generated error pages as we use getInitialProps in _app. We need this however to have CloudFront catch for example 503 errors when a lambda times out. In that case the static page wold have to come from S3.

Interested in more input though!

danielcondemarin:
> I am not sure about that because the app we're developing doesn't have statically generated error pages as we use getInitialProps in _app. We need this however to have CloudFront catch for example 503 errors when a lambda times out. In that case the static page wold have to come from S3.
> 
> Interested in more input though!

Maybe the default behaviour should be for serverless-next.js to look for Next.js statically generated error pages and wire that up automatically to CloudFront error pages. 

Also users could override / extend this behaviour by adding their own error pages as per your PR,

```
errorPages: # optional
      - code: 503
        path: "/503.html"
        ...
```

Just thinking out loud btw, your PR seems like a good addition 👍

dphang:
Currently I made it so the 404 page returns correct status code of 404, and error page is rendered in case of SSR failure. Both of which are cached according to: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/HTTPStatusCodes.html. But these cover the cases when it fails in the request/response handlers (e.g routing did not find a path or the SSR failed)

This change is nice since it covers all the CloudFront errors (which may not come from the handlers, or maybe there is another error in handler that we did not cover by the SSR 500 page or 404 page).

@danielcondemarin for 404 pages, I haven't tried it but yes, it seems we could save an S3 roundtrip by just returning a 404 in origin request handler, and let CloudFront return its custom 404 page, which I assume will be on all edge nodes.

dphang:
Also error pages are specified in the CloudFront distribution using the path (from https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesErrorPages):

```
Response Page Path
The path to the custom error page (for example, /4xx-errors/403-forbidden.html) that you want CloudFront to return to a viewer when your origin returns the HTTP status code that you specified for Error Code (for example, 403). If you want to store your objects and your custom error pages in different locations, your distribution must include a cache behavior for which the following is true:

The value of Path Pattern matches the path to your custom error messages. For example, suppose you saved custom error pages for 4xx errors in an Amazon S3 bucket in a directory named /4xx-errors. Your distribution must include a cache behavior for which the path pattern routes requests for your custom error pages to that location, for example, /4xx-errors/*.

The value of Origin specifies the value of Origin ID for the origin that contains your custom error pages.
```

It's not completely clear if CloudFront is pre-caching. I guess not since they use the maximum of either the TTL for error pages (1 minute) or the Cache-Control header set on the page in the S3 bucket.

rrooding:
Thanks for your input. 
I've updated the PR with some input validation just now. But with regards to your other points I'm afraid my knowledge on the internals of this component are not sufficient.

dphang:
> Thanks for your input. 
> 
> I've updated the PR with some input validation just now. But with regards to your other points I'm afraid my knowledge on the internals of this component are not sufficient.

No worries :) I think that other stuff can be considered later.

danielcondemarin (approved):

README.md
```diff
@@ -180,6 +180,12 @@ myNextApplication:
               maxTTL: 10
               defaultTTL: 10
       priceClass: "PriceClass_100"
+      # You can add custom error responses
+      errorPages:
+        - code: 503
+          path: "/503.html"
+          ttl: 5 # optional, minimum ttl the error is cached (default 10)
```
dphang, line 187: minor: perhaps call this `minTTL` to be clearer and also consistent with the other CloudFront inputs (defaultTTL, minTTL, maxTTL`)

dphang (approved):

README.md
rrooding: No problem, fixed :)

dphang (approved):

PR #592, opened 2020-09-08
test(e2e-tests): add initial set of end-to-end tests

dphang:
## Description

This continues the work on: https://github.com/serverless-nextjs/serverless-next.js/issues/573.

* Adds errored pages to test 500 response
* Adds Cypress tests to test SSR/SSG pages, 404 pages, error pages, caching (SSR is not cached, SSG pages are cached, public files are cached)

Note: Cypress tests are written in Mocha, not Jest (current limitation of Cypress framework). Also I am using semantic naming on the app files so it's easier for developers to understand each test case just by looking at it.

## Tests

Ran it on the distribution here: https://dcrrk28otqvm3.cloudfront.net/

## TODO

Here's the remaining work:

* Figuring out how to verify static JS files are cached, as the path will change on each deployment. We can either (1) grab it from the pages' HTML response, or (2) the `.next` build directory for buildId + static JS chunks name. For pages, probably leaning to (1) so that the tests will be more "realistic" since they will read it from the HTML, and also it probably should not know about `.next` directory. For verifying all static files pushed to S3, we can just read it from output of `.next` build directory
* Create GH Actions workflow and use AWS access key to actually run this in the pipeline


packages/e2e-tests/next-app/cypress.json
```diff
@@ -2,5 +2,7 @@
   "baseUrl": "http://localhost:3000",
   "supportFile": "cypress/support/index.ts",
   "responseTimeout": 15000,
-  "requestTimeout": 15000
+  "requestTimeout": 15000,
+  "experimentalNetworkStubbing": true,
```
dphang, line 6: These two options allow using `route2()` for network level interception and also auto-polyfilling fetch

packages/e2e-tests/next-app/cypress/integration/caching.test.ts
```diff
@@ -0,0 +1,32 @@
+describe("Caching Tests", () => {
+  before(() => {
+    cy.ensureAllRoutesNotErrored();
+  });
+
+  describe("all static file requests for a page are cached", () => {
+    [{ path: "/" }].forEach(({ path }) => {
```
dphang, line 7: I made the test cases an object so that it's easier to add to in the future

danielcondemarin (approved):

Issue #581, opened 2020-09-02
JSON data files not properly cached


thchia:
**Describe the bug**
JSON data files for SSG pages (e.g. `_next/data/{build id}/file.json) are not being cached by Cloudfront, meaning that subsequent calls always go to the origin (and trigger the Lambdas@Edge).

**To Reproduce**
1. Deploy a website with at least one SSG page.
2. Visit the SSG page once. This should cache the JSON data.
3. Refresh that page and observe the `x-cache` response header set to `Miss from cloudfront`

**Expected behavior**
On step 3 above, we should observe `Hit from cloudfront`

**Additional context**
During the upload of these files, [no Cache Control is set](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/libs/s3-static-assets/src/index.ts#L114), which I suspect is the underlying cause. There should be no concern about caching these for a long time, since the build ID is in the path and SSG data should not vary within a single build.

dphang:
Nice find. This would be a good case to cover for e2e tests I'm working on, we can inspect the `x-cache` header after hitting it a second time to ensure it says `Hit from cloudfront`

dphang:
Looks like the cache-control header needs to be set, but also the CloudFront behavior for `_next/data/*` has a default, min and max TTL of 0, so cache-control header won't help.

We will need to fix the `aws-cloudfront` component to set these three values independently. In fact, it's also affecting the other CloudFront behaviors, max TTL is always 86400, so even for static files they'll get evicted from the CloudFront edge cache after 1 day, even though we are setting Cache-Control for those in S3.

Will work on a PR to fix it this week.

Issue #524, opened 2020-07-27
How to set individual TTL times (Min TTL, Max TLL, Default TTL)?


mattdell:
Following the docs in the Readme there is a section about how to set the `ttl`

```
myNextApplication:
  component: serverless-next.js
  inputs:
    cloudfront:
      # this is the cache behaviour for next.js api pages
      api:
        ttl: 10
```

Cloudfront actually has the options for setting min, max, and default ttl times in the UI:

<img width="1080" alt="Screenshot 2020-07-27 at 13 10 17" src="https://user-images.githubusercontent.com/2536442/88540334-a2ee5480-d00a-11ea-8793-14abc71a015a.png">

How would I go about setting the Max TTL **only** to 31536000?

It seems this was supposed to be fixed in PR #282 since it enables "all cloudfront inputs" but these are missing. Looking through the code, it seems to be a limitation in [aws-cloudfront inputs](https://github.com/serverless-components/aws-cloudfront#3-configure)

See how it takes in `ttl` and then assigns that to min, max, and default? https://github.com/serverless-components/aws-cloudfront/blob/master/lib/getCacheBehavior.js

I have raised an issue there but it doesn't appear to be very active.

Does anyone else have this problem?

medv:
Yeah this is relevant to us as well. I was meant to put some time into this when I have a chance but haven't had the opportunity yet. It's still on my list (that's currently still growing from the other end, ouchie)

teemuniiranen:
I didn't understand from the documentation how to change these. I tried to change the default static files ttl:
```
nextApp:
    component: "serverless-next.js@1.9.5"
    inputs:
      nextConfigDir: "../../"
      bucketName: my-hostingbucket-develop
      cloudfront:
        origins:
          - url: /static
            pathPatterns:
              static/*:
                ttl: 600
```

dphang:
I came across this problem while working on: https://github.com/serverless-nextjs/serverless-next.js/issues/581. It looks like even for default CloudFront behaviors, min, max and default TTLs are all the same. This inflexibility is also causing issues for static files, e.g TTL is all set to 86400 (1 day), so actually our Cache-Control headers of 31 days / 365 days is really only 1 day.

I am working on a fix this week for this and also allowing to set min, max, default TTLs independently. Luckily, `aws-cloudfront` component was already copied into this package, so we can make changes ourselves. As you noted, that repository under `serverless-components` doesn't seem active.

PR #593, opened 2020-09-08
feat(aws-cloudfront, s3-static-assets): support setting individual min, max, default CloudFront TTLs, update Cache-Control headers for _next/data files and update all CloudFront behavior TTLs [breaking change for custom CloudFront inputs]

dphang:
This fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/581 and https://github.com/serverless-nextjs/serverless-next.js/issues/524

* Update `aws-cloudfront` serverless component to allow setting individual default, min and max TTLs.
* Update CloudFront `_next/data/*` behavior to have min = 0, default = 0 and max = 31536000 for TTLs. Also adds Cache-Control header of `public, max-age=0, s-maxage=2678400, must-revalidate` for these.
* Update other CloudFront behaviors with the following TTLs:

APIs have min = 0, default = 0 and max = 31536000.
_next/static/* have min = 0, default = 86400, and max = 31536000
_static/* have min = 0, default = 86400, and max = 31536000
_next/data/* have min = 0, default = 0, and max = 31536000
default have min = 0, default = 0, and max = 31536000 (same as before)

I think the above makes sense, static files anyway do add cache-control headers, but in case it doesn't we cache 86400 seconds by default. Previously all three TTLs for static files were 86400, so our cache control headers were limited to 86400 even if we set it longer (e.g 31 days or 365 days). And for _next/data/ it was all 0 so they could not be cached. For default, it remains the same. See below for info:

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html

Note: there is a breaking interface change for CloudFront custom inputs, now it requires users to set individual TTLs instead of just `ttl`. I think this is acceptable since it's now more explicit and I believe not too many users are currently using the custom inputs?

## Tests

Updated the unit tests. I've also verified the TTLs are set correctly in CloudFront distributions. Caching behavior works as expected for SSR, SSG, and static files, and now also caches _next/data files.

You can validate yourselves here with these various paths:

https://d3q8tfp2sxrn2u.cloudfront.net/anotherSSG/
https://d3q8tfp2sxrn2u.cloudfront.net/anotherSSR/
https://d3q8tfp2sxrn2u.cloudfront.net/_next/data/eFK_r7Za-Hseex4Ydgrvt/anotherSSG.json
https://d3q8tfp2sxrn2u.cloudfront.net/app-store-badge.png

I'll also add these test cases to the e2e tests later.


danielcondemarin (approved):
LGTM 🚀

danielcondemarin:
Don't worry about the breaking change to the `aws-cloudfront` library. Is not really meant to be consumed outside of the serverless-next.js component.

PR #596, opened 2020-09-09
Fix S3 bucket name extraction from origin URL

pradyumna1:
As per AWS [docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html), bucket name can consist of periods. If an S3 bucket is being used as origin, we need to extract the string before s3 in `<bucket-name>.s3.amazonaws.com`. I came up with this simple solution as a quick fix, however, we can also use regex for extracting anything before s3.amazonaws.com from the hostname. 
For eg. `const bucketName = hostname.slice().replace(/(.*)\.s3\.amazonaws\.com/g, '$1')`


dphang:
Seems reasonable if it always starts with `s3`, maybe just add some simple tests if possible.

danielcondemarin:
I'd agree tests for this are needed to avoid regression later.

danielcondemarin:
https://github.com/serverless-nextjs/serverless-next.js/issues/761 looks related. I might jump later in this branch or create a new one to tidy this up.

dphang:
Closing due to inactivity for some time, if needed please reopen new PR with tests. Thanks!

PR #597, opened 2020-09-09
chore(release): publish

dphang:
- @sls-next/next-aws-cloudfront@1.5.0-alpha.3
 - @sls-next/lambda-at-edge@1.6.0-alpha.8
 - @sls-next/s3-static-assets@1.3.0-alpha.4
 - @sls-next/aws-cloudfront@1.2.0-alpha.0
 - @sls-next/serverless-component@1.17.0-alpha.10


dphang:
@danielcondemarin nvm about this PR, the npm package was not actually published so I think we need to delete the releases/tags and start over. Sorry about that! I am guessing npm was not updated because the git push failed.

PR #600, opened 2020-09-11
fix(lambda-at-edge): exclude prerender js files from default handler if no API routes are used

dphang:
## Description

This should ensure JS files used only for prerendering at build time are excluded from the default handler, if you aren't using API routes.

I was wondering why after converting my pages to SSG pages via `getStaticProps()`, the Lambda size still wasn't decreasing. It turns out that non-dynamic pages with `getStaticProps()` will generate JS files in the `.next` build directory, and `build.ts` wasn't excluding these JS files used only for prerendering SSG pages at build time.

By reading from `prerender-manifest.json`, we can figure out which routes are only used for prerendering.

It won't exclude JS files used for dynamic routes with getStaticProps + getStaticPaths, nor any JS files if API routes are present (due to possible use of preview mode - currently no proper way to determine a page is used for preview mode via the manifest). This should help improve performance by reducing the code size for apps that don't need API routes, for me it reduced from 15 MB zipped to 6 MB zipped.

This is in line with what Vercel does: https://github.com/vercel/next.js/discussions/15631#discussioncomment-44289

## Testing

Verified on my own app (no api routes), the e2e test app, added unit tests

Also tested on app with api routes: page with dynamic route and getStaticPaths() (will have JS included correctly): https://d2fn5f8m5hw83u.cloudfront.net/tags/128128218 (try other numbers)

Verified on Preview mode: https://d2fn5f8m5hw83u.cloudfront.net/preview and https://d2fn5f8m5hw83u.cloudfront.net/api/preview.

I'll add more e2e tests for these cases as well in another PR.


danielcondemarin:
If I remember correctly those files are being used for [pre-rendering on the Lambda](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/libs/lambda-at-edge/src/default-handler.ts#L390) when using `getStaticPaths` with `fallback: true`. Have you checked any pages with `getStaticPaths`?

dphang:
Yep, with @thchia 's test app I verified the dynamic pages with getStaticPaths aren't in prerender-manifest routes, but in dynamicRoutes so those JS files should be included correctly. (will add these test cases to the e2e tests)

dphang:
Here is the app: https://d2fn5f8m5hw83u.cloudfront.net/tags/128128218

dphang:
More info here: https://github.com/vercel/next.js/discussions/15631#discussioncomment-44289, basically these JS are only used in getStaticPaths fallback true and preview mode. Will test once again and confirm.

dphang:
Looks like it's failing on preview mode pages, will make a fix on that.

dphang:
Hm, seems there's no way to identify a page is used in preview mode from the manifest, as that page can still be in `routes` in `prerender-manifest.json`.

From the comment from Vercel team:

> The server JS file is always output for pages with getStaticProps since it is needed to prerender the page during the build, it is also used when leveraging preview mode. If you don't leverage API routes (required for enabling preview mode) and the pages are completely static (don't leverage fallback or revalidate support) with getStaticProps, lambdas won't be deployed for these pages on Vercel anymore though

Few different ways:

1. Keep this PR's solution but only do it if app does not have API routes, if not then app is definitely not using preview mode.
2. Have a list of files in `serverless.yml` that user can specify NOT to include prerender JS files for.
3. (Hacky way) inspect `getStaticProps` to see if it's trying to access a variable named `preview`, if so then make sure to include that page's JS file. Probably Next.js should determine a page is using preview mode and add it to build manifest.

@thchia @danielcondemarin any thoughts? Probably #2 gives the most flexibility, user can use API routes without preview mode and specify files they want to explicitly remove from `default-handler` to reduce bundle size.

danielcondemarin:
> Hm, seems there's no way to identify a page is used in preview mode from the manifest, and that page can still be in `routes` in `prerender-manifest.json`.
> 
> From the comment from Vercel team:
> 
> > The server JS file is always output for pages with getStaticProps since it is needed to prerender the page during the build, it is also used when leveraging preview mode. If you don't leverage API routes (required for enabling preview mode) and the pages are completely static (don't leverage fallback or revalidate support) with getStaticProps, lambdas won't be deployed for these pages on Vercel anymore though
> 
> Few alternative solutions:
> 
> 1. Keep this PR's solution but only do it if app does not have API routes, then it is definitely not using preview mode.
> 2. Have a list of files in `serverless.yml` that user can specify NOT to include prerender JS files for.
> 3. (Hacky way) inspect `getStaticProps` to see if it's trying to access a variable named `preview` or `context.preview`, if so then make sure to include that page's JS file. Probably Next.js should determine a page is using preview mode and add it to routes/prerender manifest.
> 
> @thchia @danielcondemarin any thoughts? Probably (2) gives the most flexibility and is safest, user can use API routes without preview mode and specify routes for which they want to explicitly remove the JS from `default-handler` to reduce bundle size. (1) seems to be in line with Vercel. I'm leaning to (1) by default since that's what Vercel does and then we could add on (2) for those using API routes. Personally, I don't use API routes right now, so either works for me.

Sol 1. feels the best to me, however we'd need to still include pages using `getStaticPaths` fallback `true` right?

dphang:
> > Hm, seems there's no way to identify a page is used in preview mode from the manifest, and that page can still be in `routes` in `prerender-manifest.json`.
> > From the comment from Vercel team:
> > > The server JS file is always output for pages with getStaticProps since it is needed to prerender the page during the build, it is also used when leveraging preview mode. If you don't leverage API routes (required for enabling preview mode) and the pages are completely static (don't leverage fallback or revalidate support) with getStaticProps, lambdas won't be deployed for these pages on Vercel anymore though
> > 
> > 
> > Few alternative solutions:
> > 
> > 1. Keep this PR's solution but only do it if app does not have API routes, then it is definitely not using preview mode.
> > 2. Have a list of files in `serverless.yml` that user can specify NOT to include prerender JS files for.
> > 3. (Hacky way) inspect `getStaticProps` to see if it's trying to access a variable named `preview` or `context.preview`, if so then make sure to include that page's JS file. Probably Next.js should determine a page is using preview mode and add it to routes/prerender manifest.
> > 
> > @thchia @danielcondemarin any thoughts? Probably (2) gives the most flexibility and is safest, user can use API routes without preview mode and specify routes for which they want to explicitly remove the JS from `default-handler` to reduce bundle size. (1) seems to be in line with Vercel. I'm leaning to (1) by default since that's what Vercel does and then we could add on (2) for those using API routes. Personally, I don't use API routes right now, so either works for me.
> 
> Sol 1. feels the best to me, however we'd need to still include pages using `getStaticPaths` fallback `true` right?

Yes, I think this will be in `dynamicRoutes` in prerender manifest instead of `routes`, as I believe `getStaticPaths` can only be used with dynamic SSG pages?

thchia:
Solution 1 also feels best to me. It’s the safest and most correct in terms of making sure we aren’t accidentally excluding needed files.

dphang:
Thanks all. Went with Solution (1) for now.

yarn.lock
```diff
@@ -7793,6 +7793,11 @@ get-stream@^5.0.0, get-stream@^5.1.0:
   dependencies:
     pump "^3.0.0"
 
+get-stream@^6.0.0:
```
dphang, line 7796: These shouldn't be related with my PR, but I guess these lockfile updates were somehow missed in previous commits.

danielcondemarin (approved):

PR #603, opened 2020-09-13
Control invalidation paths from cloudfront inputs

kartikag01:
Instead of Invalidation of all things `/*` which currently happening now as the default value fo paths in the CloudFront package, we can override paths by providing inputs from you

Idea is that when we deploy its invalidate all things, there might different many different CloudFront behaviors users have created, it shouldn't invalidate them directly instead give control to the user via yml file.


dphang:
Thanks, it seems like a good idea. Please test and also update any unit tests as needed.

danielcondemarin:
@kartikag01 Are you still interested on this feature? Otherwise will close it to tidy PRs up :)

kartikag01:
@danielcondemarin Yes, its really required in our system, may be also required by many other. invalidating all things is not good idea.

jvarho:
This would be very useful. Also, how about adding a way to skip invalidation completely while at it? That would be nice when you know nothing cacheable has changed.

danielcondemarin:
> @danielcondemarin Yes, its really required in our system, may be also required by many other. invalidating all things is not good idea.

Thanks for getting back to me 👍  Happy to merge if you add some tests please. 

> This would be very useful. Also, how about adding a way to skip invalidation completely while at it? That would be nice when you know nothing cacheable has changed.

I don't see why not 👍

dphang:
Yeah, sometimes you may also only need to invalidate page routes and _next/data, if everything else is versioned e.g using an MD5 hash (for example, I have images in _next/static/ which is versioned). If you invalidate all paths, then these would need to be refetched.

danielcondemarin:
Fixed by https://github.com/serverless-nextjs/serverless-next.js/pull/603

Issue #454, opened 2020-06-21
Getting 403 errors on some of the POST requests


pbteja1998:
**Describe the bug**
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<HTML><HEAD><META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=iso-8859-1">
<TITLE>ERROR: The request could not be satisfied</TITLE>
</HEAD><BODY>
<H1>403 ERROR</H1>
<H2>The request could not be satisfied.</H2>
<HR noshade size="1px">
This distribution is not configured to allow the HTTP request method that was used for this request. The distribution supports only cachable requests.
We can't connect to the server for this app or website at this time. There might be too much traffic or a configuration error. Try again later, or contact the app or website owner.
<BR clear="all">
If you provide content to customers through CloudFront, you can find steps to troubleshoot and help prevent this error by reviewing the CloudFront documentation.
<BR clear="all">
<HR noshade size="1px">
<PRE>
Generated by cloudfront (CloudFront)
Request ID: S9Bit4WZWnEar0Bo6NX7h-QbmHFUrnBF79lUp2LDYiLFoVPbw_bkTg==
</PRE>
<ADDRESS>
</ADDRESS>
</BODY></HTML>

I have suddenly started getting 403 error on some of the POST requests in my application.

My serverless.yml file for reference.
```
myNextApplication:
  component: serverless-next.js@1.14.0
  inputs:
    domain:
      - ${env.subdomain}
      - ${env.domain}
  build:
    cmd: yarn build
```

danielcondemarin:
@pbteja1998 Is it an HTTP Post to a next API route?

pbteja1998:
@danielcondemarin No, the POST request is to a backend Node Server.

pbteja1998:
Everything was working till yesterday

danielcondemarin:
> @danielcondemarin No, the POST request is to a backend Node Server.

Where do you host this backend node server? I don't see how it has anything to do with serverless-next.js unless I'm missing something,

pbteja1998:
> > @danielcondemarin No, the POST request is to a backend Node Server.
> 
> Where do you host this backend node server? I don't see how it has anything to do with serverless-next.js unless I'm missing something,

[...]

PR #609, opened 2020-09-18
fix(nextjs-component, e2e-tests): allow all CloudFront HTTP methods for default caching behavior

dphang:
This fixes https://github.com/serverless-nextjs/serverless-next.js/issues/454 and similar issues.

Not sure why we don't allow all CloudFront methods on default caching behavior, which will allow more HTTP methods e.g form posts on SSR pages, similar to Vercel.

This component already adds a policy to the S3 bucket to only allow GetObject from the origin access identity (CloudFront), so requesters cannot modify S3 objects with these additional methods (e.g using DELETE to delete a public file). See: https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_AllowedMethods.html

## Tests

Updated unit and e2e tests, and ran them (e2e tests not yet automated).

SSR page: https://dcrrk28otqvm3.cloudfront.net/ssr-page (allows GET, HEAD, PUT, PATCH, DELETE, POST, OPTIONS). E.g POST/DELETE will return page itself since there is no different behavior here for POST/DELETE

For resources in S3:
SSG page: https://dcrrk28otqvm3.cloudfront.net/ssg-page (allows GET, HEAD. OPTIONS requires origin request header)
Public files: https://dcrrk28otqvm3.cloudfront.net/app-store-badge.png (allows, GET, HEAD. OPTIONS requires origin request header.

The above two, for POST gets 405 Method Not Allowed, and for PUT, PATCH, DELETE gets 403 Forbidden, which should be due to the S3 bucket policy only allow GetObject from the CloudFront distribution.

Static JS (same as before, it cannot be DELETED, as _next/static cache behavior only allows GET, HEAD): https://dcrrk28otqvm3.cloudfront.net/_next/static/chunks/framework.085e84bea8b122ad7b41.js

For comparison:

Vercel SSR page: https://nextjs-repros.vercel.app/anotherSSR (allows GET, HEAD, PUT, PATCH, DELETE, POST, OPTIONS and even more uncommon methods, since they do not use CloudFront but their own custom CDN).
Vercel SSG page: https://nextjs-repros.vercel.app/anotherSSG (allows GET, OPTIONS, HEAD)

### DIFF
--- a/packages/e2e-tests/next-app/cypress/integration/pages.test.ts
+++ b/packages/e2e-tests/next-app/cypress/integration/pages.test.ts
@@ -13,6 +13,16 @@ describe("Pages Tests", () => {
 
         cy.visit(path);
       });
+
+      ["HEAD", "DELETE", "POST", "GET", "OPTIONS", "PUT", "PATCH"].forEach(
+        (method) => {
+          it(`allows HTTP method: ${method}`, () => {
+            cy.request({ url: path, method: method }).then((response) => {
+              expect(response.status).to.equal(200);
+            });
+          });
+        }
+      );
     });
   });
 
@@ -31,6 +41,16 @@ describe("Pages Tests", () => {
 
         cy.visit(path);
       });
+
+      ["HEAD", "DELETE", "POST", "GET", "OPTIONS", "PUT", "PATCH"].forEach(
+        (method) => {
+          it(`allows HTTP method: ${method}`, () => {
+            cy.request({ url: path, method: method }).then((response) => {
+              expect(response.status).to.equal(200);
+            });
+          });
+        }
+      );
     });
   });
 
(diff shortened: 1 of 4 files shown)


danielcondemarin (approved):

danielcondemarin:
> Not sure why we don't allow all CloudFront methods on default caching behavior, which will allow more HTTP methods e.g form posts on SSR pages, similar to Vercel.

The reasoning was that the default routes would only serve pages, therefore HEAD, GET is sufficient. For an HTTP POST I believe Next.js recommends using an API route although may be possible to do via getInitialProps judging from [this thread](https://github.com/vercel/next.js/issues/7074). Having said all that I don't see much downside in allowing the other HTTP methods 👍

dphang:
> > Not sure why we don't allow all CloudFront methods on default caching behavior, which will allow more HTTP methods e.g form posts on SSR pages, similar to Vercel.
> 
> The reasoning was that the default routes would only serve pages, therefore HEAD, GET is sufficient. For an HTTP POST I believe Next.js recommends using an API route although may be possible to do via getInitialProps judging from [this thread](https://github.com/vercel/next.js/issues/7074). Having said all that I don't see downside in allowing the other HTTP methods 👍

Thanks for taking a look. Yeah, I can't imagine this is common since there are API routes or you would use an external API (like I am), but maybe someone has a use case for other methods, e.g POST of form to the same page, or they want to explicitly render a different page if non-GET method is called.

PR #621, opened 2020-09-24
test(e2e-tests): add e2e tests step in CI

dphang:
This adds e2e tests in the CI workflow. I tested it locally and it works fine, hopefully it works in the CI too (note: it is currently failing due to permissions issue in AWS). This closes https://github.com/serverless-nextjs/serverless-next.js/issues/573.

* It uses a script to:
1. Sync `.serverless` state (to reuse the same distribution) from a deployment S3 bucket.
2. Wait for CloudFront invalidations to complete, Lambda to update, S3 to update, based on the Next.js `BUILD_ID`.
3. Run the Cypress tests (`yarn e2e`) with env variables set for CloudFront URL and `BUILD_ID`.
4. Syncs back `.serverless` state and cleans up app files in `S3`. The Lambda and CloudFront distribution are untouched for future reuse.
* This is run in the 12.x workflow only, and additionally gated using https://github.com/softprops/turnstyle to ensure only one instance is running at a time.
* There is mixed use of aws-cli and aws-sdk, reason being that S3 commands like emptying a bucket or syncing a folder are one-liners in aws-cli, but take much more code using aws-sdk.

Some improvements in the future but it's not needed now as there aren't too many pull requests anyway:
* Manage multiple PR workflows running concurrently - we would need to manage serverless state across multiple CloudFront distributions.
* Update e2e tests to add more coverage of API routes, preview mode, etc.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,3 +34,17 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn test
       - run: yarn integration
+      - name: Wait for existing workflow to complete before e2e tests
+        if: ${{ matrix.node-version == '12.x' }}
+        uses: softprops/turnstyle@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run e2e tests
+        if: ${{ matrix.node-version == '12.x' }}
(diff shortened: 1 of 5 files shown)


danielcondemarin (approved):
Awesome work

Issue #587, opened 2020-09-07
RFC: Redirects/rewrites/headers support


dphang:
Just starting this thread about custom redirects/rewrites/headers support. It seems to be the most requested feature in Next.js 9.5 as of now, so I think it makes sense to support this next.

Next.js docs:

Redirects: https://nextjs.org/docs/api-reference/next.config.js/redirects
Rewrites: https://nextjs.org/docs/api-reference/next.config.js/rewrites
Headers: https://nextjs.org/docs/api-reference/next.config.js/headers

I am grouping these three features together they are similar and were added at the same time:

* The data for this is stored in `routes-manifest.json` after Next.js build, so it's already accessible in the handler code. Note, Next.js does have the regexes stored here as well.
* They all support hardcoded routes + path matching: simple matching, wildcard matching, and regex matching.

Note that there is existing code that I wrote to handle trailing slash redirects (which is **not** using the default-added redirect in `routes-manifest.json` right now for perf reasons)

Since we are now using Rollup.js to bundle the handlers, we can write separate files for better maintainability. I suggest the following steps:

* Create a matcher class that will do all matching including simple path matching, wildcard path matching, and regex matching. This code is shared with header/rewrite/redirects and other functionality that may need it in the future.
* Redirects/rewrites: in origin request handler, create new functions/classes to handle rewrites/redirects. For example, for redirects, we would (1) refactor existing trailing slash redirect code into this class, and then (2) add support for custom redirects. Similar for rewrites. These would be done close to the start (e.g replacing the redirect code).
* Headers: create new functions/class to add headers and use it in origin response handler (public files, HTML pages). I don't believe it's needed in origin request handler (SSR pages) as you can just set `res` headers in the page code itself.
* In each of these classes (redirect/rewrite/headers), at a high level, we would:

1. Load `routes-manifest.json` for the `redirects`/`rewrites`/`headers` list.
2. Check custom paths against the matcher.
3. Do the appropriate action: e.g return redirect response (with correct status code), rewrite the URI, or add a header.

I imagine the tricky part would be the path matching.

I'm already working on the e2e tests, so if someone wants to start on this or has any thoughts on the design, feel free to add your thoughts here. If not, I will work on this sometime later this month.

@danielcondemarin FYI.

dphang:
Update: sorry for the delay, I have been busy helping stabilize this component with end-to-end tests and other bug fixes. I will start working on the matcher next (probably this weekend or so)

[...]

PR #627, opened 2020-09-26
feat(lambda-at-edge): support custom redirects

dphang:
This is the first part of https://github.com/serverless-nextjs/serverless-next.js/issues/587. Summary:

* Add path matcher code using `path-to-regexp` for custom redirects (same one as Next.js uses). Some of this can also be reused for custom rewrites and headers.
* Add support for custom redirects in `next.config.js`. Custom redirects are matched one-by-one from the provided redirects in `routes-manifest` and the first matching redirect is returned. The default trailing slash redirects are ignored for now (by removing them at build time) as they are already handled in a non-regex solution. **I also need to put the redirect logic near the `default-handler` start right after trailing slash redirect, as redirects can overwrite an existing source page (same behavior in Vercel)**.
* There may be slight performance concerns here, but AFAIK `path-to-regexp` uses RegExp underneath which should be quite performant. I would only expect there to be performance issues if users have really long lists of redirects and/or are using complex paths. For the majority of users, this shouldn't be the case - and the default trailing slash redirects are anyway handled without regex.
* Caveat: redirects (both custom and trailing slash) don't work with API route sources yet (will add in later PR) - Vercel supports this, although I imagine it may be fairly uncommon to redirect an API route.

Note: big PR but a lot of it is updating the tests to have good coverage.

## Testing

Added lots of test coverage via unit tests, integration tests, and e2e tests.

Note: there is no e2e test with base path and trailingSlash = true, although I have locally tested it and unit/integration tests should cover it. I will add in a future PR as part of larger work for more e2e tests.

If anyone can pull this branch and test this and/or suggest other test cases I may have missed, that will be really helpful. I think I covered most of them. Thanks!


packages/libs/lambda-at-edge/src/routing/redirector.ts
```diff
@@ -0,0 +1,109 @@
+import { compileDestination, matchPath } from "./matcher";
+import { RedirectData, RoutesManifest } from "../../types";
+import * as http from "http";
+
```
dphang, line 4: Now that we have Rollup, we can make the component more maintainable by using different files - it will all get bundled into the handler JS files. Not sure what's a good name, let me know if you have thoughts on this.

packages/libs/lambda-at-edge/src/routing/redirector.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Whether this is the default trailing slash redirect.
+ * @param redirect
+ * @param basePath
+ */
+function isTrailingSlashRedirect(redirect: RedirectData, basePath: string) {
```
benjipott, line 10: This can me more simple with regex, and prevent lot of 'if'

benjipott (commented):
Great work 👏

packages/libs/lambda-at-edge/src/routing/redirector.ts
dphang: Thanks! Yeah, it's a bit ugly but basically trailing slash redirects are already handled by different logic, so I wanted to skip these default redirects from routes-manifest. I wanted to avoid additional RegExp here for perf reasons, so this might be the best way for now.

benjipott: Ok, i can test it tomorrow morning

dphang: Thanks, that will be helpful. I will try to clean up the code a bit more and see if there are other optimizations I can make.

packages/libs/lambda-at-edge/src/routing/redirector.ts
danielcondemarin: I think `routing/` makes sense. On a related note, would be nice if at some point we refactored all routing concerns into a `router`, similar to what the [next-server](https://github.com/vercel/next.js/blob/canary/packages/next/next-server/server/router.ts) impl does

packages/libs/lambda-at-edge/src/routing/redirector.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @param path
+ * @param routesManifest
+ */
+export function getRedirectPath(
+  path: string,
+  routesManifest: RoutesManifest
+): { redirectPath: string; statusCode: number } | null {
+  const redirects: RedirectData[] = routesManifest.redirects;
+
+  for (const redirect of redirects) {
```
danielcondemarin, line 54: Not one for now, but might be worth looking into trie router implementations out there which would scale better than this. There is an example impl. here https://github.com/pillarjs/routington

danielcondemarin (approved):

packages/libs/lambda-at-edge/src/routing/redirector.ts
dphang: Sure, we can look into that to improve routing in general

PR #631, opened 2020-09-28
chore: add --create-release github to publish commands

dphang:
I think this is needed so that when maintainers publish, GitHub releases are updated properly e.g with the changelog and "latest" release properly reflected. This `--create-release github` flag is part of https://github.com/lerna/lerna/tree/master/commands/version#--create-release-type (lerna publish also includes all lerna version flags as per https://github.com/lerna/lerna/tree/master/commands/publish#options)

Currently https://github.com/serverless-nextjs/serverless-next.js/releases doesn't have any changelogs and the main GitHub page is showing an outdated May 2019 release as the latest:

Screen Shot 2020-09-27 at 10 50 19 PM

This will help users better understand what changes went into each new release.

There's no easy way to do a lerna dry-run, but I will test it out the next time I publish.

@danielcondemarin let me know if this makes sense?


danielcondemarin (approved):
Good find! Yeah this makes sense 👍

Issue #637, opened 2020-09-29
Issue using Redirects in next.config.js


arpetti:
### Describe the bug

After updating to the latest alpha release, I received a 503 error on paged affected by the next.config redirects.

### Actual behavior

All the website pages are rendered correctly, but trying to access a page affected by one of the redirect rules defined in next.config, the following error appears in the browser:

> 503 ERROR
> The request could not be satisfied.
> The Lambda function associated with the CloudFront distribution is invalid or doesn't have the required permissions. We can't connect to the server for this app or website at this time. There might be too much traffic or a configuration error. Try again later, or contact the app or website owner.
> If you provide content to customers through CloudFront, you can find steps to troubleshoot and help prevent this error by reviewing the CloudFront documentation.
> Generated by cloudfront (CloudFront)
> Request ID: T8jK_Pu8pNiPLXKMOgMI8TOspmZMtr4m2__1fUKC4GrHvcvjXPI7bw==

### Expected behavior

Redirect to the correct page

### Steps to reproduce

Add redirects in next.config and deploy using serverless component

### Screenshots/Code

In next.config I've added:
```
 async redirects() {
    return [
      {
        source: '/professor/:slug*',
        destination: 'https://example.com/professor/:slug*',
        permanent: true,
      },
      {
        source: '/aluno/:slug*',
        destination: 'https://example.com/aluno/:slug*',
        permanent: true,
      },
      {
        source: '/corretor/:slug*',
        destination: 'https://example.com/corretor/:slug*',
        permanent: true,
      },
    ];
  },
```

My serverless.yml

```
letrusWebsiteServerless:
  component: '@sls-next/serverless-component@1.17.1-alpha.1'
  plugins:
    - serverless-plugin-split-stacks
  inputs:
    domain: ['mydomain.com']
    bucketName: my-bucket-name
    bucketRegion: sa-east-1
    runtime: nodejs12.x
    cloudfront:
      distributionId: xxxxxxxxxxxxxxxxxx
    name:
      defaultLambda: DefaultLambda
      apiLambda: ApiLambda
custom:
  splitStacks:
    perFunction: true
    perType: false
    perGroupFunction: false

```
### Versions

- OS/Environment:
- @sls-next/serverless-component version:  1.17.1-alpha.1
### Checklist

- [x] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
[...]

PR #638, opened 2020-09-30
fix(lambda-at-edge): fix broken redirects with full URLs as destinations

dphang:
This fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/637

* Handle URLs as destinations such as `https://example.com` or `http://example.com`

## Tests

Updated all tests.


packages/libs/lambda-at-edge/tests/routing/redirector.test.ts
```diff
@@ -25,18 +25,32 @@ describe("Redirector Tests", () => {
...
+          },
+          {
+            source: "/external",
+            destination: "https://example.com",
+            statusCode: 308,
+            regex: "^/external$"
+          },
+          {
+            source: "/invalid-destination",
+            destination: "ftp://example.com",
```
dphang, line 37: Right now just supporting https:// and http:// protocol, even though this is a valid URL. Let me know if you are other thoughts

danielcondemarin (approved):

Issue #633, opened 2020-09-28
Question: How to redirect www to non-www with AWS


alan345:
With AWS, Can you guide me to redirect www to non-www to avoid duplicate content?
Should we create a s3 bucket and a new Cloud front distribution? (https://www.pbxdom.com/how-redirect-non-www-to-www-with-amazon-cloud-front-route-53-and-s3)

or, can we set it up in the `serverless.yml` file?

Thanks!

dphang:
I think you can follow here for now: https://github.com/serverless-nextjs/serverless-next.js#i-was-expecting-for-automatic-subdomain-redirection-when-using-the-domaintype-wwwapex-input. Basically you need to manually create new S3 bucket and CloudFront distribution (for HTTPS redirect) and some new Route53 A records

In the future it's possible to support it in this component, but we would need to make a couple of changes:
* Support multiple domain inputs mapped to one distribution (so you can have www.example.com and example.com Route53 records pointing to same CF distribution)
* Lambda could return a redirect response from www.example.com -> example.com (or vice versa)

dphang:
Now that redirects are done (https://github.com/serverless-nextjs/serverless-next.js/issues/587#issuecomment-700902584) we could leverage similar code to do a redirect from a domain name to another in the Lambda@Edge itself.

Although it's just a little bit costlier than the S3 bucket redirect (due to use of Lambda@Edge compute), it has a few advantages:

* The redirect is handled at the edge server instead of at a single point (e.g S3 bucket). This should have lower latency for most users.
* Easier to manage within this component; no need to setup two separate CF distributions and another bucket.

Caveat: currently only the default cache behavior (`/*`) have at Lambda@Edge association, so you won't be able to redirect static files under `/_next/static/*`. We could solve in the future by adding another handler for this cache behavior, but probably not worth it just for redirects.

alan345:
thanks for your update. I have as recommended manually create new S3 bucket and CloudFront distribution and some new Route53 A record.

I have 2 CloudFront distribution :
- example.com
- www.example.com (s3 redirection)

It works well.

Now, if I deploy again via `npx serverless` I have this error:
`nextApp › CNAMEAlreadyExists: One or more of the CNAMEs you provided are already associated with a different resource.`

dphang:
Yeah, I think you may need to set `domainType` input to `apex` only since `www.example.com` is already associated with your CF distribution for redirection. Let me know if that works.

dphang:
[...]

PR #639, opened 2020-09-30
feat(lambda-at-edge, nextjs-component): add new input domainRedirects

dphang:
This new input supports redirection at domain-level e.g you can redirect from `www` domain to `apex` domain (or from/to other domains as well), i.e fixes this: https://github.com/serverless-nextjs/serverless-next.js#i-was-expecting-for-automatic-subdomain-redirection-when-using-the-domaintype-wwwapex-input.

* This uses a 308 redirect with refresh header for IE11 compatibility. It is the same as how Vercel does it (try to go to https://www.vercel.com/ and it will do a 308 redirect as well)
* Works with default handler and api handler (so just `/api/*`, `/*` and `/_next/data/*` paths)
* Meant to be used with `Host` forwarding and `domain = ["example.com"]` and `domainType = "both"`, although other domains can be used (although right now you can't configure other domains through this component doesn't support multiple domains yet besides `www` and `apex` types. However it's possible with Terraform, etc.).
* The check for domain-level redirect happens near the start of the handler and checks an object key, so it should be fast.
* This was based on ideas in: https://github.com/serverless-nextjs/serverless-next.js/issues/633#issuecomment-700985463. 

Although it is slightly costlier than the S3 bucket redirect (due to use of Lambda@Edge compute), it has a couple of advantages:

* The redirect is handled at the edge CloudFront server instead of at a region (e.g S3 bucket in a single region). This would generally mean lower latency for all users (apart from cold start).
* Built-in and easier to manage within this component; no need to setup two separate CF distributions and another bucket. Doing the redirect in Lambda means we can also change the status code or add headers if we want to later (but probably not necessary). Users can still use the S3 redirect if they want by managing their own CF distribution (via Terraform or otherwise).

@danielcondemarin what are your thoughts on this approach?

## Testing

Updated the unit tests. The e2e tests don't have a domain at the moment so I can't add tests on those yet. But I have tested it on my own distribution. Here's an example configuration after this change, which is meant to redirect `www.example.com` to `example.com`:

```yml
next-app:
  component: "../../serverless-components/nextjs-component"
  inputs:
    cloudfront:
      defaults:
        forward:
          headers:
            [
                Host
            ]
    domain: ["example.com"]
    domainType: "both"
    domainRedirects:
      www.example.com: https://example.com
```


danielcondemarin:
> @danielcondemarin what are your thoughts on this approach?

I remember looking into this a while back and doing at the edge seemed like the most practical solution 👍

danielcondemarin (approved):

dphang:
Already approved, was just merging changes

Issue #648, opened 2020-10-05
Add `handler` as an input option


emhagman:
Is it possible to add / does there already exist a way to set a custom handler via the Serverless component?

We are trying to setup https://docs.datadoghq.com/serverless/installation/nodejs/?tab=custom for our metrics and not being able to set a handler is preventing us from tracing our functions.

Essentially we'd like to use set the handler to `node_modules/datadog-lambda-js/dist/handler.handler` and then set the `default-handler` to the environment variable `DD_LAMBDA_HANDLER` like in the documentation in order to hook into the function.

Another example would be setting up Sentry: https://docs.sentry.io/platforms/node/guides/aws-lambda/ It seems like it would be impossible to use any 3rd party tools without something like this.

EDIT: Having looked at the code now, I am just curious if we could override https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-components/nextjs-component/src/component.ts#L373 via `inputs` ?

luismanuel001:
@emhagman I'm currently in the middle of this same scenario, trying o integrate DD with this, could you by any chance share a sample of your setup?

laerte-sousaneto:
@emhagman Also trying to integrate with DD and having no luck. Could you share an example?

PR #649, opened 2020-10-05
feat(lambda-at-edge, nextjs-component): allow `handler` input for custom handler code

emhagman:
Allow custom Lambda handler to be set. See https://github.com/serverless-nextjs/serverless-next.js/issues/648

Set the handler via `inputs.handler` and has `lambda-at-edge` copy `someCustomFile.js` file from `this.nextConfigDir` into the Lambda function folder by reading the first part of `someCustomFile.handler` and copying via Builder underneath:

```
new Builder(..., ..., { handler: 'someCustomFile.js' })
```

Useful for hooking into 3rd party tracing / error handling libraries such as:
https://docs.sentry.io/platforms/node/guides/aws-lambda/

For example:

```
example:
  component: "@sls-next/serverless-component@next"
  inputs:
    handler: someCustomFile.handler
```

### {nextConfigDir}/someCustomFile.js
```
const Sentry = require("@sentry/serverless");
Sentry.init({ dsn: "https://5a167080eb1c4ecea2a82f38ea16e024@o149282.ingest.sentry.io/1196981" });

const lambdaAtEdge = require("./index"); // index.js file created by lambda-at-edge
exports.handler = Sentry.AWSLambda.wrapHandler(lambdaAtEdge.handler);
```


emhagman:
12.x build seems to be failing due to S3 error

dphang:
> 12.x build seems to be failing due to S3 error

Don't worry about that, it's because forks don't have access to AWS keys, I will check how we can detect this is a fork and skip this. I thought I had added a check but it seems it may not be working.

emhagman:
Sounds good! Ideally we could add an example custom handler somewhere in the docs but I wasn't sure where to put that or if I should start a new section.

emhagman:
@dphang Yes, it overrides both handlers. I could separate it out if desired

dphang:
> @dphang Yes, it overrides both handlers. I could separate it out if desired

I think it's ok now, any thoughts @danielcondemarin?

danielcondemarin:
> > @dphang Yes, it overrides both handlers. I could separate it out if desired
> 
> I think it's ok now, any thoughts @danielcondemarin?

Probably best to make it consistent with the other lambda inputs like [memory](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-components/nextjs-component/src/component.ts#L381), name etc. See [readLambdaInputValue](https://github.com/serverless-nextjs/serverless-next.js/blob/cbd26bb43b67cbbd302053a8b8d28df586aa4f77/packages/serverless-components/nextjs-component/src/component.ts#L323)

kulesa:
@danielcondemarin a custom handler probably doesn't need lambda inputs because it does not add a new lambda function. It is only a wrapper around the default lambda and it gets bundled and deployed with the default lambda handler.

@emhagman @dphang can I contribute to this pull request in any way to get it merged? We're about to start using the fork of the component with custom handler in production 😄 , so I really hope this feature makes it to the library eventually.

dphang:
Yeah if you fix the conflicts I think I'm fine with it

emhagman:
I fixed the conflicts. One thing to note is that now that we have "minifyHandlers" as an option, this custom handler does not follow that option since it appears that all that flag does is copy the `.min.js` version of the default handlers.

We should either say custom handlers aren't minified or pull in a tool to minify them separately.

@dphang Thoughts?

dphang:
> I fixed the conflicts. One thing to note is that now that we have `minifyHandlers` as an option, this custom handler does not follow that option since it appears that all that flag does is copy the `.min.js` version of the default handlers over if enabled.
> 
> We should either say somewhere in the docs that custom handlers aren't minified or pull in a tool to minify them separately.
> 
> @dphang Thoughts?

Yeah, I think we should just update the docs for that (you can do in another PR or I can update them). If you want, you can add Terser in another PR to minify custom handlers.

I did think about minifying at build time but it makes more sense to minify in the distributed component since it does add a few seconds to build time. I figure most people won't care about how it's minified I had provided a sensible default set of Terser options.

dphang:
Ran e2e-tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/316875361

dphang (approved):

emhagman:
Sounds good! I can open another PR for the docs.

Every year I do https://hacktoberfest.digitalocean.com as part of a Fall tradition. Do you mind adding `hacktoberfest-accepted` as a label to this PR if you think this PR was helpful to the project so I can have it counted.

mattbbc:
Hi @emhagman, were the docs updated to show this off? I would like to write a custom handler to redirect users to a certain endpoint determined by CloudFront geolocation headers and this seems like a good fit.

ekasprzyk:
Hello!  @emhagman or @dphang 
I managed to get a custom handler working for logging, but there's a critical bit I'm missing around imports and I was wondering if either of you could help?  (As there really are no docs for this anywhere else)

I have the following code that works in deploying a handler:

```
const lambdaAtEdge = require('../index');

const handler = async (event, context) => {
  console.log('Pre-handler');
  const result = lambdaAtEdge.handler(event, context);
  console.log('Post-handler');

  return result;
};

exports.handler = handler;
```
I put the handler in the /src directory which is why the `..` exists on the import.

I'm trying to add in a Lumigo tracer, which seems similar to Sentry.  In your example you just require it, so I've tried the same:

```
const lumigo = require('@lumigo/tracer')({ token: 'REDACTED' });
const lambdaAtEdge = require('../index');

const handler = async (event, context) => {
  console.log('Pre-handler');
  const result = lambdaAtEdge.handler(event, context);
  console.log('Post-handler');

  return result;
};

exports.handler = lumigo.trace(handler);
```
and I'm not having any joy.  It fails with:

```
Uncaught Exception {"errorType":"Runtime.ImportModuleError","errorMessage":"Error: Cannot find module '@lumigo/tracer'
```

What am I missing?  I notice the example links to Sentry's docs which has a webpack config entry attached to it, but I don't have any similar thing to use from Lumigo.

Thanks in advance.

ekasprzyk:
Just to say, I figured out a solution.  For anyone else having the same issues as I am and who's not hugely hot on Serverless/NextJS internals or done a lot of webpack config (as is the case with me):

In order to use a wrapper (I used [Lumigo](https://lumigo.io/)) you need to set the handler as documented here, but the code you run is bundled elsewhere and separate from the lambda you're attempting to run, so adding your dependency as a project dependency isn't going to work.  

What you have to do is actually:
1. Add the handler
2. Bundle the handler's dependencies separately with webpack
3. Add the bundle to your lambdas for upload to the cloud service provider
4. Include the bundle in your code

e.g. If I want to do:

```
const lambdaAtEdge = require('../index');
const lumigo = require('@/lumigo/tracer')({token: token});

const handler = async (event, context) => {
  console.log('Pre-handler');
  const result = lambdaAtEdge.handler(event, context);
  console.log('Post-handler');

  return result;
};

exports.handler = lumigo.trace(handler);
```
then what I have to do is actually:

1.
```
const lambdaAtEdge = require('../index');
const wrapWithLumigo = require('../lumigo-wrapper');

const handler = async (event, context) => {
  console.log('Pre-handler');
  const result = lambdaAtEdge.handler(event, context);
  console.log('Post-handler');

  return result;
};

exports.handler = wrapWithLumigo.wrapWithLumigo(handler);
```

2.
```
const path = require('path');

module.exports = {
  entry: './handler.js',
  output: {
    path: path.resolve(__dirname, 'handler_dist'),
    filename: 'lumigo-wrapper.js',
    library: {
      name: 'lumigo-wrapper',
      type: 'umd',
    },
    globalObject: 'this',
  },
};
```
which refers to:
```
const lumigo = require('@lumigo/tracer')({ token: secret_token});

export function  wrapWithLumigo (handler) {
  return lumigo.trace(handler);
}
```
3.
Copy the bundle and deploy using the postBuild script:
https://github.com/serverless-nextjs/serverless-next.js/issues/767#issuecomment-722967719

e.g.
```
const fs = require('fs-extra');

console.log('-> Copying lumigo tracer...');
const wrapperSrc = './serverless_files';
const wrapperDefaultDest = './.serverless_nextjs/default-lambda';
const wrapperApiDest = './.serverless_nextjs/api-lambda';
fs.copySync(wrapperSrc, wrapperDefaultDest, { recursive: true });
fs.copySync(wrapperSrc, wrapperApiDest, { recursive: true });
console.log('Lumigo tracer was copied successfully');
```
(This assumes I manually copied the result of the dist bundle into `serverless_files`)

The bundle generated in 2. is deployed in 3. and picked up in the handler in 1.

Hope this helps.

J3tto:
@ekasprzyk Did you ever get Lumigo Tracing working on the api-lambda. 

Using the process you suggest the other all work but get `Cannot read property 'headers' of undefined` on the api-lambda. 

It seems to occur after the original handler completes so it doesn't look to be an issues on the serverless-nextjs side.

ekasprzyk:
@J3tto  Yes we did, and started to expand our Lumigo tracing into another application, where it worked (except for a bug supposedly on the Lumigo end causing stats not to be reported, which is apparently fixed).

We are using next/serverless versions slightly behind the current latest (10/1.x) so that might contribute.  

The fact that it's happening after the original handler returns is weird.

PR #655, opened 2020-10-06
Adding manual ACM SSL ARN specification.

dbhagen:
Hello,

Did a quick patch to allow manual ACM cert specification as I wanted it to use a wildcard cert that I had already generated. It is a pretty simple change, adding a `certificateArn` input variable and then processing if it was provided to avoid the search loop. Likely could do with some more logic checks, but since it still runs through the rest of the checks from the domain search, it seemed that keeping it minimal was better.


dbhagen:
I should mention that I performed tests, including the E2E specified in the contributor's doc. Everything seemed to pass.

dphang (approved):
LGTM!

Issue #477, opened 2020-06-30
RFC - Using an existing CloudFront distribution


danielcondemarin:
### Problem

Currently you can’t reuse an existing CloudFront distribution when deploying your Next.js application with `serverless-next.js`
This is a pretty common use case for many users that need a single CloudFront distribution as a gateway to different applications (e.g. multiple Next.js frontends), proxies etc.

### Proposal

Add a new input to `serverless-next.js` to pass an existing distribution id:

```yml
myNextApp:
  component: "@sls-next/serverless-component@{version}"
  inputs:
     cloudfront:
       distributionId: EDFDVBD6EXAMPLE      
```

The given CloudFront distribution will be updated as outlined below:

The distribution’s default cache behaviour will have a Lambda@Edge function attached that will allow for server side rendering Next.js pages. The function will be invoked on origin-request events. See the AWS Docs for details on possible triggers.

Four cache behaviours will be added to the distribution:
- `/api/*` - This is for serving Next.js API requests. If the app doesn’t have any API pages then the cache behaviour won’t be added.
- `/_next/*` - This is necessary for Next.js client side assets.
- `/static/*` - This is necessary for user defined assets under the static/ folder.
- `_next/data/{BUILD_ID}/*` - This will be used to support Next.js static site generation improvements

The cache behaviors above are the same ones provisioned when a brand new distribution is created.
It’s important to note that these are reserved by `serverless-next.js and the existing cloudfront distribution should not have them already defined to avoid clashes.

#### Defining a different root path

If you need to host multiple Next.js apps under the same distribution then an optional root path can be provided:

```yml
myNextApp:
  component: "@sls-next/serverless-component@1.2.3"
  inputs:
    cloudfront:
       distributionId: EDFDVBD6EXAMPLE     
    rootPath: "/blog"
```

In this case `serverless-next.js` will attach the SSR Lambda@Edge function to a `/blog/*` cache behaviour rather than modifying the Default cache behaviour.

With this approach you could have multiple apps hosted under the same CloudFront distribution:

```yml
leadsApp:
  component: "@sls-next/serverless-component@1.2.3"
  inputs:
     cloudfront:
       distributionId: EDFDVBD6EXAMPLE     
     rootPath: "/leads"
```

```yml
customersApp:
  component: "@sls-next/serverless-component@1.2.3"
  inputs:
     cloudfront:
       distributionId: EDFDVBD6EXAMPLE     
     rootPath: "/customers"
```

```yml
[...]

PR #658, opened 2020-10-06
fix(serverless-component): when adding primary domain name to CloudFront distribution aliases, don't replace other aliases

dpowell:
I missed this in ea180e67515ee5dceed1c3e19731291615a7cbbb, where the domain alias being added to the distribution would overwrite aliases that were added when we create the distribution. Append the domain alias(es) to the list instead of creating a new Aliases.Items to preserve those.

There don't seem to be any tests for this component, so I manually verified running the built component locally. If I overlooked a test suite that should be updated, let me know.

### DIFF
--- a/packages/serverless-components/domain/utils.js
+++ b/packages/serverless-components/domain/utils.js
@@ -393,21 +393,19 @@ const addDomainToCloudfrontDistribution = async (
   params.Id = subdomain.distributionId;
 
   // 5. then make our changes
-  params.DistributionConfig.Aliases = {
-    Quantity: 1,
-    Items: [subdomain.domain]
-  };
-
+  params.DistributionConfig.Aliases.Items.push(subdomain.domain);
+  params.DistributionConfig.Aliases.Quantity += 1;
   if (subdomain.domain.startsWith("www.")) {
     if (domainType === "apex") {
-      params.DistributionConfig.Aliases.Items = [
-        `${subdomain.domain.replace("www.", "")}`
-      ];
+      params.DistributionConfig.Aliases.Items[-1] = `${subdomain.domain.replace(
+        "www.",
+        ""
+      )}`;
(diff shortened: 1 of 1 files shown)


dphang (approved):

dphang:
@dpowell actually now that I think about it I think this may have conflicts by appending to existing aliases, if managing using `domain` inputs, we should let it manage and overwrite all aliases. Will it not keep appending to existing aliases when distribution needs to be updated. So you can end up with multiple entries e.g `["example.com", "example.com"]?

Otherwise if we don't want to use `domain` inputs, then just use the `aliases` input to manage the domains (and set certificate manually).

dphang:
Planning to revert this change in next couple days to original behavior which should be more correct, @dpowell please do let me know if you have any issues or concerns.

dpowell:
The push for the domain’s alias could be conditional on !contains, perhaps. 

It wasn’t clear to me that domain was optional, so I’ve been using domain for my “primary” (even though not using Route53 or ACM for it) and aliases for additional. Perhaps I just read too much into that and we could clarify in a docs setting about aliases to use only one or the other.

dphang:
@dpowell created this to revert to original behavior: https://github.com/serverless-nextjs/serverless-next.js/pull/731. Will do some more manual testing myself. If you can, please pull and build this branch and see if it works well for you still. Thanks!

danielcondemarin:
@dphang Personally I think we should aim to "update" distribution settings rather than "overriding" whenever possible. This is particularly beneficial to those reusing the same distribution for next.js deployments and other APIs that may require different CNAMES etc. Hence why I raised https://github.com/serverless-nextjs/serverless-next.js/issues/477 a while back. 

You're right though, it adds complexity and makes it more difficult to maintain. I think for now we should go with what you suggested and review it once we redo how deployments work in future.

PR #659, opened 2020-10-08
feat(lambda-at-edge): add minifyHandlers input to minify handler code using Terser

dphang:
By default handlers are not minified, but with `minifyHandlers` we can use the minified versions of handlers (pre-built into the package, so it won't add any time to the deployment), which reduces handler from 500 kB to 200 kB. It may also improve performance slightly.

## Tests

Added tests, also manually tested on e2e test app.


packages/libs/lambda-at-edge/tests/build/build.test.ts
```diff
@@ -20,221 +24,280 @@ describe("Builder Tests", () => {
   const fixturePath = join(__dirname, "./simple-app-fixture");
   const outputDir = join(fixturePath, ".test_sls_next_output");
 
-  beforeEach(async () => {
```
dphang, line 23: The original tests are not changed, not sure why git diff is this way

packages/libs/lambda-at-edge/tests/build/build.test.ts
```diff
@@ -20,221 +24,280 @@ describe("Builder Tests", () => {
...
+        ]);
+        expect(customerPages).toEqual(["[...catchAll].js", "[post].js"]);
       });
-      expect(dynamic).toEqual({
-        "/api/customers/:id": {
-          file: "pages/api/customers/[id].js",
-          regex: expect.any(String)
-        }
+
+      it("default handler is not minified", async () => {
```
dphang, line 232: New test

packages/libs/lambda-at-edge/tests/build/build.test.ts
```diff
@@ -20,221 +24,280 @@ describe("Builder Tests", () => {
...
-      // Note: JS files used only for prerendering at build time (contact.js, index.js) are included since there are API routes
-      expect(pages).toEqual([
-        "_error.js",
-        "blog.js",
-        "contact.js",
-        "customers",
-        "index.js"
-      ]);
-      expect(customerPages).toEqual(["[...catchAll].js", "[post].js"]);
+      it("API handler is not minified", async () => {
```
dphang, line 261: New test

packages/libs/lambda-at-edge/tests/build/build.test.ts
```diff
@@ -20,221 +24,280 @@ describe("Builder Tests", () => {
...
+
+        expect(countLines(apiHandler.toString())).toBeGreaterThan(100); // Arbitrary choice
+      });
     });
   });
 
-  describe("API Handler Artefact Files", () => {
-    it("copies build files", async () => {
-      expect.assertions(2);
+  describe("Minified handlers build", () => {
```
dphang, line 271: New test

PR #669, opened 2020-10-13
chore: separate workflow for e2e tests

dphang:


dphang:
Need to check in workflow first, will fix in case of any issues

PR #686, opened 2020-10-16
Fix for AWS accounts with more than 100 domains

rlehrke:
Pass in the domain name as a DNSName parameter into AWS's listHostedZonesByName method. This is particularly helpful for AWS accounts that have a large number of domains, since listHostedZonesByName only returns 100 records at a time. Instead of paging through all records, we can filter the initial search.

See: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Route53.html#listHostedZonesByName-property for documentation on this method

Fix typo "wwww" to "www".


packages/serverless-components/domain/utils.js
```diff
@@ -81,7 +81,11 @@ const getOutdatedDomains = (inputs, state) => {
  * - These don't need to be created and SHOULD NOT be modified.
  */
 const getDomainHostedZoneId = async (route53, domain, privateZone) => {
-  const hostedZonesRes = await route53.listHostedZonesByName().promise();
+  const params = {
+    DNSName: domain
```
dphang, line 85: Does this also work if domain has `www` in it or other subdomain?

rlehrke: @dphang - Hey Daniel!  Great question.

Route 53 shouldn't ever return a subdomain from this list, since the listHostedZonesByName should only return the full domain that was used when creating the zone.

image

Here is how Route 53 describes the "domain name" field when creating a zone.

dphang (approved):

dphang:
Thanks, please update the branch and I can run the latest e2e tests and merge it

dphang:
Ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/314496337

PR #688, opened 2020-10-16
Add a binary to allow use without copying example

tomyan:
This makes running builds simpler for non-serverless projects that use this library.


dphang:
Thanks, ran e2e tests workflow here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/311606442

packages/libs/lambda-at-edge/src/command.ts
```diff
@@ -0,0 +1,30 @@
+import { default as Builder } from "./build";
+import { join } from "path";
+
+async function main (args: string[]) {
+  if (args.length > 1) {
+    console.error("Usage: build-lambda-at-edge [ NEXT_APP_DIR ]")
```
dphang, line 6: Please run prettier to reformat this whole file, they have inconsistent semicolons (it should be part of git hook)

tomyan: Hi @dphang 

Thanks for the feedback. I've pushed a commit with the requested formatting using prettier. I'm not sure how to meaningfully unit test this functionality - perhaps mocking out and injecting the Builder, but it seems this would just be a test of the implementation rather than the behaviour. Let me know if you have a suggestion and I'll look at implementing it.

Thanks

Tom

dphang: Yeah, I guess just ensure this script reasonably works as expected, mocking should be fine but if you don't think it's too useful, I'm fine as well since this is not part of the core lambda-at-edge source code.

tomyan: I started to look at this but it wasn't immediately obvious how to run the unit tests for the lambda-at-edge module. If you're happy to go with it as is then that would be fine by me, or I'm happy to write a test if you can give me some guidance on how to get the tests running. Thanks

dphang: We can merge it for now, you can check the CONTRIBUTING.md for how to run tests, we use `yarn test lambda-at-edge` which will run the tests there.

Thanks.

Issue #642, opened 2020-10-01
Deploying to existing infrastructure


finferflu:
Hi, 

I can't seem to find anything that relates to this in the documentation, however, I'd like to deploy my application to existing infrastructure. Is there any way to package the application so that it's ready for deployment, but without creating all the resources via Serverless? 

Thank you.

dphang:
There's no input yet to specify to build the Lambda bundle but not deploy to AWS.

Perhaps a workaround is to set invalid AWS keys, I think it still may output the handlers to the .serverless-nextjs generated directory. However I think S3 assets are uploaded directly, there's no build artifact in that folder, so you would have to figure out that based on .next directory contents.

Feel free to add a new input to do this and I'll be happy to review it.

finferflu:
Thanks for this @dphang, I'll go back to the code and see what we can do.

dphang:
I haven't planned to make a change yet, but it does seem straightforward enough to disable the actual deployment step (separate build from the deployment) if you just want the build outputs (contents of the `serverless_next.js` folder).

We'd have to add the S3 assets there too that would have been uploaded in that case. Then you can just use the content of this directory and you'd have to deploy it yourself to Lambda/CloudFront/S3.

finferflu:
Thanks, @dphang, decoupling the asset creation from the actual deployment sounds like a great strategy, and it would surely make this component a lot more viable for people who want to deploy using different tools or providers.

I'll see what I can make of these suggestions and I'll get back to you should any more questions arise. Meanwhile, I'm glad this was labelled as "enhancement", and I hope this will become a feature soon!

dphang:
Yea, makes sense.

Short term we can make those small refactor to better decouple build and deploy step. So you can deploy the build outputs to AWS yourself.

Long term I think it makes sense to further decouple the cloud providers from this process (right now AWS is heavily coupled into the core logic). So we could have something like this:

1. Common core logic for Next.js serverless which is generic and platform/provider agnostic.

Then, we could adopt a plugin architecture to build and deploy for specific platforms:

2. Build and bundle the handlers using the specific plugin (AWS Lambda@Edge, and other providers in the future). With Rollup it should be easier to do so (for example, including only S3 dependency for AWS deployments)
3. Deploy the handler using the corresponding deployment plugin (e.g AWS Lambda@Edge right now).

It will make it easier to support other platforms (Lambda, Cloudflare, GCP, Azure).

dphang:
[...]

PR #689, opened 2020-10-17
refactor(lambda-at-edge, s3-static-assets): decouple s3 assets packaging from the upload…

dphang:
## Motivation

This will help for this issue: https://github.com/serverless-nextjs/serverless-next.js/issues/642. The idea is to better decouple the build and deploy step for the Next.js assets by separating into two steps: build (copying files to an output directory) and S3 upload (from that output directory).

* A new `.serverless_nextjs/assets` folder is generated as part of the build, similar to `default-lambda` and `api-lambda`. This is the same structure as the files uploaded to S3. The idea is to give more flexibility to users to upload their own assets in the future by decoupling the build and deploy steps.
* A new S3 upload function is written to upload from `.serverless_nextjs/assets` folder, which will set Cache Control headers as well.
* I've kept but deprecated the old way (which directly parses `.next` folder and uploads to S3 depending on directory/file type). This is undocumented, but I've also added a temporary killswitch input `uploadStaticAssetsFromBuild` which can be set to `false` to disable this, in case there are issues with this new way.
* Build times should remain similar based on my tests, although there is now a copy + upload, it is all parallelized.

## Tests

Added unit tests. For e2e, manually tested, but will run e2e tests in the CI workflow as well.


danielcondemarin (approved):
LGTM 🚀

PR #692, opened 2020-10-18
Feature flag lambda at edge compression

danielcondemarin:
**Motivation**

We've been manually doing Gzip compression in lambda@edge so far due to CloudFront not compressing lambda@edge responses. However, seems like that [has been fixed](https://twitter.com/cristiangraz/status/1306277707353853952?s=20) after discussing with Cristian from the CloudFront team and doing some testing of my own. 

This should allow us to use Gzip and Brotli compression without extra code in the Lambda@Edge handler. Furthermore the CloudFront distributions created by serverless-next.js already [enable compression](https://github.com/serverless-nextjs/serverless-next.js/blob/fb8e61dc50b11c0e5966548a8c84b58e495ea748/packages/serverless-components/aws-cloudfront/lib/getCacheBehavior.js#L9) by default. 

From my initial findings Gzip works fine, but Brotli isn't despite [being officially supported now](https://aws.amazon.com/about-aws/whats-new/2020/09/cloudfront-brotli-compression/#:~:text=You%20can%20now%20use%20Amazon,better%20compression%20ratio%20than%20Gzip.)

Seems distributions will need to set `EnableAcceptEncodingBrotli`. 

Ref docs: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html

### DIFF
--- a/packages/compat-layers/lambda-at-edge-compat/next-aws-cloudfront.d.ts
+++ b/packages/compat-layers/lambda-at-edge-compat/next-aws-cloudfront.d.ts
@@ -1,9 +1,16 @@
 import { CloudFrontResultResponse, CloudFrontRequest } from "aws-lambda";
 import { IncomingMessage, ServerResponse } from "http";
 
-declare function lambdaAtEdgeCompat(event: {
-  request: CloudFrontRequest;
-}): {
+type CompatOptions = {
+  enableHTTPCompression: boolean;
+};
+
+declare function lambdaAtEdgeCompat(
+  event: {
+    request: CloudFrontRequest;
+  },
+  options: CompatOptions
+): {
   responsePromise: Promise<CloudFrontResultResponse>;
   req: IncomingMessage;
   res: ServerResponse;
--- a/packages/compat-layers/lambda-at-edge-compat/next-aws-cloudfront.js
+++ b/packages/compat-layers/lambda-at-edge-compat/next-aws-cloudfront.js
@@ -144,7 +144,7 @@ const isGzipSupported = (headers) => {
   return gz;
 };
 
-const handler = (event) => {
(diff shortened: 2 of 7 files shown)


packages/compat-layers/lambda-at-edge-compat/README.md
```diff
@@ -20,3 +20,13 @@ module.exports.render = async (event, context) => {
...
 };
 ```
+
+## Options
+
+### Gzip compression
+
+```js
+const { req, res, responsePromise } = cloudFrontCompat(event.Records[0].cf, {
+  enableHTTPCompresssion: true // false by default
```
dphang, line 30: nit: extra s here

packages/e2e-tests/next-app/cypress/integration/pages.test.ts
```diff
@@ -46,6 +49,9 @@ describe("Pages Tests", () => {
         (method) => {
           it(`allows HTTP method for path ${path}: ${method}`, () => {
             cy.request({ url: path, method: method }).then((response) => {
+              if (method !== "HEAD") {
```
dphang, line 52: Maybe add for public files etc. but can be done later.

packages/e2e-tests/test-utils/cypress/custom-commands.ts
```diff
@@ -140,3 +143,10 @@ Cypress.Commands.add(
     }
   }
 );
+
+Cypress.Commands.add(
+  "verifyResponseIsCompressed",
+  (response: Cypress.Response) => {
+    expect(response.headers["content-encoding"]).to.equal("gzip");
```
dphang, line 150: nice, thanks for adding a test in the e2e tests as well.

dphang (approved):

Issue #694, opened 2020-10-19
"comment:"  input for Cloudfront is not working


LuisFros:
### Describe the bug
When adding custom inputs to the cloudfront distribution in `serverless.yml`, the option `"comment:"` fails with a validation error.
Serverless.yml
```
myNextApplication:
  component: '@sls-next/serverless-component@1.17.1-alpha.3'
  inputs:
    bucketName: ${env.BUCKET_NAME}
    cloudfront:
      comment: "Cloudfront comment"
      defaults:
        forward:
          headers: [Host]
```

### Actual behavior

```
  Error: CloudFront input failed validation. Could not find next.js pages for "comment"
    at NextjsComponent.validatePathPatterns (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.1-alpha.3/node_modules/@sls-next/serverless-component/dist/component.js:95:19)
    at NextjsComponent.<anonymous> (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.1-alpha.3/node_modules/@sls-next/serverless-component/dist/component.js:305:18)
    at Generator.next (<anonymous>)
    at fulfilled (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.1-alpha.3/node_modules/@sls-next/serverless-component/dist/component.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
  88s › myNextApplication › Error: CloudFront input failed validation. Could not find next.js pages for "comment"
```

### Expected behavior
For the deployment to be successful and `comment` be included in the distribution
### Steps to reproduce
- Add `comment` option inside `serverless.yml`, as  `cloudfront` options.
- If i remove the `comment` option, there is no issue to deploy the component.
### Screenshots/Code/Logs

```
  Error: CloudFront input failed validation. Could not find next.js pages for "comment"
    at NextjsComponent.validatePathPatterns (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.1-alpha.3/node_modules/@sls-next/serverless-component/dist/component.js:95:19)
    at NextjsComponent.<anonymous> (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.1-alpha.3/node_modules/@sls-next/serverless-component/dist/component.js:305:18)
    at Generator.next (<anonymous>)
[...]

PR #695, opened 2020-10-20
fix(aws-cloudfront, nextjs-component): fix inability to set cloudfront comment input

dphang:
Fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/694


packages/serverless-components/aws-cloudfront/lib/index.js
```diff
@@ -26,7 +26,7 @@ const createCloudFrontDistribution = async (cf, s3, inputs) => {
   const params = {
     DistributionConfig: {
       CallerReference: String(Date.now()),
-      Comment: inputs.comment,
+      Comment: inputs.comment ?? "",
```
dphang, line 29: Per https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_DistributionConfig.html comment is required and needs to be empty by default (if no comment is needed). Similar to update code to set to "" by default so that it can be deleted

Issue #703, opened 2020-10-21
Aliases are overwritten when not declared in `serverless.yml`


LuisFros:
Current version: `'@sls-next/serverless-component@1.18.0-alpha.13`
**Is your feature request related to a problem? Please describe.**
When updating a distribution, the `aliases` or CNAMEs get overwritten. This is a problem in the latest version starting from this PR: https://github.com/serverless-nextjs/serverless-next.js/pull/654

**Describe the solution you'd like**
I would like the aliases field to not be overwritten when existing cloudfront distribution already has values that are not empty and `aliases` field is not declared in `serverless.yml`

**Describe alternatives you've considered**
Storing aliases as env variable during CI/CD

**Additional context**
Here could be checked `params.DistributionConfig.Aliases !== null`
https://github.com/serverless-nextjs/serverless-next.js/blob/ef2d596d5a39af2009f48b39742c2aa269f103c0/packages/serverless-components/aws-cloudfront/lib/index.js#L121

dphang:
Thanks for pointing out. Looks like aliases when undefined will get overridden to default value of [] which will be destructive to existing aliases. We should keep inputs non-destructive (similar to terraform) as much as possible, i.e if not specified it should not try to update that configuration. Will fix it tonight.

PR #704, opened 2020-10-21
fix(aws-cloudfront): do not override aliases when not specified in serverless.yml inputs

dphang:
This fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/703

Essentially, we should not be destructive when inputs are not specified for CloudFront. If aliases is set to `[]` then it will override with empty aliases, but if undefined it should not override existing aliases.

@dpowell and @LuisFros can you please help pull this change, build locally and test it since you have been using this.


LuisFros:
@dphang I just ran the deploy locally and it works!

packages/serverless-components/aws-cloudfront/__tests__/general-options.test.js
```diff
@@ -130,6 +130,25 @@ describe("General options propagation", () => {
...
   });
 
+  it("update distribution with undefined aliases does not override existing aliases", async () => {
+    // Create distribution
+    await component.default({ enabled: true, origins });
+
+    // Update distribution
+    await component.default({
+      enabled: false,
+      origins
```
LuisFros, line 140: Complementing your test, you could do the following. 
```suggestion
      aliases: ["foo.example.com"],
      origins
```

dphang:
@LuisFros thanks for confirming

dpowell:
Works well here, too. Thanks!

PR #717, opened 2020-10-28
tests(e2e-tests): upgrade tests for Next.js 10.0.0, retain old tests for 9.5.5

dphang:
* Next.js 10 came out today: https://nextjs.org/blog/next-10. This upgrades tests from 9.5.5 to 10.0.0 and retains existing tests for backwards compatibility purposes. I think the e2e tests should be fairly complete (except for preview mode, which I can add in a bit), so I'm fine to upgrade them as-is.
* Many changes but most are just copying existing tests and adding a `-9-5` suffix.


dphang:
Will redo this in another PR

Issue #722, opened 2020-10-29
InconsistentQuantities error when deploying with @sls-next/serverless-component@1.18.0-alpha.5^


eddeee888:
### Describe the bug
I'm currently using `@sls-next/serverless-component@1.17.0` with `apex` domain settings ( full serverless yml below )

I'm trying to use the custom header feature as mentioned in https://github.com/serverless-nextjs/serverless-next.js/issues/587#issuecomment-706606394. so I tried to use `@sls-next/serverless-component@1.18.0-alpha.5` ( and also `@sls-next/serverless-component@1.18.0-alpha.22`)

However, when I try to deploy it, I get `InconsistentQuantities` error in CI. Probably happens if I run it locally too.

### Actual behavior
I get the following error as it's trying to deploy:
`InconsistentQuantities: The specified quantity of Aliases (1) does not match the actual quantity supplied (2)` 

### Expected behavior
It should deploy without error

### Steps to reproduce
Here's a version of my serverless yml without all the credentials:

```
org: MY_ORG
name: MY_APP

app-one:
  component: "@sls-next/serverless-component@1.18.0-alpha.5"
  inputs:
    bucketName: my-bucket-name
    domain: ${env.APP_DOMAIN}
    domainType: apex
    description: description here
    memory: 512
    timeout:
      defaultLambda: 20
    verbose: true
    build:
      cmd: yarn
      env:
        NODE_ENV: production
```

### Screenshots/Code/Logs
Here's the log when it fails ( v1.18.0-alpha.5 and v1.18.0-alpha.22 ): 
```
DEBUG ─ Analyzing the template's components dependencies.
DEBUG ─ Creating the template's components graph.
DEBUG ─ Syncing template state.
DEBUG ─ Executing the template's components graph.
DEBUG ─ Resolving the template's static variables.
DEBUG ─ Collecting components from the template.
DEBUG ─ Downloading any NPM components found in the template.
...
DEBUG ─ Deploying bucket bucket-name in region us-east-1.
DEBUG ─ Checking if bucket bucket-name exists.
DEBUG ─ Setting acceleration to "true" for bucket bucket-name.
DEBUG ─ Bucket bucket-name was successfully deployed to the us-east-1 region.
...
DEBUG ─ Starting deployment of lambda aaaa-bbbb to the us-east-1 region.
DEBUG ─ Syncing role aaaa-dddd in region us-east-1.
DEBUG ─ Saved state for role aaaa-dddd.
[...]

PR #723, opened 2020-10-29
fix(domain, nextjs-component): set aliases correctly with domainType of apex and domain starting with www

dphang:
* Also adds better logging during debug mode
* Should hopefully fix: https://github.com/serverless-nextjs/serverless-next.js/issues/722


dphang:
Merging since this is a small bug fix.

PR #731, opened 2020-10-30
fix(aws-cloudfront, nextjs-component): fix aliases updating when domain is used

dphang:
* When domain is used, it should completely manage all aliases attached to CloudFront distribution instead of trying to push onto existing aliases set. Otherwise, there could be strange behavior when it's manually updated and/or domains are changed.
* `aliases` input in `cloudfront` can be used if additional aliases apart from domain name are needed (and domain likely should be managed externally instead, as this is a more complex use case).

### DIFF
--- a/packages/serverless-components/domain/utils.js
+++ b/packages/serverless-components/domain/utils.js
@@ -409,7 +409,6 @@ const addDomainToCloudfrontDistribution = async (
 
   // 5. then make our changes
   params.DistributionConfig.Aliases = {
-    Quantity: 1,
     Items: [subdomain.domain]
   };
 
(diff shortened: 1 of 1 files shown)


packages/serverless-components/domain/utils.js
```diff
@@ -408,19 +408,24 @@ const addDomainToCloudfrontDistribution = async (
   params.Id = subdomain.distributionId;
 
   // 5. then make our changes
-  params.DistributionConfig.Aliases.Items.push(subdomain.domain);
+  params.DistributionConfig.Aliases = {
+    Quantity: 1,
```
dpowell, line 412: No real reason to set this to 1, it'll get set correctly on line 429.

dpowell (approved):

packages/serverless-components/domain/utils.js
dphang: Sure, I can remove it, thanks.

PR #758, opened 2020-11-02
fix(lambda-at-edge): for serverless-trace target, exclude files that …

dphang:
that are typescript sources.

Assuming that this should fix https://github.com/serverless-nextjs/serverless-next.js/issues/461? It works to fix TypeScript projects at least.

I can add some e2e tests for this too later.


dphang:
@danielcondemarin not sure this is the right fix but for my project this excludes all ts/tsx files that were trying to be copied with same src and destination (hence failing). However it fails the test, not sure how we can test it?

Should we just exclude .ts/.tsx files instead?

danielcondemarin:
Doesn't look right to me @dphang . That would be excluding traces that are within the app itself and not in `node_modules`. I suspect that will break for some apps. 

What are the ts / tsx files you're seeing? Dependencies of which page? I wouldn't expect to see a ts / tsx trace after building the app.

dphang:
@danielcondemarin if you see that issue I had mentioned somehow the serverless trace was trying to copy TSX pages from and to the same directory and it was failing at the file copy command due to source and destination being the same.

It worked for my typescript project but that was a quick hack at the time (a few months ago) - I also didn't understand much about how this component works.

Let me go and look more closely to make a proper fix.

dphang:
@danielcondemarin I think the @zeit/node-file-trace is somehow returning the app's pages TS/TSX files which shouldn't be copied. Not sure if it's a bug on this package, but I think we can safely exclude these files. Also looking into @vercel/nft since it mentions that @zeit/node-file-trace is deprecated, maybe it's fixed in the new version.

danielcondemarin (approved):

packages/libs/lambda-at-edge/src/build.ts
```diff
@@ -131,6 +131,13 @@ class Builder {
       .filter((file) => {
         // exclude "initial" files from lambda artefact. These are just the pages themselves
         // which are copied over separately
+
+        // For TypeScript apps, somehow nodeFileTrace will generate filelist with TS or TSX files, we need to exclude these files to be copied
+        // as it ends up copying from same source to destination.
```
dphang, line 136: Also we could prevent FS from copying if src == dest but I think this should work for now.

danielcondemarin:
Would you mind upgrading to https://github.com/vercel/nft? Hopefully shouldn't be breaking.

dphang:
> Would you mind upgrading to https://github.com/vercel/nft? Hopefully shouldn't be breaking.

Yea I tried it but it seems like it broke many unit tests, need more time to investigate it

PR #773, opened 2020-11-06
feat(nextjs-component, lambda-at-edge): support single username/password basic authen…

dphang:
…tication

* It will help to limit access by allowing a single username/password authentication mechanism, useful to protect dev environments, etc.


danielcondemarin:
Maybe there is a more secure way of handling the username & password, inlining in the manifest doesn't feel right. For example we could at deploy time, update [SecretsManager](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html) with the given username & pwd and at runtime read from it, instead of including the sensitive in the lambda artefact.

dphang:
> Maybe there is a more secure way of handling the username & password, inlining in the manifest doesn't feel right. For example we could at deploy time, update [SecretsManager](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html) with the given username & pwd and at runtime read from it, instead of including the sensitive info in the lambda artefact. 
> 
> 
> 
> 

@danielcondemarin Ya I agree, this was just a start to quickly protect an environment e.g useful for development envs. I don't think we should be using basic auth for production. Secret manager is not free ($0.40 per month and $0.05 per 10k API calls) but we can add it later if needed.

danielcondemarin:
> > Maybe there is a more secure way of handling the username & password, inlining in the manifest doesn't feel right. For example we could at deploy time, update [SecretsManager](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html) with the given username & pwd and at runtime read from it, instead of including the sensitive info in the lambda artefact.
> 
> @danielcondemarin Ya I agree, this was just a start to quickly protect an environment e.g useful for development envs. I don't think we should be using basic auth for production. Secret manager is not free ($0.40 per month and $0.05 per 10k API calls) but we can add it later if needed.

I'm concerned if we release it as public api, people will use in production and ignore the fact is not very secure. We'd need to make that quite clear in the README.

Also, I think we should start thinking of a way to pull out of the main handler things like this into their own "middlewares". Give users the ability to extend the handler by plugging in a middleware would be awesome. Would work for basic http auth, sentry, etc. [Middy](https://github.com/middyjs/middy) would be an interesting option.

danielcondemarin (approved):
Anyways, I think it's good to merge, but would be good to update the README with the not meant for prod note 👍

dphang:
> > > Maybe there is a more secure way of handling the username & password, inlining in the manifest doesn't feel right. For example we could at deploy time, update [SecretsManager](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html) with the given username & pwd and at runtime read from it, instead of including the sensitive info in the lambda artefact.
> 
> > 
> 
> > @danielcondemarin Ya I agree, this was just a start to quickly protect an environment e.g useful for development envs. I don't think we should be using basic auth for production. Secret manager is not free ($0.40 per month and $0.05 per 10k API calls) but we can add it later if needed.
> 
> 
> 
> I'm concerned if we release it as public api, people will use in production and ignore the fact is not very secure. We'd need to make that quite clear in the README.
> 
> 
> 
> Also, I think we should start thinking of a way to pull out of the main handler things like this into their own "middlewares". Give users the ability to extend the handler by plugging in a middleware would be awesome. Would work for basic http auth, sentry, etc. [Middy](https://github.com/middyjs/middy) would be an interesting option. 

Ya, I'd not make it specific to lambda as it would hamper support for other platforms, but it can be done using Rollup.js plugins. This approach could be taken to dynamically add or remove features that are not needed or used (e.g auth, rewrites, etc).

Maybe something like request middleware + (core serverless code) + response middleware.

Will keep this branch in case someone wants to use it.

dphang:
@danielcondemarin sorry, I did not see your last comment, but I updated with some warnings in bold in the docs. Are you still fine if we add it? I think this is useful for me/others who may want to protect a dev environment at least with some basic auth.

As mentioned, we should rework this and avoid bloating the handler by probably creating it as a first-party middleware, and then separating all into the modular pieces (request middleware + core + response middleware) and use Rollup.js bundling to optimize unused code.

I'll work on that next + separating Lambda@Edge deployment logic into lambda-at-edge package.

danielcondemarin:
> @danielcondemarin sorry, I did not see your last comment, but I updated with some warnings in bold in the docs. Are you still fine if we add it? I think this is useful for me/others who may want to protect a dev environment at least with some basic auth.
> 
> As mentioned, we should rework this and avoid bloating the handler by probably creating it as a first-party middleware, and then separating all into the modular pieces (request middleware + core + response middleware) and use Rollup.js bundling to optimize unused code and bundle it with the appropriate platform-specific code (e.g Lambda handler wrapping all this).
> 
> I'll work on that next + separating Lambda@Edge deployment logic into lambda-at-edge package.

Yeah, I'm fine with it if you are. I definitely think is useful, was just concerned about the security implications but if is clear in the docs that is not for production I guess that's fine. I'll keep it in the back of my mind in case Lambda@Edge adds support for env. variables which would be a more secure sol. Also if we pull it out from the default handler later even better 👍

danielcondemarin (approved):

Issue #776, opened 2020-11-06
Preview Mode 403 from Cloudfront (exit preview mode gracefully)


patricktyndall:
### Describe the bug
When a NextJs site is built, previous NextJs preview cookies expire. Using the Next server, this just sends the user to the non-preview-mode version. However using serverless-next.js, this returns a 403 from CloudFront, which is bad UX for creators.

Importantly, this affects the entire site, including pages which are not SSG ie not-preview-mode enabled. So if someone needs to go to the site for another reason, it gives the 403.

### Actual behavior
- 403 is returned, browser gives error and does not render site

### Expected behavior
- user can visit non-preview-mode version of site
- (I'm not sure what the response code should be -- haven't checked what Next does)

### Steps to reproduce
- Deploy a site and enter preview mode
- Deploy it again and refresh the page (without revisiting the preview API route URL)

### Versions
- OS/Environment: Mac
- @sls-next/serverless-component@1.18.1-alpha.1
- next: 9.5.5

dphang:
Looks like affected code is here: https://github.com/serverless-nextjs/serverless-next.js/blob/625f6bed672b8a6dfa8668a26296901be17049ef/packages/libs/lambda-at-edge/src/default-handler.ts#L354-L367

I don't use preview mode myself much so not sure all of its nuances. @thchia had implemented this, so perhaps he could help on this? Feel free to raise a PR as well.

thchia:
@patricktyndall thanks for raising this. Haven't reproduced it but it sounds about right. It's a session cookie so if you use the same browser tab it's still in preview mode.

@dphang you're right about the source of this. 

Happy to look at this but not sure when exactly I'll have the fix out so whoever else feels like taking a stab please do.

simonireilly:
@thchia and @dphang I took a stab at this as a bug fix. It seems like it would be an enhancement to return a response that clears these cookies from the browser if they are invalid.

I was also considering modularity it could be nicer to handle preview in its own module, perhaps with a lib file `lib/isPreviewRequest.ts` with a `boolean` return object. This seems good to me as it tidies away some small piece of encapsulated logic. WDYT?

thchia:
@simonireilly I support moving the preview out to a different module. We only gained the ability to have separate files _relatively_ recently, so in most cases the fact that everything is in one file is not deliberate.

dphang:
[...]

PR #780, opened 2020-11-07
Validate Existance of Cookies for Preview Mode - Default preview mode False

simonireilly:
Related issue: #776 

Examining how NextJS handle this, they return false in a catch after signing failure.

https://github.com/vercel/next.js/blob/a5cb28d907ad34865085d38d008130aaa4667427/packages/next/next-server/server/api-utils.ts#L350-L359

### This Change

Instead of returning 403 for requests which contain both cookies, but which are invalid, a normal response object should be returned.

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -347,22 +347,21 @@ const handleOriginRequest = async ({
   const hasFallback = hasFallbackForUri(uri, prerenderManifest, manifest);
   const { now, log } = perfLogger(manifest.logLambdaExecutionTimes);
   const previewCookies = getPreviewCookies(request);
-  const isPreviewRequest =
+  let isPreviewRequest = false;
+  const containsPreviewCookies =
     previewCookies[NEXT_PREVIEW_DATA_COOKIE] &&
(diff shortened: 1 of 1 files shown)


packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -347,22 +347,21 @@ const handleOriginRequest = async ({
   const hasFallback = hasFallbackForUri(uri, prerenderManifest, manifest);
   const { now, log } = perfLogger(manifest.logLambdaExecutionTimes);
   const previewCookies = getPreviewCookies(request);
-  const isPreviewRequest =
```
simonireilly, line 350: Previously the presence of these headers was coupled with the validity of the encrypted cookies in the headers.

This should not be the case as stale cookies may exist in the browser when there is no explicit call to [clear the cookies manually](https://nextjs.org/docs/advanced-features/preview-mode#clear-the-preview-mode-cookies)

dphang:
Thanks, ran e2e tests here (although it's not testing for preview mode yet): https://github.com/serverless-nextjs/serverless-next.js/actions/runs/351429150. I'll add some tests later for preview mode.

PR #779, opened 2020-11-07
Add baseDir and resolve build options to serverless-trace

ramosbugs:
This PR makes two changes to the `lambda-at-edge` builder to better
support Yarn v2 projects when using the `serverless-trace` target.

1. Adds a `baseDir` build option to specify the base directory to search
   for `node_modules`. Currently, the builder sets this to `process.cwd()`,
   but Yarn v2 often hoists dependencies across multiple workspaces so
   that they can be shared. Without this change, all dependencies from
   ancestor directories are omitted, leading to import errors at runtime.
2. Adds a `resolve` build option which allows projects to specify their
   own custom resolvers, such as when supporting Yarn v2's PnP mode.
   This leverages the resolver hook added in vercel/nft#153, which also
   required upgrading from `@zeit/node-file-trace` to the latest version
   of `@vercel/nft`. Note that even after this change, Yarn v2's PnP mode
   is still incompatible with the builder since the builder ends up
   collapsing all of the PnP dependencies into a single `node_modules`
   directory. This causes multiple versions of a package to clobber one
   another.

cc: @sergioramos (no relation 😆)

PR #787, opened 2020-11-11
E2E for useServerlessTraceTarget

danielcondemarin:
This is literally a copy of `e2e-tests/next-app` with one difference, it sets `useServerlessTraceTarget: true`.

I think we should aim to set serverless trace as the default build target in future, maybe for 1.19. Hopefully the last few fixes and https://github.com/serverless-nextjs/serverless-next.js/pull/779 have made it stable enough.


packages/e2e-tests/next-app-using-serverless-trace/serverless.yml
```diff
@@ -0,0 +1,6 @@
+next-app:
```
dphang, line 1: change this name to `next-app-using-serverless-trace

packages/e2e-tests/next-app-using-serverless-trace/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "next-app",
```
dphang, line 2: Change this to `next-app-using-serverless-trace`

packages/e2e-tests/next-app-using-serverless-trace/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "version": "1.0.0",
+  "description": "Basic test Next.js app for running serverless-next.js end-to-end tests",
+  "scripts": {
+    "dev": "next",
+    "build": "next build",
+    "start": "next start -p $PORT",
+    "serverless": "serverless",
+    "e2e": "cypress run",
+    "e2e:ci": "SSG_PAGE_PATH=/ssg-page SSR_PAGE_PATH=/ssr-page APP_NAME=next-app ts-node --project tsconfig.scripts.json ../test-utils/scripts/run-e2e-ci.ts"
```
dphang, line 11: Change `next-app` to `next-app-using-serverless-trace`. Needed so it doesn't conflict with the existing e2e tests

dphang:
@danielcondemarin LGTM besides that.

I'm not so sure about setting serverless trace target as the default build though, since although it reduces code size it can have cold start perf impact, since it's now having to load a bunch of smaller JS modules instead of a single file.

There are ways to optimize code size like minifying the pages JS files or manually removing page JS only used for pre-rendering (e.g we do it automatically if you do not use any API routes). I added `postBuildCommands` input which can help with this if you do use API routes.

packages/e2e-tests/next-app-using-serverless-trace/package.json
dphang: Also you need to modify the CI YML file to add this to the strategy matrix

dphang (approved):

danielcondemarin:
> I'm not so sure about setting serverless trace target as the default build though, since although it reduces code duplication, it can have cold start perf impact, since it's now having to load a bunch of smaller JS modules instead of a single file for a page. From my experience AWS Lambda code size has very little impact on cold start, it's more the I/O/requires that makes up the majority of the time.

That's a fair point. I'm mainly proposing the change because the Next team uses this as their build target for Vercel deployments. The other is that I suspect they will be deprecating the serverless target. 
Happy to keep as as an opt-in for now. I might do a bit more digging on this area and post any findings.

dphang:
> > I'm not so sure about setting serverless trace target as the default build though, since although it reduces code duplication, it can have cold start perf impact, since it's now having to load a bunch of smaller JS modules instead of a single file for a page. From my experience AWS Lambda code size has very little impact on cold start, it's more the I/O/requires that makes up the majority of the time.
> 
> That's a fair point. I'm mainly proposing the change because the Next team uses this as their build target for Vercel deployments. The other is that I suspect they will be deprecating the serverless target.
> Happy to keep serverless-trace as an opt-in for now. I might do a bit more digging on this area and post any findings.

Yep, I did try Vercel's deployment and found it slow for that same reason - a while back I had opened an issue but had no responses. I hope they do not remove the serverless target.

dphang:
Looks like e2e tests failed here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/358701235

packages/e2e-tests/next-app-using-serverless-trace/next.config.js
```diff
@@ -0,0 +1,25 @@
+module.exports = function (...args) {
```
dphang, line 1: Woops, overlooked this, I will fix for you @danielcondemarin

PR #803, opened 2020-11-16
feat(lambda-at-edge, nextjs-component, s3-static-assets): support ver…

dphang:
…sioned static pages (currently only next data is versioned) for proper fallback SSG generation, never cache fallback pages

* This will create versioned static pages directory for assets e.g `static-pages/build1`, `static-pages/build2`. It also copies the `BUILD_ID` file to `assets/BUILD_ID` or `assets/basepath/BUILD_ID` to represent the current latest version that was just uploaded to S3.
* When the `BUILD_ID` file is present in S3, it uses it and the AWS S3 SDK to clear all previous build versions of `_next/data/<build_id>` and `static-pages/<build_id>` except the version in `BUILD_ID`. It doesn't remove existing pages in the root `static-pages` directory, so it won't be destructive when transitioning from old to new Lambda versions. This happens before the S3 upload of the new build version. Then once the new S3 assets are uploaded, only the `new` + `previous` build versions are present. This is to help us save on S3 storage costs and ensure Lambda doesn't break on deployment transitions.
* This (along with cache control fixes) is needed to fix fallback SSG generation, as currently fallback pages are not versioned and put into `static-pages/pages.html`. This never gets deleted, so subsequent deployments may be getting an old generated page for an SSG fallback. By using versioned pages instead, we can ensure that it gets updated (and old ones get cleared after deployments).
* Versioning pages could also fix a general issue (not fallback specific) where when transitioning between Lambda deployments, a newer version's page is served (via S3) by an older Lambda, due to S3 uploading finishing before the newer Lambda could propagate, which could possibly cause issues (but probably rare?)
* Fallback pages are never cached (Cache-Control in both rendering + S3 upload is set to `no-store`) and now can also use the versioned pages properly.

## Tests

Updated unit and e2e tests. Added tests for new S3 removal method for both with and without basePath.


packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -570,7 +570,9 @@ const handleOriginResponse = async ({
         "cache-control": [
           {
             key: "Cache-Control",
-            value: "public, max-age=0, s-maxage=2678400, must-revalidate"
+            value: hasFallback.fallback
+              ? "no-store" // fallback should never be cached
```
dphang, line 574: possibly not the right cache-control header, maybe `max-age=0, must-revalidate` is better?

packages/libs/s3-static-assets/src/index.ts
```diff
@@ -334,4 +357,50 @@ const uploadStaticAssets = async (
...
+    credentials: options.credentials
+  });
+
+  // Get BUILD_ID file from S3 if it exists
+  let buildId = await s3.getFile({ key: "BUILD_ID" });
+
+  // If above exists, remove all versioned assets that are not BUILD_ID file (we don't remove unversioned pages static-pages as those are the previous way)
+  if (buildId) {
+    // Delete old _next/data versions except for buildId
+    await s3.deleteFilesByPattern({
```
dphang, line 387: Should be safe to do this in parallel.

danielcondemarin (approved):
LGTM 👍

PR #809, opened 2020-11-19
Add ability to persist cache behaviors not set in serverless

### LINKED ISSUE #805: Deploy wipes existing CloudFront behaviours
We're looking to manage Cloudfront outside of serverless as we have multiple origins to support,

We're noticing that if we create cache behaviours in the AWS console and deploy they get wiped on deploy.

Is there any way to persist behaviours created outside of serverless?

raffij:
Work to close #805 

Due to the way that CacheBehaviors are managed we needed to persist those for origins not controlled by serverless.

Tests have been written and tested locally in our next project, but didn't run the e2e tests.

### DIFF
--- a/packages/serverless-components/aws-cloudfront/test-utils.js
+++ b/packages/serverless-components/aws-cloudfront/test-utils.js
@@ -25,7 +25,9 @@ module.exports = {
       expect.objectContaining({
         DistributionConfig: expect.objectContaining({
           CacheBehaviors: expect.objectContaining({
-            Items: [expect.objectContaining(cacheBehavior)]
+            Items: expect.arrayContaining([
+              expect.objectContaining(cacheBehavior)
+            ])
           })
         })
       })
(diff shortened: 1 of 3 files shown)


packages/serverless-components/aws-cloudfront/test-utils.js
```diff
@@ -25,7 +25,9 @@ module.exports = {
       expect.objectContaining({
         DistributionConfig: expect.objectContaining({
           CacheBehaviors: expect.objectContaining({
-            Items: [expect.objectContaining(cacheBehavior)]
+            Items: expect.arrayContaining([
```
raffij, line 28: Would you prefer this is a separate helper? i copied the `arrayContaining` approach from your origins test.

dphang: I think it seems fine, we are anyway trying to not manage AWS manually in the future (instead using CDK or something similar)

dphang (approved):

dphang:
Ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/373623392

Issue #705, opened 2020-10-21
Conflicting rewrites #2 - Incorrect paths in Preview Mode (1.18.0-alpha.16)


patricktyndall:
New issue per maintainer admonishments. This is continued from https://github.com/serverless-nextjs/serverless-next.js/issues/696, same setup as there (@dphang).

### Describe the bug
When using top-level rewrite (per past issues -- see rewrites / repo below), upon visiting a route in preview mode, the incorrect URL is passed to `getStaticProps` (via `params`). Issue goes away if rewrites are removed, or if not in preview mode.

### Actual behavior
URL slug passed in `params` to `getStaticProps` is incorrect.

### Expected behavior
URL slug passed in `params` to `getStaticProps` is correct.

### Steps to reproduce
- Deploy my example
- Visit this URL: `/api/preview?secret=96ffe7d4-2ad2-4699-b64a-a5f74750962e&slug=/dynamic-1`
  - Either check logs, or simply look at the page. The `params` key incorrectly has part of the rewrite destination in it.
- For added fun, visit the non-rewritten URL for the same page `/api/preview?secret=96ffe7d4-2ad2-4699-b64a-a5f74750962e&slug=/content/dynamic-1`
  - The app now thinks we're at `/content/content/dynamic-1`

### Screenshots/Code/Logs
- Reproduction: https://github.com/patricktyndall/sls-next-bug-reproduction
- The rewrite in question:
```
[{
	source: '/:path*',
	destination: '/content/:path*',
}]
```
### Versions
- OS/Environment: Mac
- @sls-next/serverless-component version: `1.18.0-alpha.16`
- Next.js version: `9.5.5`

### Additional context
~If you visit `/content/dynamic-1` not in preview mode, you get a 404. This is also a bug, and is actually a regression added in `alpha.16`.~

dphang:
Thanks for creating the new issue - I thought it was related to just preview mode, not rewrites.

> If you visit /content/dynamic-1 not in preview mode, you get a 404. This is also a bug, and is actually a regression added in alpha.16.

In the local next server (yarn build && yarn start), `http://localhost:3000/content/dynamic-1` is also returning a 404 as well for me, when the rewrite below is added as you mentioned:

```js
[{
	source: '/:path*',
	destination: '/content/:path*',
}]
```

It seems to be the same behavior as `serverless-next.js`, so it seems the change is in line with this. Can you please confirm this?

[...]

PR #812, opened 2020-11-19
Update the rewrite behaviour to more closely match the Next.js Server

kirkness:
When attempting to use the default-lambda with `next-i18next` I noticed a couple of issues:
1. When a `getIntialProps` method doesn't resolve after calling `res.end()` the `page.render()` call never resolved and therefore responding with a 503. An example of this is shown in line [69](https://github.com/isaachinman/next-i18next/blob/master/src/middlewares/next-i18next-middleware.ts#L44-L75) of this middleware call. When redirecting, `next()` is not called.
    1. Fixed by wrapping both `page.render` and `responsePromise` in a `Promise.race` so that if `responsePromise` resolves before `page.render` we respond with the value of `responsePromise`.
2. The URI on `req.url` is being set as the value of the rewrite, this is different behavior than Next.js' server, the value of `req.url` should be the value of the originally requested URI.
    1. Fixed by passing the original URI into `nextAwsCloudfront` so that the return `req` has a `url` property that matches the original URI.


dphang:
Thanks! Does this fix this issue too? https://github.com/serverless-nextjs/serverless-next.js/issues/705#issuecomment-718272571

dphang:
Ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/373619191

kirkness:
> > The URI on req.url is being set as the value of the rewrite, this is different behavior than Next.js' server, the value of req.url should be the value of the originally requested URI.
> 
> Thanks! Does this fix this issue too? [#705 (comment)](https://github.com/serverless-nextjs/serverless-next.js/issues/705#issuecomment-718272571). Also maybe add a small test to verify this behavior (either in e2e tests or unit test)?

I don't think it will solve this, no, given `rewrittenUri` will only be set if `!isNonDynamicRoute && !isDataReq && customRewrite`, and the case you have referred to is where `isDataReq` is true. It does sound like the same problem though, essentially in the next.js data fetching methods `req.url` should be the originally fetched path rather than the file path. I will see if I can expand my solution to cover this case. Also, I'll add some tests, likely by end of next week 👍.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -401,7 +403,8 @@ const handleOriginRequest = async ({
   const { req, res, responsePromise } = lambdaAtEdgeCompat(
     event.Records[0].cf,
     {
-      enableHTTPCompression: manifest.enableHTTPCompression
+      enableHTTPCompression: manifest.enableHTTPCompression,
+      rewrittenUri
```
thchia, line 407: Although you're passing this, it isn't being used by the compat handler. Is that part missing?

kirkness: Good catch, added now.

kirkness:
> > The URI on req.url is being set as the value of the rewrite, this is different behavior than Next.js' server, the value of req.url should be the value of the originally requested URI.
> 
> Thanks! Does this fix this issue too? [#705 (comment)](https://github.com/serverless-nextjs/serverless-next.js/issues/705#issuecomment-718272571). Also maybe add a small test to verify this behavior (either in e2e tests or unit test)?

Test added!

dphang:
Thanks, looks like there are some conflicts. Could you please fix, then I'll merge it?

dphang:
Ran e2e tests: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/390452375

dphang:
Ah, looks like some e2e rewrite tests failed. I think now `{"query": {"slug": "123"}}` is being passed (same as Next.js server now) e.g for rewrites like `/path-rewrite/123` -> `/ssr-page` whereas we are currently expecting nothing in `{"query" : {}}`

Let me know if you can update the e2e tests, I think you just need to change the `expectedRewrite` e.g for here (https://github.com/serverless-nextjs/serverless-next.js/blob/2427517e093a7f938d22600771b86a6a36c40666/packages/e2e-tests/next-app/cypress/integration/rewrites.test.ts#L20) it would be `/ssr-page?slug=123` instead of `/ssr-page`. If not I can update it later this week.

dphang:
Thanks! Actually there are 7 other apps that need the same update to e2e tests, but I'll merge this first and fix up the tests later.

PR #818, opened 2020-11-24
klaw seems to be a dependency for "lambda-at-edge"

skorfmann:
When installing `lambda-at-edge` independently and running the example code

```
const path = require('path');
const { Builder } = require("@sls-next/lambda-at-edge");

const nextConfigDir = __dirname;
const outputDir = path.join(nextConfigDir, ".serverless_nextjs");

const builder = new Builder(
  nextConfigDir,
  outputDir,
  {
    cmd: './node_modules/.bin/next',
    cwd: process.cwd(),
    env: {},
    args: ['build']
  }
);

(async () => await builder.build())();
```

I get the following error:

```
 node build.js
node:internal/modules/cjs/loader:903
  throw err;
  ^

Error: Cannot find module 'klaw'
```

Installing `klaw` via npm / yarn fixes it.

### DIFF
--- a/packages/libs/lambda-at-edge/package.json
+++ b/packages/libs/lambda-at-edge/package.json
@@ -45,7 +45,6 @@
     "@types/klaw": "^3.0.1",
     "@types/node": "^14.0.14",
     "@types/path-to-regexp": "^1.7.0",
-    "klaw": "^3.0.0",
     "rollup": "^2.26.6",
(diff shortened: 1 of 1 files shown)


dphang:
Thanks. Ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/384259596. Will merge after they succeed.

dphang (approved):

PR #825, opened 2020-11-30
Fix not found for public files with encoded characters in path

### LINKED ISSUE #822: Files served from public folder where names contain spaces are not found
### Describe the bug
Files having names containing spaces inside public folder result in 404

### Actual behavior
When a file with a name containing spaces is requested, the request path is encoded like so: `/files/my%20special%20file.txt`.
The `manifest.json`  looks something like this
```json
{
   "publicFiles": {
    "/files/my special file.txt": "/files/my special file.txt"
  }
}
```

Hence the following lookup yields `undefined`
https://github.com/serverless-nextjs/serverless-next.js/blob/2427517e093a7f938d22600771b86a6a36c40666/packages/libs/lambda-at-edge/src/default-handler.ts#L259

and we end up with a 404 at some later point.

### Expected behavior
The file should be served.

### Steps to reproduce
1. put a file with a filename containing spaces inside the public directory
2. deploy
3. request the file

### Versions

- OS/Environment: Debian buster
- @sls-next/serverless-component version: current master

### Checklist

- [ x ] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
- [ x ] Please first try using the most recent `latest` or `alpha` `@sls-next/serverless-component` [release version](https://github.com/serverless-nextjs/serverless-next.js/releases), which may have already fixed your issue or implemented the feature you are trying to use. Note that the old `serverless-next.js` component and the `serverless-next.js` plugin are deprecated and no longer maintained.

confix:
URL-decode request path before checking for public file in `manifest.json`. Prevent 404 responses when filename contains encoded characters.

Another unit test for public directory serving is added to verify functionality.

Fixes #822

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -255,8 +255,10 @@ const handleOriginRequest = async ({
 
   const basePath = routesManifest.basePath;
   let uri = normaliseUri(request.uri);
+  const decodedUri = decodeURI(uri);
   const { pages, publicFiles } = manifest;
-  let isPublicFile = publicFiles[uri];
+
+  let isPublicFile = publicFiles[decodedUri];
   let isDataReq = isDataRequest(uri);
 
   // Handle redirects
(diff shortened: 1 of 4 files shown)


dphang:
Thanks! I ran the e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/390805926. I'll add another e2e test to test a file with spaces can be accessed after merging this.

dphang (approved):

Issue #821, opened 2020-11-26
Unable to create subdomain: The certificate that is attached to your distribution doesn't cover the alternate domain name (CNAME) that you're trying to add


huksley:
### Describe the bug

Unable to use 3rd level domain e.g. app.example.com

### Actual behavior

When I execute ´serverless´it gives me error "InvalidViewerCertificate: The certificate that is attached to your distribution doesn't cover the alternate domain name (CNAME) that you're trying to add."

### Expected behavior

CloudFront distribution with the correct cert and CNAMEs should be created successfully.

### Steps to reproduce

I am using subdomain like app.example.com and wildcard cert for it.
I have disabled www.app.example.com generation, and when I do `serverless` it gives me the error:

```
{
  code: 'InvalidViewerCertificate',
  time: 2020-11-26T07:21:25.232Z,
  requestId: '793c6576-6245-4915-8897-91ac4736bd96',
  statusCode: 400,
  retryable: false,
  retryDelay: 50.03442615769107
}

  47s › app › InvalidViewerCertificate: The certificate that is attached to your distribution doesn't cover the alternate domain name (CNAME) that you're trying to add. For more details, see: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/CNAMEs.html#alternate-domain-names-requirements
```

### Screenshots/Code/Logs

**serverless.yml**
```
app:
  component: "@sls-next/serverless-component@1.19.0-alpha.11"
  inputs:
    domain: ["app", "example.com"]
    publicDirectoryCache: false
    memory:
      defaultLambda: 1024
      apiLambda: 2048
    timeout:
      defaultLambda: 20
      apiLambda: 15
    bucketName: myapp-serverlessdeployment-bucket
    bucketRegion: eu-west-1
    description: "Serverless deployment for App"
    domainType: apex
```

### Versions

- OS/Environment: MacOS 10.15.7
- @sls-next/serverless-component version: @sls-next/serverless-component@1.19.0-alpha.11
- Next.js version: ^10.0.0 pinned 10.0.3

### Additional context

I created wildcard certificate in AWS Certificate Manager in `us-east-1` North Virginia region. 
I tried to specify this certificate by adding `certificateArn` to serverless.yml

### Checklist

- [x] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
[...]

Issue #690, opened 2020-10-18
Add tests for domain component 


dphang:
As seen in code coverage: https://codecov.io/gh/serverless-nextjs/serverless-next.js/tree/master/packages/serverless-components/domain, domain component has 0% coverage.

Adding tests for this should get us to ~90% overall coverage.

@SarKurd I remember you wanted to add tests for this. Let me know if you still interested to add tests or someone else can also take it up

SarKurd:
Currently in a very tight schedule but i'm still interested. it may take a while but going to try when i have time if nobody else has taken it yet.

confix:
I actually started writing a few tests for utils last night. 

As @dphang mentioned in https://github.com/serverless-nextjs/serverless-next.js/issues/826#issuecomment-737091177 I am wondering whether it makes sense to invest more time in testing the domain component.

Do you think it's worth it since all of the unit tests would break anyway? Anyhow, I could provide a PR anytime soon.

PR #826, opened 2020-12-01
Pass certificateArn to domain component

huksley:
Currently, declared on README.md page `certificateArn` does not passed to `domain` component, resulting in searching for domain certificate, which works bad for subdomains, i.e. app.example.com

See issue here: https://github.com/serverless-nextjs/serverless-next.js/issues/821

### DIFF
--- a/packages/serverless-components/nextjs-component/src/component.ts
+++ b/packages/serverless-components/nextjs-component/src/component.ts
@@ -633,7 +633,7 @@ class NextjsComponent extends Component {
       await createInvalidation({
         distributionId: cloudFrontOutputs.id,
         credentials: this.context.credentials.aws,
-        paths: cloudFrontPaths,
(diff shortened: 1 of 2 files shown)


dphang:
Thanks, since domain component doesn't have any test coverage yet, please test the change if it is working fine, then can merge it.

huksley:
@dphang Would you like me to write a test on it? I am currently deploying it to my infra and it is working.

dphang:
If you have time and it's simple that'd be great (there is an issue for this: https://github.com/serverless-nextjs/serverless-next.js/issues/690), but we may move away from it in the future (looking into more robust IaC tools like CDK). But we can merge this one first. Thanks!

dphang:
Ran e2e tests: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/395710571

PR #827, opened 2020-12-01
Add inputs.roleArn for existing lambda role

huksley:
Allows to specify ARN of existing role instead of creating a new one.

```yml
myNextApplication:
  component: "@sls-next/serverless-component@{version_here}"
  inputs:
    roleArn: "arn:aws:iam::123456789012:role/MyCustomLambdaRole"
```

- [x] Added to README.md
- [x] Added unit test to packages/serverless-components/nextjs-component/__tests__/custom-inputs.test.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,9 @@ myNextApplication:
 
 ### AWS Permissions
 
-By default the Lambda@Edge functions run using AWSLambdaBasicExecutionRole which only allows uploading logs to CloudWatch. If you need permissions beyond this, like for example access to DynamoDB or any other AWS resource you will need your own custom policy arn:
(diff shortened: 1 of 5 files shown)


dphang (approved):
LGTM! Just ran e2e test: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/395679068, can merge after

Arditc:
@dbhagen Just one question regarding this, why aren't we following the same approach of the lambda name?, for constancy and giving the lambda the correct roles:

```
  roleArn:
      defaultLambda: arn
      apiLambda: arn
```

The reason I say this is because I wouldn't want to give the `defaultLambda` access to the database.

huksley:
Yes, this can be improved if needed @Arditc. I.e use one role if specified directly or use specific roles per lambda if specified as a map as per syntax you show in example.

Issue #725, opened 2020-10-30
NextJS 10.0.0 <Image /> component causes 404


Ali762:
### Describe the bug

I realise NextJS 10 is only a couple of days old, but thought I'd just mention this to make sure it's captured.
Using the new NextJS <Image /> Component doesn't work.

### Actual behavior

Using the Image component results in a 404 error from the server. It works in dev, but deploying to AWS doesn't.
### Expected behavior

The image should return correctly.

### Screenshots/Code/Logs

Request URL: https://mywebsite.com/_next/image?url=https%3A%2F%2Fcdn.sanity.io%2Fimages%2F9h...etc etc
Request Method: GET
Status Code: 404 

Response Headers:
cache-control: public, max-age=0, s-maxage=2678400, must-revalidate
content-encoding: gzip
content-type: text/html
date: Fri, 30 Oct 2020 01:42:59 GMT
etag: W/"6805d78c611deceec9b5bd040ab84078"
last-modified: Fri, 30 Oct 2020 01:37:01 GMT
server: AmazonS3
status: 404
vary: Accept-Encoding
via: 1.1 2b782f5f082f9e98adf8c50f24b6bb6d.cloudfront.net (CloudFront)
x-amz-cf-id: XDhH5J8y7n052fhPQ6T9Rvh1yAptUwU9eh7sT1RQWbqBNNlyEGyuNw==
x-amz-cf-pop: HAM50-C3
x-cache: Error from cloudfront

### Versions

- OS/Environment:
@sls-next/serverless-component@1.17.0
"next": "^10.0.0",

dphang:
Thanks for reporting - we'll prioritize the Next.js 10 features right after 1.18. Just want to address a couple small things before releasing 1.18.

nfantone:
Came here for this. Experiencing the same, at the moment.

The [`Image` component introduced in Next 10](https://nextjs.org/blog/next-10#built-in-image-component-and-automatic-image-optimization), generates URLs such as https://domain.cloudfront.net/_next/image?url=https%3A%2F%2domain.cloudfront.net%2Fclients%2Fclien-58ab57a5-3575-5f32-8eeb-e3b77d7b7757.svg&w=256&q=75 and they all currently return `404`.

dphang:
[...]

PR #829, opened 2020-12-02
feat(lambda-at-edge): support image optimization

dphang:
Fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/725 by supporting the image optimizer introduced in Next.js 10 and hence the image component.

* Added new image lambda handler to handle image optimization requests only
* Most of this is a port of the Next.js server image-optimizer and related files, with a modification to retrieve the file from S3 instead (e.g for files in public directory).
* Added a bunch of node_modules used for sharp (image library). These have been specifically built for Lambda Node 12.x platform (using the lambdci/docker-lambda image) and added statically.


dphang:
@danielcondemarin this is done, could you please review when you have some time?

Unfortunately I don't know how to exclude the `sharp_node_modules` files from appearing in the PR. But the changes are mostly in `lambda-at-edge` in the `images` directory (for image optimizer) and a new `image-handler`. There are also some changes in Next.js component to add the new cache behaviors.

packages/libs/lambda-at-edge/src/images/imageOptimizer.ts
```diff
@@ -0,0 +1,400 @@
+/**
```
danielcondemarin, line 1: Did you consider using next's exported image-optimiser instead of copying it over? i.e.
```import imageOptimizer from "next/dist/next-server/server/image-optimizer"```

File: packages/libs/lambda-at-edge/types.d.ts [hunk 2]
```diff
@@ -11,6 +11,20 @@ export type DynamicPageKeyValue = {
   };
 };
 
+// Image optimization
+export type ImageConfig = {
+  deviceSizes: number[];
+  imageSizes: number[];
+  loader: "default" | "imgix" | "cloudinary" | "akamai";
```
danielcondemarin, line 18: Do we need `imgix`, `cloudinary`, `akamai` here?

packages/libs/lambda-at-edge/src/images/imageOptimizer.ts
dphang: Oh I didn't but I think it doesn't exactly work because it tries to write to the local directory and also it has some extra code that isn't needed

danielcondemarin: Fair enough

danielcondemarin (approved):

packages/libs/lambda-at-edge/src/images/imageOptimizer.ts
dphang: And also it tries to read images from local directory but we have it on s3 instead

packages/libs/lambda-at-edge/types.d.ts
dphang: Actually I didn't try these other loaders, will check on it but I think they go directly to those URLs

oddnavy:
Awesome work! 🤞  for a release including this one soon

marcfielding1:
Hey guys, I can test this on a LOT of images if you'd like - this is in the alpha release?

Issue #832, opened 2020-12-03
External URL rewrites


benjaminkay93:
**Is your feature request related to a problem? Please describe.**
Next.JS has as part of its core funcationality the ability to rewrite to external URL's.

In my particular example, we have our old servers running on a subdomain, as part of our migration we wish to point our www traffic to the subdomain if the route is not covered by next.js

**Describe the solution you'd like**

Respect the URL rewrite functionality of Next.JS

**Describe alternatives you've considered**

Having proxies to Next js, but that is now utililising all the joys of this component

**Additional context**
Readme hints this is being worked on, but the RFC has been closed https://github.com/serverless-nextjs/serverless-next.js/issues/587

dphang:
Yes, I had left it out at the time as when I tried to implement it, there were some issues due to merging the external rewrite's response headers into CloudFront (blacklisted headers from CloudFront). Will try to work on it this week. I think we can use `fetch` API to request the rewritten web page.

benjaminkay93:
i had a few problems myself writting a local `/api/graphql` proxy to an external graphql service (to get around CORS), I've posted what I had here although I will admit it was very much a plaster until I learnt about rewrites / this feature is added to serverless-next.js, i had nothing for page rewrites to proxy though :( 

```
import fetch from 'isomorphic-fetch';	

const ApolloProxy = async ({ body, headers }, res) => {	
  const requestHeaders = {	
    accept: 'application/json',	
  };	

  if (headers.authorization && headers.authorization !== 'undefined')	
    requestHeaders.authorization = headers.authorization;

  const apolloResponse = await fetch(process.env.APOLLO_SERVER_URL, {	
    method: 'POST',	
    body: JSON.stringify(body),	
    headers: {	
      ...requestHeaders,	
    },	
  });	

  const data = await apolloResponse.json();	
  res.send({ ...data });	
};	

export default ApolloProxy;
```

patricktyndall:
Hey @dphang is this on the latest alphas? I tried on `@sls-next/serverless-component@1.19.0-alpha.19` and no luck. I double checked all your troubleshooting steps in the RFC as well.

dphang:
@patricktyndall I think it is in 1.19.0-alpha.20. Not sure why it's not showing up in the GitHub releases page, maybe it failed to create the releases but it does show up in NPM. Anyway, I have also published another version as there are a few dependency updates.

PR #839, opened 2020-12-08
feat(lambda-at-edge): support external rewrites

dphang:
Fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/832


packages/libs/lambda-at-edge/src/routing/rewriter.ts
```diff
@@ -37,3 +38,65 @@ export function getRewritePath(
...
+  "x-accel-limit-rate",
+  "x-accel-redirect",
+  "x-cache",
+  "x-forwarded-proto",
+  "x-real-ip",
+  "content-length",
+  "host",
+  "transfer-encoding",
+  "via",
+  "content-encoding"
```
dphang, line 69: Not actually blacklisted but because the data returned is uncompressed (via fetch), copying this header tells CloudFront not to compress content, and then it fails at the browser since it cannot decompress already uncompressed content.

Issue #860, opened 2021-01-03
Types are not included in lambda-at-edge


simonireilly:
### Describe the bug

When using lambda-at-edge stand-alone the typescript compiler will throw an error trying to resolve `../types` from `dist/build.ts`.

```ts
node_modules/@sls-next/lambda-at-edge/dist/build.d.ts:3:137 - error TS2307: Cannot find module '../types' or its corresponding type declarations.

3 import { OriginRequestDefaultHandlerManifest, OriginRequestApiHandlerManifest, RoutesManifest, OriginRequestImageHandlerManifest } from "../types";
```

This is because the types file is not included in the `dist`.

### Expected behavior

The types should be included in the `dist`

### Steps to reproduce

This bug is reproduced in a PR here, where the standalone package is configured: https://github.com/simonireilly/nextjs-aws-cdk/pull/1

### Versions

- OS/Environment: Linux Ubuntu 20.04.1 LTS
- @sls-next/serverless-component version: "@sls-next/lambda-at-edge@latest": version "1.8.0-alpha.15"
- Next.js version: N/A

### Checklist

- [x] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
- [x] Please first try using the most recent `latest` or `alpha` `@sls-next/serverless-component` [release version](https://github.com/serverless-nextjs/serverless-next.js/releases), which may have already fixed your issue or implemented the feature you are trying to use. Note that the old `serverless-next.js` component and the `serverless-next.js` plugin are deprecated and no longer maintained.

PR #865, opened 2021-01-06
Update type declarations to be included in `dist` and exported at the root

simonireilly:
>closes: #860 

This PR will ensure that the types used in the `@sls-next/lambda-at-edge` are included in the distribution.

- Move the `types.d.ts` into `src/types.ts`
- Export all types from `src/types.ts` in `src/index.ts`

This means the types that have cross package uses are accessible from the root e.g:

```ts
import type { RoutesManifest } from '@sls-next/lambda-at-edge'
```

### DIFF
--- a/packages/libs/lambda-at-edge/src/types.ts
+++ b/packages/libs/lambda-at-edge/src/types.ts
@@ -1,4 +1,4 @@
-import {
+import type {
   CloudFrontRequest,
   CloudFrontEvent,
   CloudFrontResponse
--- a/packages/serverless-components/nextjs-component/src/component.ts
+++ b/packages/serverless-components/nextjs-component/src/component.ts
@@ -2,13 +2,13 @@ import { Component } from "@serverless/core";
 import { readJSON, pathExists } from "fs-extra";
 import { resolve, join } from "path";
 import { Builder } from "@sls-next/lambda-at-edge";
-import {
+import type {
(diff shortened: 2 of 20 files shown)


packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -2,13 +2,13 @@ import { Component } from "@serverless/core";
 import { readJSON, pathExists } from "fs-extra";
 import { resolve, join } from "path";
 import { Builder } from "@sls-next/lambda-at-edge";
-import {
+import type {
```
simonireilly, line 5: This was pulling from `lambda-at-edge/types.d.ts` now it will be from `lambda-at-edge/dist/index.ts`

packages/libs/lambda-at-edge/src/types.ts
```diff
@@ -1,4 +1,4 @@
-import {
+import type {
```
simonireilly, line 1: THis is the change, where this file is moved into `src` and renamed to be a `.ts` extension so it will be included in the dist without updating the `tsconfig.json` etc.

dphang:
Ran tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/467940780

dphang (approved):

dphang:
Thanks, LGTM!

Issue #777, opened 2020-11-07
RFC: state of Serverless Next.js and next steps


dphang:
Hi all!

I hope you are having a great weekend. I just wanted to take some time to give some updates on the current state of Serverless Next.js. This is a long post - TL;DR on bottom if you want. Ever since I've started using this early this year, and contributing just a few months back, we as a community have accomplished, as of version **1.18** or **1.19 alpha**:

* Stability improvements: for example, we added multiple end-to-end tests that deploy a real Lambda@Edge for both the current-minor and previous-minor Next.js versions, pinned dependencies, enabled security updates (automated dependency updates are a WIP via Renovate), and added a bunch more unit test coverage for the handler code.
* Build/deploy and configuration improvements: thanks to all the bug reports and suggestions, we've merged many improvements to configuring CloudFront inputs and other `serverless.yml` outputs, decoupling build step from the deploy step, etc.
* Feature parity: except for the recently released Next.js 10 features, some more involved features (like revalidate on SSG), and a few miscellaneous bugs or quirks, the majority of Next.js 9.x features should be implemented and fairly stable and well-tested.

...and more. Thank you for all your support and contributions, and for helping us test and report bugs with the latest alpha versions!

That being said, we've also noticed these two biggest pain points:

* Lambda@Edge and CloudFront are highly coupled with the core code of routing and rendering the correct pages. As an end user of this package, you may not notice it, but this actually makes it harder to support new platforms, such as Lambda, Cloudflare, GCP, Azure, etc. As a maintainer, it also makes it hard to separate testing the platform integration vs. testing the core code.
* Infrastructure management has been difficult and sometimes has inconsistent results, as there is no first-class state management like a proper infrastructure-as-code (IaC) tool. The reason for this is that the Serverless Components we use (aws-cloudfront, domain, etc.) weren't well maintained and interfaced with the AWS SDK directly and do not manage state except for saving the Lambda function name, CloudFront distribution ID, etc (in the `.serverless` directory).

This brings us to the next set of challenges:

* Decoupling and modularizing the runtime code itself so it's not tied Lambda@Edge. This will help us in separating concerns to the respective packages. Fortunately, we have [Rollup.js](https://rollupjs.org/guide/en/) or other bundlers which can largely help with this. I propose the following:
  - Core Next.js code: this is the core routing and rendering logic, all specific to Next.js rendering of SSR, SSG pages and doing non-dynamic, dynamic routing.
[...]

PR #878, opened 2021-01-18
CDK Construct

kirkness:
This may be wildly out of scope for this project at the moment(⁉️) as per https://github.com/serverless-nextjs/serverless-next.js/issues/777. I've been using CDK + Next.js with the sls-next `Builder` for around a year now on a couple of different projects. To make this possible I've evolved this CDK Construct which is by now fairly well battle-tested; all the while trying to keep it in line with the way that the `serverless-component` works. I'm quite keen on open-sourcing it and think it makes the most sense to exist in this project. Keen to know whether this is something that is of interest including in the project at the moment?

Todo:
- [x] Documentation
- [x] Unit tests
- [x] Reuse cache logic from serverless component
- [x] Function names should be more descriptive
- [x] Fix custom behaviors (should use new format)
- [x] Fix static image bevaviour
- [x] Identify why minify handlers returns 503

Future possibilities:
- Abstract infrastructure synthesis so that logic is shared between serverless and cdk deployments
- Add built-in optional builder
- Write using Jsii so that this construct can be used within other CDK languages

**Side-note:** I'm also keen on getting more involved with the maintenance and support for this project if there still a need (especially helping on the work outlined in #777).


danielcondemarin:
This is awesome @kirkness 🙌 

Actually I think this aligns well with what @dphang outlined in https://github.com/serverless-nextjs/serverless-next.js/issues/777. There is the question of CDK vs CDK for Terraform, but I don't see why we shouldn't make a start with plain CDK if you've done all the hard work already. 
The most important thing is that CDK solves the infrastructure state problem we've always had with serverless components beta. It also opens up the possibility of supporting serverless components GA via CloudFormation which I've briefly discussed with @dphang. Plus CDK is not going away any time soon as AWS is investing a lot in it. 

I will try to test out this branch over the weekend and will feedback. 

> 'm also keen on getting more involved with the maintenance and support for this project if there still a need

We do need more contributors and I can add you to the repo's maintainers if that's something you're interested in 🙂

CC'ing @dphang for any thoughts on this.

packages/serverless-components/nextjs-cdk-construct/src/index.ts
```diff
@@ -0,0 +1,376 @@
...
+        staticAsset
+      );
+      if (!fs.existsSync(assetPath)) return;
+      new s3Deploy.BucketDeployment(this, `AssetDeployment_${staticAssetKey}`, {
+        destinationBucket: bucket,
+        distribution: this.distribution,
+        destinationKeyPrefix: staticAsset,
+        distributionPaths: ["/*"],
+        sources: [s3Deploy.Source.asset(assetPath)],
+        cacheControl: [
```
dphang, line 309: Pages would need different cache control here: `public, max-age=0, s-maxage=2678400, must-revalidate`.

And also since versioning pages, we are clearing old versions of pages with the latest deployment code (1.19 alpha) (keeping only last two deployments) to save S3 space

packages/serverless-components/nextjs-cdk-construct/__tests__/fixtures/next-boilerplate/default-lambda/index.js
```diff
@@ -0,0 +1 @@
+exports.handler = async () => ({ statusCode: 200 });
```
dphang, line 1: For better readability of all these added JS/JSON files, could you please format these test files using prettier?

dphang:
@kirkness thanks a lot for this! Yea, we have been talking about adding CDK for a while now but just haven't had resources. So this is a good first step! We are also trying to upgrade to Serverless Components GA, if you are interested as mentioned it might go hand-in-hand with this as we are trying to make deployments more seamless in the near future (+ there is financial sponsorship from Serverless Framework for doing so, which we would be happy to split based on resourcing). Do let @danielcondemarin know and we can add you to the Slack too.

kirkness:
Great news, would be awesome to join you on Slack to kick this off! My email is just my first name (Henry) + my company domain (planes.studio).

In the meantime, I'll revert this PR to a draft and flesh out a plan of attack with some of the additional work that's needed to go into it.

packages/serverless-components/nextjs-cdk-construct/src/index.ts
kirkness: I will have a look into how static pages are being versioned, but I presume this means no invalidation is required for these either which is nice! This is something I wanted to revisit as it's currently creating an invalidation on all resources every deploy which is unnecessary given everything is versioned?

kirkness: > And also in the latest 1.19 alpha, we are now versioning pages so only the last two deployments are kept to save S3 space.

I believe this is actually [handled by CDK itself](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-s3-deployment-readme.html#prune) as it automatically 'syncs' source assets, deleting ones that no longer exist. I'm assuming the reason you are keeping the last 2 deployments is for safety and thus this behavior is not needed for the construct?

packages/libs/s3-static-assets/package.json
```diff
@@ -33,13 +33,12 @@
   "homepage": "https://github.com/serverless-nextjs/serverless-next.js#readme",
   "dependencies": {
     "aws-sdk": "2.728.0",
+    "fast-glob": "^3.2.5",
```
kirkness, line 36: Refactored from using `klaw` to `fast-glob` as CDK relies on synchronous operations during synthesis.

packages/libs/s3-static-assets/src/index.ts
```diff
@@ -24,149 +23,160 @@ type UploadStaticAssetsOptions = {
...
-    credentials,
-    basePath,
-    publicDirectoryCache,
-    nextConfigDir
-  } = options;
-  const s3 = await S3ClientFactory({
-    bucketName,
-    credentials: credentials
-  });
+const getAssetDirectoryFileCachePolicies = (
```
kirkness, line 38: I've taken all the logic from `uploadStaticAssetsFromBuild` that determines the cache-control for each file in `/assets` and refactored it to be agnostic to the deployment mechanism. This file is imported into the CDK construct now, whilst also used below in `uploadStaticAssetsFromBuild`. I think we could do with a grander refactor of this form that removes the deployment code from the synthesis.

packages/libs/s3-static-assets/src/lib/filterOutDirectories.ts
```diff
@@ -1,3 +0,0 @@
-import { Item } from "klaw";
-
-export default (fileItem: Item): boolean => !fileItem.stats.isDirectory();
```
kirkness, line 3: Not needed anymore as we get `fast-glob` to only read files.

packages/serverless-components/nextjs-cdk-construct/src/index.ts
kirkness: @dphang - update on this [here](https://github.com/serverless-nextjs/serverless-next.js/pull/878/commits/ed731a2e90c7427cbfe2287612d3be7d6914ebc1). I've refactored the `uploadStaticAssetsFromBuild` which was determining the cache-control for files so that it can be used in the CDK construct.

packages/serverless-components/nextjs-cdk-construct/__tests__/fixtures/next-boilerplate/default-lambda/index.js
kirkness: Done in https://github.com/serverless-nextjs/serverless-next.js/pull/878/commits/ea6cc1605efab672a2aa586fe43cd3de8fec38d2

kirkness: Although, prettier has not updated `__tests__/fixtures/next-boilerplate/default-lambda/index.js`...

iDVB:
Thanks for working on this @kirkness!
Our company may be interested in the plain construct and helping out if we can.

One thing to think about is, have you considered the potential order of events that may be needed before the frontend assets are deployed? One think I've come up against when using CDK to deploy frontend assets is that there is currently no way that I saw to have "backend" infra deploy first and to pass API endpoints, outputs, into the frontend construct so they can be bundled into the frontend assets. Have you found away around this? Is there anything I might be able to help out with in these efforts?

kirkness:
> One thing to think about is, have you considered the potential order of events that may be needed before the frontend assets are deployed? One think I've come up against when using CDK to deploy frontend assets is that there is currently no way that I saw to have "backend" infra deploy first and to pass API endpoints, outputs, into the frontend construct so they can be bundled into the frontend assets. Have you found away around this? Is there anything I might be able to help out with in these efforts?

👍, sure I think I know what you mean, as you might:
1. Build the next app (this only point in time in which you can set Next.js env variables)
2. Synth + Deploy your app via CDK (this is the point in time in which you have the values that you might use for your env vars, e.g. API endpoints, S3 bucket names...)

I think at the very least, if the bundling was done within the construct, then you'd be able to pass in non-tokenized values as env vars, all inline with your CDK code.

iDVB:
> I think at the very least, if the bundling was done within the construct, then you'd be able to pass in non-tokenized values as env vars, all inline with your CDK code.

This would be SUPER ideal. Any work towards this working now?

kirkness:
> > I think at the very least, if the bundling was done within the construct, then you'd be able to pass in non-tokenized values as env vars, all inline with your CDK code.
> 
> This would be SUPER ideal. Any work towards this working now?

I think all that would be needed is to set up [Asset Bundeling](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-s3-assets-readme.html#asset-bundling), which would allow running the `Builder` in a Docker container... By the looks of it you can pass [`environment`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.BundlingOptions.html#environmentspan-classapi-icon-api-icon-experimental-titlethis-api-element-is-experimental-it-may-change-without-noticespan) into the Docker container. If this accepts tokenized values, it solves the entire problem, if not, then still pretty neat.

kirkness:
> Might have a look at a POC now. Watch this space. 👍

Parking this now, the CDK asset bundling process starts after the CDK constructs are created. At the point when we create the Next.js construct we need the next.js app to be bundled, this is because the infrastructure bases itself on the shape of the bundled output. So similar in part to the issue you outlined with env variables (that this was trying to resolve). On top of this, the env vars passed into the asset bundling process are not resolved, and so you end up with tokenized values.

More research needed - an easy way would be just to bundle without the use of the asset bundling construct, although keen to find a best-practice approach!

iDVB:
> the infrastructure bases itself on the shape of the bundled output
@kirkness So in this case you need Nextjs to bundle FIRST so you can look at the assets and setup the infrastructure accordingly? Can you give a specific example? 

This sounds like the opposite issue as the env vars needing infra to be created first. I've often wondered if the solve for env vars wasnt just to make the frontend code expect a static file, say, `infra.json` and then build the frontend first.... then the backend, then produce that file and ensure both the frontend assets and that file make it into deployed code.

kirkness:
> > the infrastructure bases itself on the shape of the bundled output
> > @kirkness So in this case you need Nextjs to bundle FIRST so you can look at the assets and setup the infrastructure accordingly? Can you give a specific example?
> 
> This sounds like the opposite issue as the env vars needing infra to be created first. I've often wondered if the solve for env vars wasnt just to make the frontend code expect a static file, say, `infra.json` and then build the frontend first.... then the backend, then produce that file and ensure both the frontend assets and that file make it into deployed code.

Yes, sorry I meant more in the sense that it's an issue with the ordering of events generally. I think creating some form of env file, as you say, generated once the variables are available may be the only way.

iDVB:
> issue with the ordering of events

Ya, I've been chasing this challenge for some time with CDK. In the Serverless Framework world, we have programmatic access to all stages of the workflows with plugins.  With CDK, there is even a way to create a single construct that nets out into two related Stacks (I believe), Currently however, I think you [can't use the outputs from one stack as inputs to the other](https://github.com/aws/aws-cdk/issues/49)... (I could be wrong here) which would be the primary reason to go this route in this case. Also, there are a number of people asking for [CDK programmatic access](https://github.com/aws/aws-cdk/issues/601), but I haven't seen anything yet. I think that could also be an avenue here. 

Super interested in all these topics though, as its the main gating factors in front of us to dive way into CDK.

kirkness:
> > issue with the ordering of events
> 
> Ya, I've been chasing this challenge for some time with CDK. In the Serverless Framework world, we have programmatic access to all stages of the workflows with plugins. With CDK, there is even a way to create a single construct that nets out into two related Stacks (I believe), Currently however, I think you [can't use the outputs from one stack as inputs to the other](https://github.com/aws/aws-cdk/issues/49)... (I could be wrong here) which would be the primary reason to go this route in this case. Also, there are a number of people asking for [CDK programmatic access](https://github.com/aws/aws-cdk/issues/601), but I haven't seen anything yet. I think that could also be an avenue here.
> 
> Super interested in all these topics though, as its the main gating factors in front of us to dive way into CDK.

One possibility could be to use [Custom Resources](https://docs.aws.amazon.com/cdk/api/latest/docs/custom-resources-readme.html), which sort-of allows programmatic access to AWS, whilst keeping it in line with CDK constructs, etc. Theoretically, you could have the source code of a next app staged to S3, a custom construct that has a lambda with access to the source code, then have the lambda compose and deploy the serverless-next app. I think that you'd be able to read the resolved tokens from within the lambda, meaning that you could have client and server env vars fairly trivially. Likely plenty of downside to this approach though!

iDVB:
Yup, very familiar with Custom resources. I’m a bit of a CFN purest though
and wouldn’t want and resources to be managed (Create, Update, Delete)
inside of the custom resource vs in a CFN template. You can basically do
anything you want in a custom resource.

What about that multi-stack approach though? Sounds like they are close to
a feature that could create artifacts from one stack to the other?

packages/libs/s3-static-assets/src/index.ts
kirkness: This didn't make sense to use in the end as the data structure required to work with the constructs didn't fit this output enough to make it worth it. Happy to revert the changes to this module if you think its needed.

packages/serverless-components/nextjs-cdk-construct/src/utils/readAssetsDirectory.ts
```diff
@@ -0,0 +1,63 @@
...
+      path: staticPages,
+      cacheControl: SERVER_CACHE_CONTROL_HEADER
+    },
+    nextData: { path: nextData, cacheControl: SERVER_CACHE_CONTROL_HEADER },
+    nextStatic: {
+      path: nextStatic,
+      cacheControl: IMMUTABLE_CACHE_CONTROL_HEADER
+    }
+  });
+};
```
kirkness, line 61: This should match the behavior in `getAssetDirectoryFileCachePolicies`, although rather than allowing users (via inputs) to add cache control to files individually, and also never adding cache control to HTML dynamic files, the construct instead relies on invalidating all paths.

packages/serverless-components/nextjs-cdk-construct/src/utils/readInvalidationPathsFromManifest.ts
```diff
@@ -0,0 +1,26 @@
...
+    ...Object.keys(manifest.pages.html.nonDynamic),
+    ...Object.keys(manifest.pages.ssr.dynamic).map(
+      dynamicPathToInvalidationPath
+    ),
+    ...Object.keys(manifest.pages.ssr.nonDynamic),
+    ...Object.keys(manifest.pages.ssg?.dynamic || {}).map(
+      dynamicPathToInvalidationPath
+    ),
+    ...Object.keys(manifest.pages.ssg?.nonDynamic || {})
+  ];
```
kirkness, line 25: Invalidate all known nonDynami paths, and invalidate all dynamic paths using a wildcard i.e. `/users/*`

packages/libs/s3-static-assets/src/index.ts
dphang: I think it should be ok to keep it.

packages/libs/s3-static-assets/tests/utils.test.ts
```diff
@@ -0,0 +1,21 @@
+import readDirectoryFiles from "../src/lib/readDirectoryFiles";
+import * as path from "path";
+
+describe("readDirectoryFiles", () => {
+  it("returns an empty array when the file is not found", async () => {
```
dphang, line 5: I guess `async` is not needed for these tests now

packages/serverless-components/nextjs-cdk-construct/__tests__/construct.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      Runtime: Runtime.NODEJS_10_X.name
+    });
+  });
+
+  it("passes correct lambda options to underlying lambdas when object passed", () => {
+    const stack = new Stack();
+    new NextJSLambdaEdge(stack, "Stack", {
+      serverlessBuildOutDir: path.join(__dirname, "fixtures/next-boilerplate"),
+      runtime: {
+        defaultLambda: Runtime.PYTHON_3_8,
```
dphang, line 38: Not sure why we are using Python/Go/Java runtimes here. Is it just to test that you can pass different runtimes for unit tests?

packages/serverless-components/nextjs-cdk-construct/__tests__/construct.test.ts
```diff
@@ -0,0 +1,169 @@
+import "@aws-cdk/assert/jest";
+import { SynthUtils } from "@aws-cdk/assert";
+import { Stack } from "@aws-cdk/core";
+import path from "path";
+import { NextJSLambdaEdge } from "../src";
+import { Runtime } from "@aws-cdk/aws-lambda";
+import { Certificate } from "@aws-cdk/aws-certificatemanager";
+import { HostedZone } from "@aws-cdk/aws-route53";
+
+describe("CDK Construct", () => {
```
dphang, line 10: New to CDK myself, but these LGTM in terms of style

packages/serverless-components/nextjs-cdk-construct/__tests__/fixtures/next-boilerplate/default-lambda/manifest.json
```diff
@@ -0,0 +1 @@
+{"buildId":"IOjzqKr7ylEZVL7Xy7FMd","logLambdaExecutionTimes":true,"pages":{"ssr":{"dynamic":{"/users/:id":{"file":"pages/users/[id].js","regex":"^\\/users(?:\\/([^\\/#\\?]+?))[\\/#\\?]?$"}},"nonDynamic":{"/users":"pages/users.js","/_error":"pages/_error.js"}},"html":{"dynamic":{},"nonDynamic":{"/about":"pages/about.html","/":"pages/index.html","/404":"pages/404.html"}}},"publicFiles":{"/planes.jpg":"planes.jpg"},"trailingSlash":false,"domainRedirects":{},"enableHTTPCompression":false}
```
dphang, line 1: Format this JSON file for better readability (since they are used in unit tests)

packages/serverless-components/nextjs-cdk-construct/__tests__/fixtures/next-boilerplate/image-lambda/index.js
```diff
@@ -0,0 +1 @@
+exports.handler = async () => ({ statusCode: 200 });
```
dphang, line 1: Format

packages/serverless-components/nextjs-cdk-construct/__tests__/fixtures/next-boilerplate/image-lambda/manifest.json
```diff
@@ -0,0 +1 @@
+{"domainRedirects":{},"enableHTTPCompression":false}
```
dphang, line 1: Format

packages/serverless-components/nextjs-cdk-construct/src/utils/readInvalidationPathsFromManifest.ts
dphang: Can user still invalidate all paths or custom set of paths? I think because if there are many individual invalidations the costs would add up eventually.

dphang (commented):
I think LGTM in general, just a few minor comments and not sure if it's working with i18n subpaths (that was a recent change I made)

packages/serverless-components/nextjs-cdk-construct/__tests__/construct.test.ts
kirkness: Correct, i thought this might look a lil strange. Yes, this test just asserts that the runtime prop correctly passes through to the CF template.

packages/serverless-components/nextjs-cdk-construct/__tests__/construct.test.ts
kirkness: Im sure there are many many more test cases, however, keeping light for now whilst this is still "experimental" and subject to change.

kirkness:
> I think LGTM in general, just a few minor comments and not sure if it's working with i18n subpaths (that was a recent change I made)

Cheers @dphang - just ran through those changes. Is i18n a blocker/substantial change? Will put in follow up PR if all good with you?

dphang:
> > I think LGTM in general, just a few minor comments and not sure if it's working with i18n subpaths (that was a recent change I made)
> 
> Cheers @dphang - just ran through those changes. Is i18n a blocker/substantial change? Will put in follow up PR if all good with you?

I think it should probably work as is since it was just modifying the output HTML/JSON files to `.serverless_nextjs` to add the locale files, and also I updated the default Lambda handler. As long as you are uploading those, it should work. Yea, you can follow up in another PR.

Also, looks like some of the new tests in the CI build is failing now.

Thanks for this!

kirkness:
> > > I think LGTM in general, just a few minor comments and not sure if it's working with i18n subpaths (that was a recent change I made)
> > 
> > 
> > Cheers @dphang - just ran through those changes. Is i18n a blocker/substantial change? Will put in follow up PR if all good with you?
> 
> I think it should probably work as is since it was just modifying the output HTML/JSON files to `.serverless_nextjs` to add the locale files, and also I updated the default Lambda handler. As long as you are uploading those, it should work. Yea, you can follow up in another PR.
> 
> Also, looks like some of the new tests in the CI build is failing now.
> 
> Thanks for this!

Nice, sounds like it'll work then as each directory of the output is uploaded to s3/lambda as the sls component does. Ive updated the test snapshots so that they pass.

packages/serverless-components/nextjs-cdk-construct/src/utils/readInvalidationPathsFromManifest.ts
kirkness: Good catch, Ill add an option for this. I also think that this function should probably filter out any subpaths of wildcards, as currently it would still create an invalidation for `/users/*` and `/users/some/path` if `/users/:id` and `/users/some/path` exist... However `/users/some/path` would actually be invalidated by `/users/*`.

kirkness: @dphang, I've just put in a fix for this here: https://github.com/serverless-nextjs/serverless-next.js/pull/878/commits/7aeaec92c91f42f051b28c31863cb3cbe75e9966

dphang:
> > > > I think LGTM in general, just a few minor comments and not sure if it's working with i18n subpaths (that was a recent change I made)
> > > 
> > > 
> > > Cheers @dphang - just ran through those changes. Is i18n a blocker/substantial change? Will put in follow up PR if all good with you?
> > 
> > 
> > I think it should probably work as is since it was just modifying the output HTML/JSON files to `.serverless_nextjs` to add the locale files, and also I updated the default Lambda handler. As long as you are uploading those, it should work. Yea, you can follow up in another PR.
> > Also, looks like some of the new tests in the CI build is failing now.
> > Thanks for this!
> 
> Nice, sounds like it'll work then as each directory of the output is uploaded to s3/lambda as the sls component does. Ive updated the test snapshots so that they pass.

Thanks, I ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/565470296

dphang:
@kirkness looks like everything is good. LMK if ready to merge?

kirkness:
> @kirkness looks like everything is good. LMK if ready to merge?

Cheers @dphang, good to merge from my perspective!

dphang (approved):

ibrahimcesar:
Great work! I am looking towards this for some time. There are some example repo for me take a look? Tried to add to a codebase and even started from scratch but cdk deploy only generates a new config, cdk.out remains empty and the  command stalled. When I added the builder debug flag true it just showme the log for the application starting. 

Just a barebones example besides  the code provided so far in the docs would be great, Because of my limits I was not able to use it 😓 

Thanks in advance!

kirkness:
Hey @ibrahimcesar - sounds odd, I don't suppose you could push your example up to a repo (just the one you did from scratch)?

ibrahimcesar:
Sure @kirkness !
Please note I maybe missing something crucial, just trying to understand how to use. I think is a really great work. This is my PoC repo: https://github.com/ibrahimcesar/nextjs-starter-cdk-constructor

Thanks!

kirkness:
> Sure @kirkness !
> Please note I maybe missing something crucial, just trying to understand how to use. I think is a really great work. This is my PoC repo: https://github.com/ibrahimcesar/nextjs-starter-cdk-constructor
> 
> Thanks!

Cheers @ibrahimcesar, I've updated the docs in this [PR](https://github.com/serverless-nextjs/serverless-next.js/pull/908), it looks like you may have a couple of other unrelated issues in that repo, but hopefully they'll be obvious when it actually starts to build.

ibrahimcesar:
Thanks @kirkness ,

Yes, there are some configs missing, but "life finds a way"
I was able to deploy as expected, thanks a lot for the help! 👍🏼 

One hookie mistake I made, and l'll leave here as reference, is that I changed my `next.config.js` to

```
module.exports = {
  distDir: 'build',
}
```

Since I saw it was accessing the files at "build", but just leave the default build, or will fail because it will lookup in `.next`.

Updated my simple repo with a working version! Thanks a lot!

iDVB:
Wow, greet work @kirkness !

Now, my specific interest is in the CDK construct alone. 
What is the current state of that now since this merge?
- does it work on its own?
- Did you overcome the issue we spoke about in regards to baking stack values into the frontend build?
- Does it have feature parody with hosting on Vercel? What’s missing?

donaldpipowitch:
(FYI for others: Not sure if it's my fault, but I tried to play around with this CDK Construct and always hit a `s3:PutBucketPolicy Access Denied` error. As far as I can tell I should have this permission (also tried with `{Allow:*, Resource:*}`). There are some other `Access Denied` errors reported recently, but they seem to be a different problem. I tried a to downgrade several libs - either from `@sls-next` or `@aws-cdk` and tried different regions (`us` and `eu`), but no luck.)

iDVB:
@kirkness  When using the CDK construct outside of serverless components. what is the best way to organize your 
CDK stack files, package.json etc, vs all the files as part of the nextjs src/build files. 

I thought it would just be as simple as building in completely separate directory and just copying in the build files, but the CDK build process needed all the next apps deps and now its saying it needs the src pages directory too. 

Not sure if its possible or the best way to keep some separation between the CDK stack files and the nextjs app its deploying?

kirkness:
Glad you got it working @iDVB, feel free to create issues if you have any other problems!

revmischa:
Curious, is there a supported/recommended way to deploy a NextJS app with CDK but without Serverless?

ibrahimcesar:
@revmischa What do you mean by supported? Like a package? Because you only need to build your own constructor but there's a lot for React applications as whole in community. CDK is a general purpose tool so you can create an EC2 Instance and run NextJS inside or use ECS if your application is on Docker or even Fargate, but this service is considered serverless. But CDK gives you access to create any kind of infrastructure on AWS so you can do as you want.

revmischa:
> @revmischa What do you mean by supported? Like a package? Because you only need to build your own constructor but there's a lot for React applications as whole in community. CDK is a general purpose tool so you can create an EC2 Instance and run NextJS inside or use ECS if your application is on Docker or even Fargate, but this service is considered serverless. But CDK gives you access to create any kind of infrastructure on AWS so you can do as you want.

Yeah ideally deploying it using the https://docs.aws.amazon.com/cdk/api/latest/docs/aws-lambda-nodejs-readme.html construct. Just not sure on what to put in my handler.

ibrahimcesar:
@revmischa I don't think you gonna achieve what you want with this constructor in particular. It is serverless and is a Lambda. The architecture is a bit different, I don't think this is the right forum. You can ask me directly and I can help you.

Issue #721, opened 2020-10-29
Next.js 10 localization in pathname


katspaugh:
The new built-in localization in Next.js 10 allows to have the locale as a path prefix, e.g.:

```
/about
/de/about
/es/about
```

All three routes will open `pages/about.tsx`. This works fine when I run it locally.
When I deploy it using serverless-next.js, it gives a 404 for `/de/about` and `/es/about`. Just `/about` still works.

Is this something that the Serverless Next.js Component doesn't support yet, or am I doing something wrong?
Thanks!

dphang:
Yeah, Next.js 10 just came out yesterday, so it'll take some time to add those new features, since we have custom routing/rendering logic optimized for Lambda@Edge. I've tagged this with enhancement.

But I've tested the existing end-to-end tests with Next.js 10 and nothing broke, so looks like all the existing features work as expected.

LuisFros:
I am also have the same issue `404` on not root pages. 
- On the root page (`/en` for example) though, there is a strange behaviour when refreshing page that has locale subpath: 
![image](https://user-images.githubusercontent.com/21247721/97755130-66356c00-1ad7-11eb-912d-2eba19ab8ac5.png)
- There are **many** redirects with 307 and it crashes.

Lunnatica:
Hi! I was wondering if there are any updates on this?

dphang:
Yes, I have not had a chance to look at it yet, been mostly on vacation mode the past few weeks so haven't worked on the code recently. Hoping to get to it once I'm back next week :)

Lunnatica:
Thanks for the update! :) If it helps, when I tried deploying my site with the i18n solution it worked well on all locales, except for the default one, which returned 404 for all routes unless I manually added the locale prefix to the URL path. 

Hope you enjoy the rest of your holiday! :)

Arditc:
Hi, I was wondering if there was any updates on this issue?

dphang:
I am planning to continue on this during this week. Apologies since lately I was the only contributor for this, so haven't had enough time to both triage issues and work on the codebase..

benjipott:
related to https://github.com/serverless-nextjs/serverless-next.js/issues/830

dphang:
[...]

PR #883, opened 2021-01-22
feat(lambda-at-edge): start copying locale-specific page/data files t…

dphang:
…o output assets

* First part of: https://github.com/serverless-nextjs/serverless-next.js/issues/721, to copy all locale-specifi page/data files to the assets output directory.
* Second part is to add the routes for these pages in manifest
* Need to add some unit tests


karthikeyan676:
When can we start to use this feature?

PR #889, opened 2021-01-27
Update for README.md

peniakoff:
- update related to new configurable inputs

### DIFF
--- a/packages/libs/lambda-at-edge/README.md
+++ b/packages/libs/lambda-at-edge/README.md
@@ -8,23 +8,32 @@ This library was created to decouple the core logic of deploying serverless rend
 const path = require('path');
 const { Builder } = require("@sls-next/lambda-at-edge");
 
-const nextConfigDir = '/dir/to/my/nextapp';
-const outputDir = path.join(nextConfigDir, ".serverless_nextjs");
+const nextConfigPath = '/path/to/my/nextapp';
+const outputDir = path.join(nextConfigPath, ".serverless_nextjs");
 
 const builder = new Builder(
-  nextConfigDir,
+  nextConfigPath,
   outputDir,
   {
     cmd: './node_modules/.bin/next',
     cwd: process.cwd(),
     env: {},
-    args: ['build']
+    args: ['build'],
+    minifyHandlers: true,
+    enableHTTPCompression: true
   }
 );
 
-await builder.build();
+await builder.build()
+    .then(() => {
+      console.log("Application built successfully!");
+    })
+    .catch((e) => {
+      console.log("Could not build app due the exception: ", e);
+      process.exit(1);
+    });
 ```
-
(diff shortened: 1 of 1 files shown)


packages/libs/lambda-at-edge/README.md
```diff
@@ -8,42 +8,61 @@ This library was created to decouple the core logic of deploying serverless rend
...
+  nextConfigPath,
   outputDir,
   {
     cmd: './node_modules/.bin/next',
     cwd: process.cwd(),
     env: {},
-    args: ['build']
+    args: ['build'],
+    minifyHandlers: true,
+    enableHTTPCompression: true
```
danielcondemarin, line 23: Minor thing but best to show enableHTTPCompression as false to discourage users from using it. Gzip compression can be provided by CloudFront (see https://github.com/serverless-nextjs/serverless-next.js/pull/692).

danielcondemarin (approved):
Just a minor suggestion as above 🙂

packages/libs/lambda-at-edge/README.md
peniakoff: Yep, I agree, but I'm using only this @sls-next/lambda-at-edge component to building my Next.js application, I have configured correct my Cloudfront distributions (by setting CacheBehavior's Compress as `true`) and gzip compression didn't work. That is why I started my investigation, what's wrong. Finally I have found new configuration inputs I could use for `Builder`. Without `enableHTTPCompression` set as `true` Lambda@Edge didn't compress response with gzip. Built lambda had `enableHTTPCompression` in defaults set to `false` and variable `let shouldGzip = enableHTTPCompression && isGzipSupported(headers)` was always `false`. Maybe we should change here for  `let shouldGzip = isGzipSupported(headers)` and leave decission "use or not use gzip compression" only for lambda based on the headers comming from Cloudfront? What do you think, @danielcondemarin ?

danielcondemarin: Sounds like an issue with your distribution. I confirmed recently with the folks from Cloudfront that it should automatically compress the lambda@response if enabled, https://twitter.com/cristiangraz/status/1306277707353853952?s=20

Having said that of course you can use the flag, just bear in mind it might not be as efficient as Cloudfront doing it, and might add some extra ms of latency

peniakoff: I see. IMO the problem was to large response send back from Lambda@Edge to Cloudfront. Lambda sent correct payload, but distribution could not read it propertly. Maybe it was related to limitations described here: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-requirements-limits.html#lambda-requirements-see-limits
But I'm pretty sure distribution was configured fine.

danielcondemarin: I'll merge the PR as is and add some clarification around the enableHTTPCompression flag 👍

peniakoff: Good, thanks @danielcondemarin! :smile: 
I'll let you know about situation with my Cloudfront distribution config. I'm testing now another settings.

PR #912, opened 2021-03-02
fix(cdk): resources should not be assigned names

kirkness:
When deploying a second Next.js app to the same AWS account the hard-coded names in the construct (function names and cache policy names) will prevent the stack from deploying as you can't have duplicates under the same account. It's preferred to leave undefined for CDK to assign unique names to the resources instead.


dphang:
Looks like some unit tests failed due to the name change?

dphang (approved):
LGTM!

Issue #929, opened 2021-03-05
Fetch redirect mode for rewrites is following 302 that is not setting expected cookies on the client


ungarida:
**Is your feature request related to a problem? Please describe.**
I'm integrating the NextJS app with a legacy WebApp following the https://nextjs.org/blog/incremental-adoption#rewrites approach. The legacy WebApp is returning 302 redirects with cookies on some paths and this is not working correctly because the Default Lambda is using fetch in redirect='follow' mode. This is the code I see in the Lambda handling redirects:

```js
                                // HTTP fetch step 5.5
				switch (request.redirect) {
					case 'error':
						reject(new FetchError(`uri requested responds with a redirect, redirect mode is set to error: ${request.url}`, 'no-redirect'));
						finalize();
						return;
					case 'manual':
						// node-fetch-specific step: make manual redirect a bit easier to use by setting the Location header value to the resolved URL.
						if (locationURL !== null) {
							// handle corrupted header
							try {
								headers.set('Location', locationURL);
							} catch (err) {
								// istanbul ignore next: nodejs server prevent invalid response headers, we can't test this through normal request
								reject(err);
							}
						}
						break;
					case 'follow':
						// HTTP-redirect fetch step 2
						if (locationURL === null) {
							break;
						}

						// HTTP-redirect fetch step 5
						if (request.counter >= request.follow) {
							reject(new FetchError(`maximum redirect reached at: ${request.url}`, 'max-redirect'));
							finalize();
							return;
						}

```
If in the Default Lambda I force it to manual:

```js
		this[INTERNALS$2] = {
			method,
			redirect: init.redirect || input.redirect || 'manual',
			headers,
			parsedURL,
			signal
		};
```

**Describe the solution you'd like**
An option to specify how to handle redirects in rewrites.

dphang:
I think the Next local server actually doesn't seem to follow redirects when doing rewrites. We could update the code to be more in sync with that.

Though the returned redirect location will be a relative path. For example, on my local server if I create a rewrite like so:

```js
      {
        source: "/external-rewrite",
        destination: "https://nextjs.org/blog/next-10/"
      }
```

It will return a 308 redirect with location of `/blog/next-10` which goes to `http://localhost:3000/blog/next-10` which is a 404. Hopefully it's fine for your case where the redirect should point to somewhere valid at the current domain?

ungarida:
You are right, on localhost I did not have the problem I found out when I was setting up the production environment.

[...]

PR #942, opened 2021-03-10
fix(lambda-at-edge): do not follow redirect in fetch during rewrite

dphang:
Next.js doesn't follow redirect during a rewrite, so we should be in sync with it.

Fixes https://github.com/serverless-nextjs/serverless-next.js/issues/929


GiancarlosIO:
👀

Issue #943, opened 2021-03-10
Getting access to underlining resources in CDK construct


iDVB:
I'm still pretty new to CDK so go easy on me. I think there is a way to fundamentally get access to the underlining resources of a construct from an implementing app. Either way, thats what I'm curious if I can do with this next construct. 

I would like to configure things like the buckets `autoDeleteObjects` prop etc. 

If this is a purely CDK related question and not depended on this construct, I apologies, and feel free to flip me to a CDK doc that might explain how to do this? (still googling)

kirkness:
Hey @iDVB, this should be possible using roughly what is outlined [here](https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html). You can always set the cfn properties on the underlying resources if need be. That said, I could do with exposing all the resources as properties on the construct, as you can with the bucket:
```ts
new NextJSLambdaEdge(this, 'ID', {}).bucket
```

If there is justification for allowing users to pass in this config as a prop then it should be an easy one to go and add.

iDVB:
Well, right now, we have a lambda service who's job it is to empty all stack buckets so the stack can be deleted.
But with CDKs `autoDeleteObjects` it's handled by a custom resource as part of the stack. Would love to decom our service in favour of that.

danschultz:
@kirkness our team also has a desire to customize the S3 bucket that's managed by the construct. What do you think of this?

```ts
export interface Props extends StackProps {
  s3Props?: s3.BucketProps;
}
```

If this direction looks good to you, I can submit a PR.

kirkness:
Gotcha - working through a few issues today so should be able to include this change @danschultz. Cheers for raising @iDVB.

PR #955, opened 2021-03-18
fixes: #943 give access to underlying resources and enable custom s3 props

kirkness:
It's quite often the case that you need to access the underlying resources, for example, to update their roles or to read their configuration to pass into another resource. This PR exposes all created resources as properties on the construct. It also allows users to pass in custom S3 configuration.


dphang:
LGTM!

dphang (approved):

PR #992, opened 2021-04-14
Fix SSG fallback with basepath

### LINKED ISSUE #990: 403 AccessDenied from S3 for dynamic SSG pages when basePath is defined
### Describe the bug
When using a base path there is a 403 from S3 when opening a SSG dynamic route that isn't in getStaticPaths paths.

### Actual behavior
Both `fallback: true` and `fallback: false` seem to result in an error.

### Expected behavior
Depending on fallback value it should either return the 404 page or the fallback page.

### Steps to reproduce
1. create-next-app
2. basic serverless.yml with the component
3. basePath /test
4. `pages/[slug].js` with:
    
        export const getStaticPaths = async () => {
          return {
            paths: ['/foo', '/bar'],
            fallback: false,
          }
        }

5. Try to open /test/baz. Works using next dev. Not after deploying.

### Screenshots/Code/Logs
Nothing in logs. Seems to be an S3 error passed through.

### Versions
Reproducible with latest nextjs release. Both 1.18 and current 1.19 alpha of the component.

jvarho:
Fixes #990 

There were two issues: 

1. Page requests happened with `basePath` in `s3Origin.path` while data requests had it in `request.uri`, so `normalizeUri` broke with fallback pages.
2. Since `basePath` starts with `/` all the S3 operations from response handler broke when using basePath.

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -495,10 +495,12 @@ const handleOriginRequest = async ({
         // 1. URI routes to _error.js
         // 2. URI is not unmatched, but it's not in prerendered routes nor is for an SSG fallback, i.e this is an SSR data request, we need to SSR render the JSON
         break S3Check;
+      } else {
+        // Otherwise, this is an SSG data request, so continue to get to try to get the JSON from S3.
+        // For fallback SSG, this will fail the first time but the origin response handler will render and store in S3.
+        s3Origin.path = basePath;
+        request.uri = uri;
       }
-
-      // Otherwise, this is an SSG data request, so continue to get to try to get the JSON from S3.
-      // For fallback SSG, this will fail the first time but the origin response handler will render and store in S3.
     }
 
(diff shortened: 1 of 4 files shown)


dphang:
Thanks, I thought these cases covered in the e2e tests but maybe the fallback tests are not working correctly, might be worth trying to take a look at those if you have time too.

Let me run the e2e tests and see if something needs fixing there.

jvarho:
I don't think any of the e2e tests combine base path and dynamic paths.

I was going to look into implementing `fallback: blocking` when I have the time, but it might be worthwhile to write some more tests and refactor the handler first.

dphang:
> I don't think any of the e2e tests combine base path and dynamic paths.
> 
> I was going to look into implementing `fallback: blocking` when I have the time, but it might be worthwhile to write some more tests and refactor the handler first.

Makes sense, yea we have a separate test package for more dynamic pages tests but that was just without the basepath. Maybe we can just add the test to the next-app-with-basepath package

dphang (approved):

PR #997, opened 2021-04-16
Fallback e2e tests

jvarho:
The first commit tests for the issues fixed in #992 and the second seems to have uncovered another issue with dynamic SSG routes, this time with locales.

I'm not yet sure what's going on exactly. Writing these e2e tests is a bit of a pain, since there doesn't seem to be a good way to run them locally. Almost all check for cache headers that are only present from cloudfront. So could be the test is wrong?

I'll look into fixing this in the next few days unless someone beats me to it.


dphang:
@jvarho thanks for this. Yeah right now the e2e tests are meant to check CloudFront headers (to check whether it is caching things correctly), we could definitely make it easier on local testing such as by skipping checking CloudFront specific stuff (maybe using an env variable to disable it).

dphang:
ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/754749700

jvarho:
A couple of the failing e2e tests were wrong – I hadn't realized only default locale is prerendered by default. However, most of them showed bugs. The 404 fixes in master take care of some failures, but I found several bugs in build and default lambda.

With these fixes I *think* the tests will now pass, unless I broke something else. However, there are still some issues I ran into that don't seem to matter in practice but are "wrong". At least:

* build step locale expansion creates manifest.json routes that won't exist, like prerendered dynamic routes without locale
* fallback routes end up in `pages.ssr.dynamic` and not having them there would simplify lambda logic
* `if (!hasFallback) return response;` in default handler should probably return an error page instead

Further locale related fixes and cleanups would be simpler if support for old manifest versions was dropped.

dphang:
@jvarho thanks so much for this! Yeah I agree that it is a bit strange and especially with different manifest versions for Next.js, the locale code is a bit messy. Plan is definitely to try to clean it up a bit - the idea was to try to support at least the last 2 minor versions of Next.js, so I think after 10.2 we can get rid of it?

As for the fallback behavior, yes I also think it is convoluted since the origin response handler is used to modify the response after it tries to get from S3 (actually leading to slightly weird semantics like "Error from CloudFront" in the response header sometimes). I am thinking we could actually reduce to origin request handler and have it directly request from S3 using the client (though then your response are all limited to 1 MB), instead of extra network hops through CloudFront retrieving from S3. I can look into this part after 1.19 is released.

Anyway, I will run the e2e tests again and hopefully there are no more failures

dphang:
Ran tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/759087474

jvarho:
The 404 tests failed because the changes from master were missing. I rebased this branch so those can be tested if you want.

Also added a unit test fixture for dynamic paths and locales. It tests for current behavior and has some comments on things that could be changed in the future.

Regarding your comments, waiting for after 10.2 to clean up the old code sounds ok.

jvarho:
Trying to handle everything in the request handler could end up looking even more complex. I wonder if it would slow things down to always require the S3 client for static pages. While writing these patches I was actually thinking if it might be worth including some static files (404 and fallback html) in the lambda function to avoid the second S3 roundtrip.

dphang:
> Trying to handle everything in the request handler could end up looking even more complex. I wonder if it would slow things down to always require the S3 client for static pages. While writing these patches I was actually thinking if it might be worth including some static files (404 and fallback html) in the lambda function to avoid the second S3 roundtrip.

Well, the idea was to help perf mostly, and secondly it also makes it hard to adapt for other platforms (regular Lambda, etc.). I do agree that if the origin request handler becomes too large, it might be complex so it should take some thought to refactor. 

For storing pages in Lambda - yeah, this was my thought before but just hadn't had a chance to work on this optimization. Might be good to add in version 1.20. Beyond that, any arbitrary prerendered page can be put into Lambda itself to avoid hitting S3 - though some projects may generate many pages so it should probably be `serverless.yml` config option. We could at least put all non-dynamic pages (404 page, fallback pages, other static pages) by default, and it can be turned off if space saving is an issue.

Anyway I'll run the e2e tests again, thanks!

dphang (approved):

Issue #999, opened 2021-04-16
Enable passing deploy and build.enabled flags as environment variables


giovai:
**Is your feature request related to a problem? Please describe.**
I'd like to separate the build, package and deploy stages in my CI/CD pipeline. Currently, there is no packaging command available (but that's a separate issue).
It's not possible to inject the value of `deploy` and `build.enabled` flags as environment variables as they are always passed as strings.  

**Describe the solution you'd like**
I'd like to change the types of  `ServerlessComponentInputs.deploy` and `BuildOptions.enabled` to a union type `boolean | string`. This will enable passing these flags as environment variables and perform additional checks like `inputs.deploy === "false"`.

**Describe alternatives you've considered**
To work around it, I've created two separate serverless.yml files - one with hardcoded `build: false` and one with `deploy: false`, which is far from ideal.

dphang:
Merged the PR and published in latest alpha.

PR #1000, opened 2021-04-16
Enable passing deploy and build.enabled flags as environment variables (#999)

giovai:
Enable passing deploy and build.enabled flags as environment variables (#999)

### DIFF
--- a/packages/serverless-components/nextjs-component/src/component.ts
+++ b/packages/serverless-components/nextjs-component/src/component.ts
@@ -205,7 +205,9 @@ class NextjsComponent extends Component {
     const buildConfig: BuildOptions = {
       enabled: inputs.build
         ? // @ts-ignore
-          inputs.build !== false && inputs.build.enabled !== false
+          inputs.build !== false && // @ts-ignore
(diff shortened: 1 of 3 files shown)


packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -205,7 +205,9 @@ class NextjsComponent extends Component {
     const buildConfig: BuildOptions = {
       enabled: inputs.build
         ? // @ts-ignore
-          inputs.build !== false && inputs.build.enabled !== false
+          inputs.build !== false && // @ts-ignore
```
dphang, line 208: Guess we should figure out how to not have to use `@ts-ignore`, though not really due to your PR

dphang:
Yeah I think this makes sense since env variables would be parsed as strings, it would make it easier instead of having to have multiple serverless.yml files. Will approve and merge after e2e tests pass: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/755235925

dphang (approved):

Issue #804, opened 2020-11-16
[RFC] SSG with incremental-static-regeneration and revalidate


arelaxend:
Hi,

**Is your feature request related to a problem? Please describe.**
When you setup a page `/[id].ts` for SSG with `fallback: true` and an API route `/api/preview.ts` to enable `preview mode`, when you makes "preview" changes this works pretty well and you can see live changes of the page. 👍 

Still, changes you make that appears correctly in the `preview mode` does not update the `s3 files static-pages/*.html _next/data/*/*.json` accordingly. 

So whenever you display such a page "publicly" (with preview mode disabled), it does not display the latest changes. What it displays is the latest version of the page stored in s3. 

Since currently, the preview mode does not update s3 files, we cannot see the changes with preview mode disabled.

So currently, for a specific page `youdomain/abcde`, the file `static-page/abcde.html` never updates (this is also true for the corresponding json file). When one makes changes in preview mode, the plugin does not update the static-pages `abcde.html` so the change does not appear publicly.

**How to reproduce the issue**
To reproduce this problem, please follow the steps:
1. Render **without preview mode (no cookies)** the page `/abced`. It should create `_next/data/<id>abced.json` and `static-pages/abced.html` files in s3 and render the page according to the RFC. 👍 
2. Modify the content of `/abced` **with preview mode ON (with cookies)** (to do this, use an API route, e.g. `/api/preview`).
3. Render again the page `/abced`, as the result, the content shows the updates make in 2/ 👍 
4. **Remove the `preview mode cookies`**
5. Display again the page `/abced`, since preview mode is OFF this time, the content are not the latest. It displays the result in 1/ instead of displaying the result with modified content in 3/
6. If you look, steps 2. and 3. do not update files `_next/data/<id>abced.json` and `static-pages/abced.html` in s3. It does not invalidate cloudfront cache neither.

**Describe the solution you'd like**
A good solution requires that when you make changes made in `preview mode`, the next time you render without preview mode the page, it shows the changes.

To fix it, we need to make the fallback page never cached `max-age: 0`, so the lambda can return the newly generated page (and cache it) the next time the same route is hit. But we also need to store the html and json files in preview mode for each preview request (RFC STEP 2)
![Capture d’écran 2020-11-16 à 16 14 01](https://user-images.githubusercontent.com/8854658/99270064-cd1c8a00-2826-11eb-8b16-b563f49c6c40.png)

**Describe alternatives you've considered**
[...]

PR #1007, opened 2021-04-19
Fallback: "blocking"

### LINKED ISSUE #833: RFC: Support getStaticPaths `fallback: 'blocking'`

next.js 10 add support fallback blocking.
https://nextjs.org/blog/next-10#blocking-fallback-for-getstaticpaths

This feature is useful for SEO & OGP support.

I want  html response like existing ssg fallback.

**Describe alternatives you've considered**

We are considering whether to support it with SSR.
However, from a performance standpoint, I plan to give up OGP support and use SSG fallback for the time being.

**Additional context**

I'm trying this feature in following repository.
https://github.com/strozw/sample-serverless-nextjs

jvarho:
Implements [fallback: "blocking"](https://nextjs.org/docs/basic-features/data-fetching#fallback-blocking) and tests for it.

Fixes #833 

Actual change in default-handler was pretty minimal now that the rest of the SSG stuff works correctly.

### DIFF
--- a/packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
+++ b/packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
@@ -136,12 +136,54 @@ describe("Pages Tests", () => {
     });
   });
 
+  describe("Dynamic SSG fallback blocking", () => {
+    [
+      { path: "/fallback-blocking/a" },
+      { path: "/en/fallback-blocking/a" },
+      { path: "/en/fallback-blocking/c" },
+      { path: "/fr/fallback-blocking/a" }
+    ].forEach(({ path }) => {
+      it(`serves page ${path} with correct content immediately`, () => {
+        cy.visit(path);
+        cy.location("pathname").should("eq", path);
+        cy.contains(`Hello ${path.slice(-1)}`);
+      });
+    });
+  });
+
   describe("Dynamic SSG fallback", () => {
     [
       { path: "/fallback/c" },
       { path: "/en/fallback/d" },
       { path: "/fr/fallback/e" },
       { path: "/fr/fallback/a" }
+    ].forEach(({ path }) => {
+      it(`serves page ${path} with fallback at first`, () => {
+        cy.visit(path);
+        cy.location("pathname").should("eq", path);
+        cy.contains("Hello fallback");
+      });
+    });
+
+    [
+      { path: "/fallback/c" },
+      { path: "/en/fallback/d" },
+      { path: "/fr/fallback/e" },
+      { path: "/fr/fallback/a" }
+    ].forEach(({ path }) => {
+      it(`serves page ${path} with correct content soon`, () => {
+        cy.visit(path);
+        cy.location("pathname").should("eq", path);
+        cy.contains(`Hello ${path.slice(-1)}`);
+      });
+    });
+  });
+
+  describe("Dynamic SSG fallback blocking", () => {
+    [
+      { path: "/fallback/c" },
+      { path: "/en/fallback/d" },
+      { path: "/fr/fallback/e" }
     ].forEach(({ path }) => {
       it(`serves page ${path} with fallback at first`, () => {
         cy.visit(path);
--- a/packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
+++ b/packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
@@ -179,33 +179,6 @@ describe("Pages Tests", () => {
     });
   });
 
-  describe("Dynamic SSG fallback blocking", () => {
-    [
-      { path: "/fallback/c" },
-      { path: "/en/fallback/d" },
-      { path: "/fr/fallback/e" }
-    ].forEach(({ path }) => {
-      it(`serves page ${path} with fallback at first`, () => {
-        cy.visit(path);
-        cy.location("pathname").should("eq", path);
-        cy.contains("Hello fallback");
-      });
-    });
-
-    [
-      { path: "/fallback/c" },
-      { path: "/en/fallback/d" },
-      { path: "/fr/fallback/e" },
-      { path: "/fr/fallback/a" }
-    ].forEach(({ path }) => {
-      it(`serves page ${path} with correct content soon`, () => {
-        cy.visit(path);
-        cy.location("pathname").should("eq", path);
-        cy.contains(`Hello ${path.slice(-1)}`);
-      });
-    });
-  });
-
   describe("Dynamic SSG no fallback", () => {
     [
       { path: "/no-fallback/c" },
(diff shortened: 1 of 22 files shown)


adamelmore:
@jvarho just calling this out for visibility: if you wanted to take on #804, I'm offering a $2000 USD bounty for a PR that adds ISR (revalidate) to this repo.

jvarho:
@adamelmore I'd also really like to see ISR – I'm currently using fallback + a separate lambda that deletes pages from S3 periodically to get a poor man's version of ISR.

I haven't been able to come up with a way to implement ISR without stepping outside lambda@edge, so I've been assuming CDK support and some core rework is needed first.

adamelmore:
> @adamelmore I'd also really like to see ISR – I'm currently using fallback + a separate lambda that deletes pages from S3 periodically to get a poor man's version of ISR.
> 
> 
> 
> I haven't been able to come up with a way to implement ISR without stepping outside lambda@edge, so I've been assuming CDK support and some core rework is needed first.

I think the consensus was to submit to an SQS queue from the Lambda@Edge function, and have a lambda triggered from that queue. Dedup on the queue would be used to prevent multiple regen operations within the revalidate window.

Noting that this would carry a bit of a latency hit; it would be faster to emit an SNS mitigation from Lambda@Edge to identically named topics in every region, with an SQS queue subscribed to each of those topics. The queue would handle dedup, and the function would still be triggered off of the queue. This would be much faster from the Lambda@Edge side (non-blocking), but also more complex.

Any of that seem clear to you?

jvarho:
> I think the consensus was to submit to an SQS queue from the Lambda@Edge function, and have a lambda triggered from that queue. Dedup on the queue would be used to prevent multiple regen operations within the revalidate window.

That's more or less along the lines I was thinking. However, it would only support revalidate values larger than the 5-minute minimum deduplication window for SQS queues, right?

> Noting that this would carry a bit of a latency hit; it would be faster to emit an SNS mitigation from Lambda@Edge to identically named topics in every region, with an SQS queue subscribed to each of those topics. The queue would handle dedup, and the function would still be triggered off of the queue. This would be much faster from the Lambda@Edge side (non-blocking), but also more complex.

I don't really see the advantage. Whatever the edge lambda does, whether it's sending a message to a region-local SNS or a global SQS, it does not need to wait for confirmation, since at worst the message is lost and the next request sends a new one. The semantics of revalidation are (AFAIU) that rendering happens "at most" once in the revalidate window.

jvarho:
Ah, sorry, I'm an idiot. The dedup could be based on metadata from S3, not just the key, so revalidate windows shorter than 5 minutes are certainly possible to support with SQS deduplication.

adamelmore:
> I don't really see the advantage. Whatever the edge lambda does, whether it's sending a message to a region-local SNS or a global SQS, it does not need to wait for confirmation, since at worst the message is lost and the next request sends a new one. The semantics of revalidation are (AFAIU) that rendering happens "at most" once in the revalidate window.

You could be right here. I had thought it would be more latency to put a message on an SQS queue (living in a specific region) from any arbitrary edge location, even ignoring any ack back from SQS, but I've not tested this.

packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
```diff
@@ -164,6 +179,33 @@ describe("Pages Tests", () => {
     });
   });
 
+  describe("Dynamic SSG fallback blocking", () => {
```
dphang, line 182: I think you meant "Dynamic SSG fallback"?

dphang:
Ran e2e tests: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/768060840

packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
jvarho: Not only that, but it duplicates (some of) the test above. It's a rebase artifact, fixed now.

jvarho: And since they duplicated the "not cached" test, they failed the e2e test run.

dphang:
ran e2e tests again: https://github.com/serverless-nextjs/serverless-next.js/runs/2393671265?check_suite_focus=true. Looks like just one test had failed, can take a look later unless you have time to look at it

jvarho:
> ran e2e tests again: https://github.com/serverless-nextjs/serverless-next.js/runs/2393671265?check_suite_focus=true. Looks like just one test had failed, can take a look later unless you have time to look at it

I was able to reproduce. The request sometimes returns a cloudfront error because it has headers:

    date: [object Object]
    server: [object Object]

The actual response code and content are correct (the test successfully checks them), but cloudfront understandably doesn't like those headers. It's probably already an issue before this pull request, but retries can get around it if the test case returns the object directly from S3 and then cloudfront has a chance to cache it.

I think the bug is due to lambda-at-edge-compat rewriting the headers, but I haven't tried to understand that code yet.

dphang:
> > ran e2e tests again: https://github.com/serverless-nextjs/serverless-next.js/runs/2393671265?check_suite_focus=true. Looks like just one test had failed, can take a look later unless you have time to look at it
> 
> I was able to reproduce. The request sometimes returns a cloudfront error because it has headers:
> 
> ```
> date: [object Object]
> server: [object Object]
> ```
> 
> The actual response code and content are correct (the test successfully checks them), but cloudfront understandably doesn't like those headers. It's probably already an issue before this pull request, but retries can get around it if the test case returns the object directly from S3 and then cloudfront has a chance to cache it.
> 
> I think the bug is due to lambda-at-edge-compat rewriting the headers, but I haven't tried to understand that code yet.

Hm, I didn't see this fail before in previous tests though...this one also failed in the previous e2e test run. So it might be some be related to your change? I know there's some existing strangeness in origin response, since we modifying whatever comes from s3 (the origin), sometimes I also saw "x-cache: Error from cloudfront" even if a proper page got returned

jvarho:
> Hm, I didn't see this fail before in previous tests though...this one also failed in the previous e2e test run. So it might be some be related to your change? I know there's some existing strangeness in origin response, since we modifying whatever comes from s3 (the origin), sometimes I also saw "x-cache: Error from cloudfront" even if a proper page got returned

I added a failing unit test and a fix for it. Looks to be already introduced in #544.

jvarho:
Actually, looking at the headers brings up another issue. The initial SSG fallback data request (or page with blocking) had no cache control header so it took an extra request or two for cloudfront to cache it.

Also shows up in the most recent e2e run as an extra retry for "serves and caches page /optional-catch-all-ssg-with-fallback/not-found" and likewise for /not-prerendered.

dphang:
> Actually, looking at the headers brings up another issue. The initial SSG fallback data request (or page with blocking) had no cache control header so it took an extra request or two for cloudfront to cache it.
> 
> Also shows up in the most recent e2e run as an extra retry for "serves and caches page /optional-catch-all-ssg-with-fallback/not-found" and likewise for /not-prerendered.

Nice, thanks for a great find! I never realized that missing cache-control headers were why it was had the x-cache error at origin response, thought it was just CloudFront being flaky. Let me run the tests once again

dphang (approved):

PR #1009, opened 2021-04-19
feat: adding support for multiple domainNames

Ruivalim:
The problem: 
Not having possibility to use multiple domainNames in the app.

Case:
We are developing an app using this lib, and we want the possibility to have `www.` in our url.

Temporary solution: 
To solve this problem now, we manually updated cloudfront to accept the same url with `www.` on it and then created the route53 record for that.

Final solution: 
The lib having a way to do that by it self, so I changed the `domainName` prop to `domainNames` which is an array of strings ( Cloud front accept that ) and looped this array to create all the route53 records for the given domain list.


dphang (approved):

dphang:
LGTM!

mavdi:
Awesome, thanks!

PR #1018, opened 2021-04-23
Test for and fix external trailing hash redirects

### LINKED ISSUE #1012: Open Redirect vulnerability on trailingSlash redirect
### Describe the bug
example.com//otherdomain.com will redirect with header //otherdomain.com/ when redirecting from without trailing slash to with.(Or vice versa with trailingSlash false.) Which opens up for open redirect exploits.

### Actual behavior
example.com//otherdomain.com will redirect with header //otherdomain.com/

### Expected behavior
example.com//otherdomain.com will redirect with header /otherdomain.com/ or absolute path.

### Steps to reproduce
(Found this was reproducible on this domain posted in other issue)
https://d3vk8t6m4jw2vm.cloudfront.net//otherdomain.com/

### Screenshots/Code/Logs
See code here: https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/libs/lambda-at-edge/src/default-handler.ts#L342-L363

### Versions
- OS/Environment: AWS
- @sls-next/serverless-component version: 1.19.0-alpha.29 and 1.19.0-alpha.45
- Next.js version: 10.0.7 and 10.1.3

### Additional context
Security issue

### Checklist

- [X ] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
- [ X] Please first try using the most recent `latest` or `alpha` `@sls-next/serverless-component` [release version](https://github.com/serverless-nextjs/serverless-next.js/releases), which may have already fixed your issue or implemented the feature you are trying to use. Note that the old `serverless-next.js` component and the `serverless-next.js` plugin are deprecated and no longer maintained.

jvarho:
Fixes #1012

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -338,11 +338,7 @@ const handleOriginRequest = async ({
     if (newUri.endsWith("/")) {
       newUri = newUri.slice(0, -1);
     }
-  } else if (
-    request.uri !== "/" &&
-    request.uri !== "" &&
-    !uri.endsWith("/404")
-  ) {
+  } else if (/^\/[^/]/.test(request.uri) && !uri.endsWith("/404")) {
     // HTML/SSR pages get redirected based on trailingSlash in next.config.js
     // We do not redirect:
     // 1. Unnormalised URI is "/" or "" as this could cause a redirect loop due to browsers appending trailing slash
--- a/packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
+++ b/packages/libs/lambda-at-edge/tests/default-handler/default-handler.test.ts
@@ -199,6 +199,24 @@ describe("Lambda@Edge", () => {
         }
       );
 
+      it.each`
+        path
+        ${"//example.com"}
+        ${"//example.com/"}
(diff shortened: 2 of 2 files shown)


dphang:
Ran e2e tests: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/779056703

dphang:
> I'll update the e2e tests after this to test this scenario, thanks!

Maybe we can also start to use the regex provided by Next.js to do the matching instead, which might handle all sorts of bad redirect issues... reason for doing this was for perf reasons but looks like we do need to start testing for "/" at the start like in this PR. But that can be done later.

dphang (approved):

jvarho:
> Maybe we can also start to use the regex provided by Next.js to do the matching instead, which might handle all sorts of bad redirect issues... reason for doing this was for perf reasons but looks like we do need to start testing for "/" at the start like in this PR. But that can be done later.

It would be cleaner if the trailing slash redirects from Next.js could just be handled in the normal code with other redirect patterns. But the file/notfile distinction in the trailingSlash: true case is messy.

Are you already working on some sort of core refactor? I was thinking that all the routing decisions could be moved into a platform agnostic library, which would only return the matched route as an object.

PR #1021, opened 2021-04-25
RFC S3 client revert

jvarho:
Looking at the generated lambda, the dynamic S3 client imports do not seem to do anything. The code for them gets bundled up and there's just a wrapper promise in place of the `await import()`.

This is a not-for-merging partial revert of f9eef458 that changes it back to the previous approach. I didn't even fix the tests yet.

There's probably a better way to solve this – I've never used rollup so I didn't try any packaging changes – but this more than halved the default-handler size:

Before:

    636K	default-handler.js
    232K	default-handler.min.js

After:

    288K	default-handler.js
    104K	default-handler.min.js

It seems to have shaved some memory and initialization time off the lambda as well, although that was just based on a few poorly controlled test runs. But I've not measured what happens to fallback response times.

### DIFF
--- a/packages/libs/lambda-at-edge/rollup.config.js
+++ b/packages/libs/lambda-at-edge/rollup.config.js
@@ -9,7 +9,8 @@ const LOCAL_EXTERNALS = [
   "./manifest.json",
   "./routes-manifest.json",
   "./prerender-manifest.json",
-  "./images-manifest.json"
+  "./images-manifest.json",
+  "./s3-client.js"
 ];
 const NPM_EXTERNALS = ["aws-lambda", "aws-sdk/clients/s3"];
 
--- a/packages/libs/lambda-at-edge/rollup.config.js
+++ b/packages/libs/lambda-at-edge/rollup.config.js
@@ -47,5 +47,7 @@ export default [
   { filename: "api-handler", minify: false },
   { filename: "api-handler", minify: true },
   { filename: "image-handler", minify: false },
-  { filename: "image-handler", minify: true }
+  { filename: "image-handler", minify: true },
+  { filename: "s3-client", minify: false },
+  { filename: "s3-client", minify: true }
 ].map(generateConfig);
--- a/packages/libs/lambda-at-edge/src/build.ts
+++ b/packages/libs/lambda-at-edge/src/build.ts
@@ -230,7 +230,11 @@ class Builder {
    * @param shouldMinify
    */
   async processAndCopyHandler(
-    handlerType: "api-handler" | "default-handler" | "image-handler",
+    handlerType:
+      | "api-handler"
+      | "default-handler"
+      | "image-handler"
+      | "s3-client",
     destination: string,
     shouldMinify: boolean
   ) {
@@ -296,6 +300,11 @@ class Builder {
       this.processAndCopyHandler(
         "default-handler",
         join(this.outputDir, DEFAULT_LAMBDA_CODE_DIR, "index.js"),
+        !!this.buildOptions.minifyHandlers
+      ),
+      this.processAndCopyHandler(
+        "s3-client",
+        join(this.outputDir, DEFAULT_LAMBDA_CODE_DIR, "s3-client.js"),
         !!this.buildOptions.minifyHandlers
       ),
       this.buildOptions?.handler
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -620,7 +620,7 @@ const handleOriginResponse = async ({
   const bucketName = domainName.replace(`.s3.${region}.amazonaws.com`, "");
 
   // Lazily import only S3Client to reduce init times until actually needed
-  const { S3Client } = await import("@aws-sdk/client-s3/S3Client");
+  const { S3Client } = require("./s3-client.js");
 
   const s3 = new S3Client({
     region: request.origin?.s3?.region,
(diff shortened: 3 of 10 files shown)


dphang:
The reason for the drastic reduction in handler size is because the revert would now use back the default aws-sdk that's built into the lambda, so no code needs to be bundled. However, one problem is that when you import even just the S3 part, it imports all S3 operations, which I found caused poor cold start times (added 100 ms or so)

However, now it's using the aws sdk v3, which is not in Lambda yet, so its code has to be bundled. But all operations are modularized so we end up only needing to import S3 operations we need (get and put). I believe it only added a few ms (around 3 ms) to the cold start times.

Let me double check on the dynamic imports thing if it's working correctly, I believe it was. But either way even if it's not, the cold start should be much better.

jvarho:
> However, now it's using the aws sdk v3, which is not in Lambda yet, so its code has to be bundled. But all operations are modularized so we end up only needing to import S3 operations we need (get and put). I believe it only added a few ms (around 3 ms) to the cold start times.

Yes, I get that it is more efficient to be able to import just the parts used. But moving the old aws sdk import to be as-needed, like I did in this pseudo-revert, avoids the cold start until the fallback paths are hit (which might be never). My admittedly poor testing seemed to indicate a performance boost.

But I agree this is likely not the correct change, like I wrote above.

dphang:
Ok, let me test it out again and confirm.

dphang:
Ok, let me test it out again and confirm.

dphang:
So this was the original issue: https://github.com/serverless-nextjs/serverless-next.js/issues/580.

I believe I saw ~250 ms cold start init time but that was AWS-SDK (built-in to Lambda, i.e v2) being initialized on cold start outside the handler (so every time even if it's not used) and when using the S3 client there, AFAIK, negligible init time (bulk was already initialized in cold start). Then with the AWS-SDK v3, I used the modularized client but it adds extra code to the handler, but is is dynamically imported. I saw during cold starts, init time drop to ~180 ms and then the dynamic imports were ~3 ms, for a total of ~183 ms. I thought cold start was just ~3 ms, but yeah there might be cold start outside of the handler that a user would always pay for even if they don't use s3 client.

I guess yours is slightly different since you moved AWS-SDK v2 to be a dynamic import instead. So I need to measure what the new cold start times are: 1) outside handler time (should be none added here besides other core initialization due to dynamic import) + 2) dynamic import time. Will try to setup a test later this week.

dphang:
In conclusion, yes it does look like the use of AWS-SDK v3 for its modularized s3 client is adding a bit of cold start time to all requests, maybe around 10-15 ms? Though I think it is much better than AWS-SDK v2 which is taking ~330 ms to dynamically import the client (since it gets all S3 operations, not just GET/PUT).

Ideally, AWS Lambda should hopefully make AWS-SDK v3 available in the Lambda environment soon. Meanwhile, I think we can investigate how to get Rollup.js to inline all the S3 GET/PUT code into that line where I dynamically import S3 client.

jvarho:
> In conclusion, yes it does look like the use of AWS-SDK v3 for its modularized s3 client is adding a bit of cold start time to all requests, maybe around 10-15 ms? Though I think it is much better than AWS-SDK v2 which is taking ~330 ms to dynamically import the client (since it gets all S3 operations, not just GET/PUT), albeit this only occurs when fallback path is hit.

Thanks for confirming!

> Meanwhile, I think we can investigate how to get Rollup.js to inline all the S3 GET/PUT code into that line where we dynamically import S3 client.

Yeah, or the S3 client could be bundled in a separate file that is dynamically imported.

dphang:
> > In conclusion, yes it does look like the use of AWS-SDK v3 for its modularized s3 client is adding a bit of cold start time to all requests, maybe around 10-15 ms? Though I think it is much better than AWS-SDK v2 which is taking ~330 ms to dynamically import the client (since it gets all S3 operations, not just GET/PUT), albeit this only occurs when fallback path is hit.
> 
> Thanks for confirming!
> 
> > Meanwhile, I think we can investigate how to get Rollup.js to inline all the S3 GET/PUT code into that line where we dynamically import S3 client.
> 
> Yeah, or the S3 client could be bundled in a separate file that is dynamically imported.

So I tried some things and we could solve it using a Rollup.js config like so:

```js
const generateConfig = (input) => ({
  input: `./src/${input.filename}.ts`,
  output: {
    dir: "./dist/",
    entryFileNames: `${input.filename}${input.minify ? ".min" : ""}.js`,
    format: "cjs",
    manualChunks(id) {
      if (id.includes("aws-sdk")) {
        return "aws-sdk";
      } else {
        return "serverless-next-js";
      }
    }
  },
  plugins: [
    json(),
    commonjs(),
    externals({
      exclude: "@sls-next/next-aws-cloudfront"
    }),
    nodeResolve(),
    typescript({
      tsconfig: "tsconfig.bundle.json"
    }),
    input.minify
      ? terser({
          compress: true,
          mangle: true,
          output: { comments: false } // Remove all comments, which is fine as the handler code is not distributed.
        })
      : undefined
  ],
  external: [...NPM_EXTERNALS, ...LOCAL_EXTERNALS],
  inlineDynamicImports: false
});
```

The magic is basically in `manualChunks` which is allowing us to create a chunk specifically for `aws-sdk` and rest is in `serverless-next-js` chunk. However, the two caveats are:

1. Every chunk is minified regardless of minify option (the minify option in the config seems to only apply to the entry file, not dependent chunks)
2. The entry file is just a small wrapper of ~10 lines, real code will be in generated `serverless-next-js-<hash>.js` file and the entry file will import this. It will be something like this Entry -> Serverless-next.js core code -> AWS-SDK S3 code. However, we want something like this Entry (with core code) -> AWS-SDK S3 code, so we don't need an extra file lookup for no reason.

It might just be a matter of playing around with Rollup.js config to get to where we want.

jvarho:
@dphang Again, no experience with rollup... This produces more or less the desired result, I think. But maybe too ugly and may not resolve types correctly?

dphang:
> @dphang Again, no experience with rollup... This produces more or less the desired result, I think. But maybe too ugly and may not resolve types correctly?

Yeah no worries, I was trying to check if we can do it in rollup but it seems it's still doing some stuff outside the handler... maybe I'll create an issue in Rollup to ask how to do it correctly as I couldn't figure it out.

I will check your solution and see how it performs, I think it seems ok.

jvarho:
Obsolete

PR #1022, opened 2021-04-25
Images lambda is not needed with non-default loader

jvarho:
With a non-default loader declaration there's no way to use the image lambda from image components, so no need to generate it. 

Would be nice to not need it if `next/image` isn't used, but it doesn't seem like there's a way to tell from the build artifacts.


dphang:
Thanks, this makes sense.

Yeah, I was also trying to check if Next/image is not used then we don't need the image lambda, but like you said, there was no non-hacky way to tell.

dphang (approved):

PR #1024, opened 2021-04-27
feat(sls-next/cdk-construct): edgeLambdas concatenated with default edge lambdas

### LINKED ISSUE #1014: @sls-next/cdk-construct - allow additional edgeLambdas to be set without overriding the default ones.
I'd like to be able to add Lambda@Edge functions to the default behaviour for the Cloudfront Distribution generated as part of `@sls-next/cdk-construct` without overriding the default Lambda@Edge functions. 

Rather than spreading over all props here https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-components/nextjs-cdk-construct/src/index.ts#L250 I'd like the supplied `edgeLambdas` to be concatenated. I don't see a use case for always overriding the default lambdas after all isn't that why this `Construct` exists?

**Describe alternatives you've considered**
Using the `Construct` and adding the behaviour with `addBehaviour` however this is not allowed for default behaviours.

afenton90:
Rather than spreading over all props here https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-components/nextjs-cdk-construct/src/index.ts#L250 `edgeLambdas` are concatenated to the default ones created in the construct.

closes #1014

### DIFF
--- a/packages/serverless-components/nextjs-cdk-construct/__tests__/construct.test.ts
+++ b/packages/serverless-components/nextjs-cdk-construct/__tests__/construct.test.ts
@@ -3,9 +3,10 @@ import { SynthUtils } from "@aws-cdk/assert";
 import { Stack } from "@aws-cdk/core";
 import path from "path";
 import { NextJSLambdaEdge } from "../src";
-import { Runtime } from "@aws-cdk/aws-lambda";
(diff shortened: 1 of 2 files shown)


kirkness (approved):
This looks good to me, nice on @afenton90!

dphang (approved):

PR #1025, opened 2021-04-28
fix(nextjs-cdk-construct): use posix paths for s3 assets and invalida…

dphang:
…tions


kirkness (approved):
Nice one @dphang - unable to test this myself

PR #1026, opened 2021-04-28
Fix router path in fallback SSG

jvarho:
Fixes a bug in #1007 – with fallback: blocking the page render function gets passed the S3 uri with .html which makes the backend router disagree with the front-end about the page path.

This complexity results from the split between request and response lambdas. If they are combined the request object can have the correct uri throughout.

### DIFF
--- a/packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
+++ b/packages/e2e-tests/next-app-with-locales/cypress/integration/pages.test.ts
@@ -147,6 +147,12 @@ describe("Pages Tests", () => {
         cy.visit(path);
         cy.location("pathname").should("eq", path);
         cy.contains(`Hello ${path.slice(-1)}`);
+        cy.request({ url: path }).then((response) => {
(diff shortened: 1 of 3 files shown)


dphang:
Thanks for this! BTW, I noticed you have been contributing a lot lately - let me know if you want to be added as a collaborator so you can create a branch on this and have e2e tests run automatically. Also possibly share some of the funds from https://opencollective.com/serverless-nextjs-plugin as a thanks (will also check other major contributors during the past months)

dphang:
e2e tests LGTM: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/794053818

jvarho:
> Thanks for this! BTW, I noticed you have been contributing a lot lately - let me know if you want to be added as a collaborator so you can create a branch on this and have e2e tests run automatically.

Thanks, I'm afraid my contributions may dry up once the features I want to use are done and the bugs I've run into fixed. But if it means less hassle for you, sure, you could add me. My email is in the comment if needed.

Issue #995, opened 2021-04-15
ISR(Incremental Static Regeneration) feature not generating dynamic pages in s3 bucket


techexe-code:
pages are generated in s3 on this folder[static-pages/{**}.html and _next/data/{id}/*.json] build time only 

when I open url like page/3 in my browser it's always show me loading state then comes up main content and this page is not being generated in s3 
 

```javascript
function Post({ post }) {
  const router = useRouter()
  if (router.isFallback) {
    return <div>Loading...</div>
  }

  // Render post...
}

// This function gets called at build time
export async function getStaticPaths() {
  return {
    // Only `/posts/1` and `/posts/2` are generated at build time
    paths: [{ params: { id: '1' } }, { params: { id: '2' } }],
    // Enable statically generating additional pages
    // For example: `/posts/3`
    fallback: true,
  }
}

// This also gets called at build time
export async function getStaticProps({ params }) {
  const res = await fetch(`https://.../posts/${params.id}`)
  const post = await res.json()

  return {
    props: { post },
    revalidate: 5,
  }
}

export default Post
```

dphang:
We do have e2e test for fallback page here: https://github.com/serverless-nextjs/serverless-next.js/blob/2b9dfd8f36e751de62d3d009d740c06d5fc975b0/packages/e2e-tests/next-app-dynamic-routes/pages/optional-catch-all-ssg-with-fallback/%5B%5B...catch%5D%5D.tsx#L38-L38

`/a` and `/b` are generated at build time, rest are first with fallback page but then data request should generate and save the page/data in S3.

You can try here with some random route: https://d2aheyeo29vn1i.cloudfront.net/optional-catch-all-ssg-with-fallback/<random string> and then break the cache with adding random parameters

I am wondering if you are maybe caching the fallback page causing this behavior? What is your min cache TTL? If you set some min cache in CloudFront, it will force CloudFront to cache that irrespective of cache control header (for fallback it is `public, max-age=0, s-maxage=0, must-revalidate`)

techexe-code:
```
Cache behaviour of CloudFront

path pattern : Default (*)
Minimum TTL : 0
Maximum TTL : 31536000
Default TTL : 0

path pattern : _next/data/*
Minimum TTL : 0
Maximum TTL : 31536000
Default TTL : 0

path pattern : _next/static/*
Minimum TTL : 0
Maximum TTL : 31536000
Default TTL : 86400

path pattern : static/*
[...]

PR #1028, opened 2021-04-28
Incremental static regeneration

kirkness:
Adds [Incremental Static Regeneration](https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration)!

Related issues #804 #995

- [x] Add unit tests for regeneration logic
- [x] Test assumptions around how rollup.js bundles dynamic imports as per @dphang's [comment](https://github.com/serverless-nextjs/serverless-next.js/pull/1028#discussion_r622548607)
- [x] Add e2e tests for ISR fallbacks
- [x] Setup deployment of infra via `serverless`
- [x] Add e2e tests
- [x] Make code for ISR possible to be agnostic to the provider as per @dphang's [comments](https://github.com/serverless-nextjs/serverless-next.js/pull/1028#pullrequestreview-647630746)
- [x] Only pass what's needed to the queue as [per comment](https://github.com/serverless-nextjs/serverless-next.js/pull/1028#discussion_r624517308) from @jvarho 
- [x] Find a workaround for max 300 SQS calls per second (using lambda async invoke)
- [x] Setup deployment of infra via `CDK`
- [x] Create a demo site for reviewers - https://d3kjdl6gtxbc5l.cloudfront.net/about
- [x] Update cache logic to handle changes to `revalidate` property, as per @dvarho's [comment](https://github.com/serverless-nextjs/serverless-next.js/pull/1028#issuecomment-828900860)
- [x] Load test the SQS queue to more than 300 requests in one seconds to [test this assumption](https://github.com/serverless-nextjs/serverless-next.js/pull/1028#issuecomment-828793170) from @evangow - assumption was correct, it's max 300 API calls per second, will find a workaround
- [x] Tidy-up code by moving duplicate code (for example the page generation code) into functions/seperate files

The implementation looks like:
- At "origin response" we check whether the response page is an ISR page (has an `Expires` header, or the original build page had a revalidate property)
- We check whether the `Expires` header is in the future and if so apply a cache-control header such that the object is stored in the edge until its ready for revalidation.
- If the Expires header is in the past then we trigger the page to be regenerated, whilst still returning the existing page with no-cache
- To trigger a page rebuild we send a message to a FIFO SQS Queue, deduplicated using the identifier of the S3 object we are replacing so that we at most trigger one rebuild per page. FIFO SQS queues do come with a 300/second rate limit on all API Requests, which might be hit on very active sites, in which case we cache the page for an extra second to back-off requests.

You can check out a demo of this feature [here](https://d3kjdl6gtxbc5l.cloudfront.net/about), check the headers and the response times to try to get your head around its behaviour. The code for this page looks like:
```tsx
export async function getStaticProps() {
  return {
    revalidate: 60,
    props: {
      date: new Date().toLocaleString()
    }
  };
}

const AboutPage = (props: {date: string}) => (
    // ...
    <p>This is the about page, and the time is about {props.date}</p>
    // ...
)

export default AboutPage
```

👍


packages/libs/lambda-at-edge/src/regeneration-handler.ts
```diff
@@ -0,0 +1,86 @@
...
+        manifestString
+      );
+
+      const s3 = new S3Client({
+        region: cloudFrontEventRequest.origin?.s3?.region,
+        maxAttempts: 3,
+        retryStrategy: await buildS3RetryStrategy()
+      });
+
+      const { req, res } = lambdaAtEdgeCompat(
```
dphang, line 38: This looks to be just regular Lambda, so this should not be needed, right? You can just create a Node req, res in that case?

kirkness: Correct! Good catch.

packages/libs/lambda-at-edge/src/regeneration-handler.ts
```diff
@@ -0,0 +1,86 @@
...
+        );
+      }
+      const cloudFrontEventRequest: AWSLambda.CloudFrontRequest = JSON.parse(
+        cloudFrontEventRequestString
+      );
+      const manifest: OriginRequestDefaultHandlerManifest = JSON.parse(
+        manifestString
+      );
+
+      const s3 = new S3Client({
```
dphang, line 32: I know we are just on lambda-at-edge not but since this is new, if possible maybe try to separate the AWS/CloudFront specific code from the regeneration code. At least maybe just separating via different files for now, and Rollup.js can do the right bundling as needed? That way it is easier to adapt to other platforms if needed in the future (that is kind of the long term vision).

For example you can have the following:
* Core regeneration code (generic)
* Lambda wrapper (hooks up SQS event data -> core code)
* Storage client (in core code, it is abstract interface and you can write an implementation in S3 that can be used if on AWS)

packages/libs/lambda-at-edge/package.json
```diff
@@ -62,6 +62,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-s3": "1.0.0-rc.3",
+    "@aws-sdk/client-sqs": "1.0.0-rc.3",
```
dphang, line 65: I believe https://www.npmjs.com/package/@aws-sdk/client-s3/v/3.13.1 is the latest for both, maybe update this and client-s3 as well (I still need to get the dependency updates working...)

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -602,12 +602,95 @@ const handleOriginResponse = async ({
...
+              ? "public, max-age=0, s-maxage=0, must-revalidate"
+              : `public, max-age=0, s-maxage=${timeToRevalidate}, must-revalidate`
+        }
+      ];
+
+      if (timeToRevalidate < 0) {
+        const { SQSClient, SendMessageCommand } = await import(
+          "@aws-sdk/client-sqs"
+        );
+        const sqs = new SQSClient({
```
dphang, line 647: Maybe let's try to start to abstract this into a non-platform-specific interface so it's more adaptable for other platforms (e.g non AWS), maybe create an interface such as `triggerRegeneration(message: RegenerationMessage)` and right now we can just have one implementation which is the SQS implementation.

Anyway rollup will bundle things into one file.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -602,12 +602,95 @@ const handleOriginResponse = async ({
...
+          key: "Cache-Control",
+          value:
+            timeToRevalidate < 0
+              ? "public, max-age=0, s-maxage=0, must-revalidate"
+              : `public, max-age=0, s-maxage=${timeToRevalidate}, must-revalidate`
+        }
+      ];
+
+      if (timeToRevalidate < 0) {
+        const { SQSClient, SendMessageCommand } = await import(
```
dphang, line 644: As per here: https://github.com/serverless-nextjs/serverless-next.js/pull/1021 actually I don't think Rollup is bundling dynamic imports efficiently (might be a problem with Rollup or maybe we are not using it correctly)? Could you check and maybe adopt the same strategy as the latest revision of this PR and see if the generated code is a true dynamic import

evangow:
I didn't see any handler for rate limit errors on SQS queue in your code.

Thought about ISR quite a bit before. See my comment here: https://github.com/serverless-nextjs/serverless-next.js/issues/804#issuecomment-782565190

After I posted the comment above, I spent a while thinking about the edge cases I listed and the SQS queue.

The downside of using SQS to deduplicate requests is that you have to use a FIFO queue, which only supports 300 messages per second (300 send, receive, or delete operations per second) without batching.

The example I had in my head when thinking about this issue was Reddit. They could benefit a great deal from ISR on /r/popular (as an example), because the content can be stale for 60 seconds and it wouldn't really matter.

The second that the page expires though, there would be 10,000s (100Ks? 1Ms?) of requests to regenerate /r/popular.

The SQS queue would throw a rate limit error on any of those requests past the first 300 and all of the other pages requesting revalidation would also run into a rate limit error.

To handle the rate limit, you could have all initial requests go to a non-FIFO SQS queue, then have that queue send the messages to the FIFO queue.

You would also be able to batch the requests from initial queue and send them over to the FIFO SQS queue. By batching the requests, you can process 3,000 requests / second (300 message limit * 10 messages / batch).

If the FIFO queue returns a rate limit error, then you can just leave those in the initial queue and retry later.

In other words:
Queue 1) Non-FIFO. Receives all regeneration requests. Batches them 10 at a time to send to Queue 2. Retries if Queue 2 returns a rate limit error.
Queue 2) FIFO. Deduplicates messages received from Queue 1 and calls the page regeneration handler.

Ultimately, I think adding a FIFO queue makes things too complicated for deduplicating requests due to the rate limit.

I would just regenerate the page via an async lambda, and if it gets regenerate a whole bunch of times because a page is popular, you're just spending some additional money to handle those requests.

Those pages are probably currently using SSR with right now anyway where every request already currently regenerates the page, so ISR is already going to be a money saver.

If you just use an async lambda, you'll save a little bit less money but you can avoid the complexity that SQS would involve to do it The Right Way.

I could probably be convinced either way is good, but I think just 1 FIFO queue might be an issue.

Unrelated: `s-maxage = initialRevalidateSeconds - createdAgo` Isn't s-maxage supposed to be how long the page should be cached before revalidation? In other words: `s-maxage = Date.now() + initialRevalidateSeconds` or something? I'm not super familiar with the various expiration headers, so ignore me if I'm just wrong here.

packages/libs/lambda-at-edge/package.json
kirkness: 👍

packages/libs/lambda-at-edge/src/regeneration-handler.ts
kirkness: Makes a lot of sense.

packages/libs/lambda-at-edge/src/default-handler.ts
kirkness: 👍

packages/libs/lambda-at-edge/src/default-handler.ts
kirkness: Cool - I'll have a play.

kirkness:
Hey @evangow, cheers for the message. I've not investigated this in-depth yet, however, have implemented the queue with fifo + message deduplication so that the receiving lambda will only be invoked at most once for every time a page needs to be regenerated. What I don't know is whether the queue's rate limit is depleted even when the message is discarded when its a duplicate?

evangow:
I didn't see anything in the docs about that.

If it's the case that only the 1st instance counts against the API usage limits, then that would be awesome.

I assumed the docs to mean that any sqs.send() call would count against your usage even if the message is deduplicated.

jvarho:
Nice :+1:

I also started working on this yesterday, but your efforts are clearly much further along.

However, I don't think your cache-control logic in default-handler will work. It isn't sufficient to just check the prerender manifest for revalidate times, since 1) dynamic pages won't be there and 2) revalidate is returned from `getStaticProps` and can differ from one render to the next. Instead, I think you need to store and check when the page will actually expire.

Here's my series up to the point where actual revalidation logic would be needed: https://github.com/serverless-nextjs/serverless-next.js/compare/master...jvarho:isr-wip
Feel free to use (or not) as you like.

kirkness:
> Nice 👍
> 
> I also started working on this yesterday, but your efforts are clearly much further along.
> 
> However, I don't think your cache-control logic in default-handler will work. It isn't sufficient to just check the prerender manifest for revalidate times, since 1) dynamic pages won't be there and 2) revalidate is returned from `getStaticProps` and can differ from one render to the next. Instead, I think you need to store and check when the page will actually expire.
> 
> Here's my series up to the point where actual revalidation logic would be needed: [master...jvarho:isr-wip](https://github.com/serverless-nextjs/serverless-next.js/compare/master...jvarho:isr-wip)
> Feel free to use (or not) as you like. (Edit: fixed an issue in the topmost commit's tests.)

Thanks, @jvarho! I hadn't considered these cases and I think your approach of using the `Expires` header (based on what I can see in the AWS docs) would work a treat! I'll give it a go implementing and see if I can push a demo of it working (will add tests eventually as well).

jvarho:
> Thanks, @jvarho! I hadn't considered these cases and I think your approach of using the `Expires` header (based on what I can see in the AWS docs) would work a treat! I'll give it a go implementing and see if I can push a demo of it working (will add tests eventually as well).

The way I used it, the Expires header is only used to store the time in S3. It is the Cache-Control I set here that actually matters for Cloudfront: https://github.com/serverless-nextjs/serverless-next.js/compare/master...jvarho:isr-wip#diff-f4709477772e5badfaaf543f6104de4b4ab3bf8300a056c088f84c184d800288R828

The Expires header could – and probably should – be removed at that point.

(The reason to use Cache-Control is of course the distinction between Cloudfront and browser caching.)

kirkness:
> > Thanks, @jvarho! I hadn't considered these cases and I think your approach of using the `Expires` header (based on what I can see in the AWS docs) would work a treat! I'll give it a go implementing and see if I can push a demo of it working (will add tests eventually as well).
> 
> The way I used it, the Expires header is only used to store the time in S3. It is the Cache-Control I set here that actually matters for Cloudfront: [master...jvarho:isr-wipdiff-f4709477772e5badfaaf543f6104de4b4ab3bf8300a056c088f84c184d800288R828](https://github.com/serverless-nextjs/serverless-next.js/compare/master...jvarho:isr-wipdiff-f4709477772e5badfaaf543f6104de4b4ab3bf8300a056c088f84c184d800288R828)
> 
> The Expires header could – and probably should – be removed at that point.
> 
> (The reason to use Cache-Control is of course the distinction between Cloudfront and browser caching.)

Ok, so you set the Expires header when creating the file in S3, and then use that to define the smax header at which point in the Origin Response the Expires header can be removed?

jvarho:
> Ok, so you set the Expires header when creating the file in S3, and then use that to define the smax header at which point in the Origin Response the Expires header can be removed?

Exactly :+1:

kirkness:
> > Ok, so you set the Expires header when creating the file in S3, and then use that to define the smax header at which point in the Origin Response the Expires header can be removed?
> 
> Exactly 👍

Added this in 0ee2f55 and fc8bd38.

adamelmore:
tumblr_ljh0puClWT1qfkt17

jvarho:
> > > Ok, so you set the Expires header when creating the file in S3, and then use that to define the smax header at which point in the Origin Response the Expires header can be removed?
> > 
> > 
> > Exactly +1
> 
> Added this in [0ee2f55](https://github.com/serverless-nextjs/serverless-next.js/commit/0ee2f553011ec376353d4865ed570d120a1dd702) and [fc8bd38](https://github.com/serverless-nextjs/serverless-next.js/commit/fc8bd3895d53ea1f6953484ed348ec7033ce407f).

That looks correct for the regeneration path. However, you still need something like my changes in https://github.com/jvarho/serverless-next.js/commit/bd988a42e2769ebbff1f3130680ff759a2835300 and https://github.com/jvarho/serverless-next.js/commit/4423a4716a333b6c509d2db3247ec9bd869ac3c5 to get prerendered and fallback-rendered pages to expire, do you not? (I think you could just cherry-pick those two commits.)

kirkness:
> Like I wrote in elsewhere, dynamic pages do not have `initialRevalidateSeconds` in the manifest. The only way to know that they should be revalidated is to store that information in S3 – i.e. the Expires header you now use.

Update: 8a0d5ff

Frustratingly there is no simple way of setting the `Expires` header per-file when uploading the initial build via CDK (obviously possible via the serverless component). With the CDK construct, we can only set headers for an entire directory. Because of this, I've updated the `default-handler` (logic moved in `getStaticRegenerationResponse`) to either check for an `Expires` header or for an `initialRevalidateSeconds`. This should result in new pages being regenerated when needed and initial pages being regenerated, however, the narrow case you've just mentioned, where you might convert an ISR page to a static page will not be. Unless you can think of another edge-case that may not be fulfilled, just this one feels acceptable until we've figured a method of setting `Expires` per-file in CDK.

packages/libs/lambda-at-edge/src/build.ts
```diff
@@ -414,6 +425,36 @@ class Builder {
...
   }
 
+  async buildRegenerationHandler(
+    buildManifest: OriginRequestDefaultHandlerManifest
+  ): Promise<void> {
+    await Promise.all([
+      this.copyTraces(buildManifest),
+      this.processAndCopyHandler(
+        "regeneration-handler",
+        join(this.outputDir, REGENERATION_LAMBDA_CODE_DIR, "index.js"),
```
jvarho, line 435: Extra "index.js" parameter here, after your merge from master.

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -590,7 +595,6 @@ const handleOriginRequest = async ({
 const handleOriginResponse = async ({
   event,
   manifest,
-  prerenderManifest,
```
jvarho, line 593: If you remove this here, you could also remove it from the call in `handler` above and the type below.

packages/libs/lambda-at-edge/src/lib/getStaticRegenerationResponse.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  const expiresAt = options.expiresHeader
+    ? new Date(options.expiresHeader)
+    : firstRegenerateExpiryDate(
+        options.lastModifiedHeader as string,
+        initialRevalidateSeconds as number
+      );
+
+  // isNaN will resolve true on initial load of this page (as the expiresHeader
+  // won't be set), in which case we trigger a regeneration now
```
jvarho, line 62: This is probably left over from an earlier version? AFAICT `expiresAt` always gets a value from one header or the other.

packages/libs/lambda-at-edge/src/lib/triggerStaticRegeneration.ts
```diff
@@ -0,0 +1,70 @@
...
+          StringValue: bucketName
+        },
+        CloudFrontEventRequest: {
+          DataType: "String",
+          StringValue: JSON.stringify(options.request)
+        },
+        Manifest: {
+          DataType: "String",
+          StringValue: JSON.stringify(options.manifest)
+        },
```
jvarho, line 51: Instead of sending the (same) manifest in every message, you could copy and import it in regeneration lambda as done in default lambda. Might even run into message limit with large manifests?

Also, you could avoid the multiple stringify/parse boilerplate by using MessageBody instead of MessageAttributes. But maybe you have some routing/filtering with attributes in mind?

kirkness:
> I didn't see anything in the docs about that.
> 
> If it's the case that only the 1st instance counts against the API usage limits, then that would be awesome.
> 
> I assumed the docs to mean that any sqs.send() call would count against your usage even if the message is deduplicated.

I've just done some research and I think you are bang-on, its a 300/second limit on any API call to a FIFO queue. I've just pushed this update [here](https://github.com/serverless-nextjs/serverless-next.js/commit/ef7d03b872ffde3d6b3da24b0fa77c31537ce38d) which uses the approach you suggested via async invocation.

packages/e2e-tests/next-app/cypress/integration/redirects.test.ts
```diff
@@ -42,7 +42,7 @@ describe("Redirects Tests", () => {
       {
         path: "//example.com/",
         expectedPath: "/example.com",
-        expectedStatus: 404
+        expectedStatus: 308
```
kirkness, line 45: This was not passing after the update to cypress - the 308 is returned initially and the 404 is the status code of the subsequent request.

packages/e2e-tests/next-app-dynamic-routes/package.json
```diff
@@ -36,7 +36,7 @@
     "aws-sdk": "2.758.0",
     "babel-plugin-istanbul": "6.0.0",
     "babel-plugin-transform-class-properties": "6.24.1",
-    "cypress": "5.1.0",
+    "cypress": "7.2.0",
```
kirkness, line 39: Updated when I noticed `ensureRouteNotCached` and similar intercepting commands weren't behaving as expected (they never threw or failed a test before the update).

dphang: @kirkness hmm yeah I think this was experimental so might be why it wasn't working correctly all the time - hopefully newer versions are more stable. Should also update it in other e2e tests

packages/libs/lambda-at-edge/src/default-handler.ts
kirkness: Assuming this is resolved as per #1029, also seems to have proper split bundles and dynamic imports after pulling in the changes in master.

jvarho:
@kirkness Sorry to bikeshed, but I notice you switched from FIFO SQS to just calling the async lambda directly. I'm not sure about that.

Next.js documentation says that the revalidation is "at most once" per interval and advertises "Low database and backend load. Pages are re-computed at most once concurrently" as an advantage of the feature. Using an async lambda with no deduplication seems to throw that away. (I'm personally interested in a case where a backend api has a monthly request limit.)

IMHO the SQS approach would be a better default. Depending on how the FIFO rate limit works (does it return a 403 ThrottlingException?) it could be possible to (optionally?) fall back to calling the lambda directly if the queue is busy.

packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -474,6 +484,41 @@ class NextjsComponent extends Component {
...
+        "timeout",
+        "regenerationLambda",
+        10
+      ) as number,
+      runtime: readLambdaInputValue(
+        "runtime",
+        "regenerationLambda",
+        "nodejs12.x"
+      ) as string,
+      name: bucketOutputs.name
```
jvarho, line 516: You are using bucket name & region to call the lambda, but only give the name here. Shouldn't you also define region here?

kirkness:
> @kirkness Sorry to bikeshed, but I notice you switched from FIFO SQS to just calling the async lambda directly. I'm not sure about that.
> 
> Next.js documentation says that the revalidation is "at most once" per interval and advertises "Low database and backend load. Pages are re-computed at most once concurrently" as an advantage of the feature. Using an async lambda with no deduplication seems to throw that away. (I'm personally interested in a case where a backend api has a monthly request limit.)
> 
> IMHO the SQS approach would be a better default. Depending on how the FIFO rate limit works (does it return a 403 ThrottlingException?) it could be possible to (optionally?) fall back to calling the lambda directly if the queue is busy.

Ah, this makes sense. It would be ideal to match the vercel approach. That said, for your case note that your backend API would receive a small subset of the amount it was receiving before - unless you experience large surges in requests, or your page rendering is slow. The regeneration will only be called when the cache has expired.

It sounds as if one possible solution (if we feel the rate-limit will cause issues) could be to post all messages to a non-fifo queue (no rate limit), receive messages from that queue in another lambda with up to batches of 10, and then sending the entire batch of up to 10 messages to a fifo queue. This means you could start to see rate-limit errors if you were seeing above 3,000 hits within the time it takes to queue and rerender a page (and also AWS do allow you to up that limit to 30,000 batched messages if you request).

Your recommendation could also be a good shout and would require less infrastructure to be set up which feels kinda cumbersome and complex.

I will do some further research but if you have any thoughts on the above please sing out!

packages/serverless-components/nextjs-component/src/component.ts
kirkness: Im not sure I follow, im naming the function the same as the bucket so that these can be referenced easily (kinda hacky perhaps?), you can see here where I invoke: https://github.com/serverless-nextjs/serverless-next.js/pull/1028/files/30d0e5113b3fa2f6871aebdb70ae93b450d9c31c#diff-3a1cb611298615fb539fd21d0f14dacf7eb5837a67ca0009dedee1418962cc03R43

jvarho: My point is that you are using the S3 region in the lambda client here: https://github.com/serverless-nextjs/serverless-next.js/pull/1028/files#diff-3a1cb611298615fb539fd21d0f14dacf7eb5837a67ca0009dedee1418962cc03R29

So shouldn't you create the lambda in the same region? Rather than the default us-east-1.

kirkness: Gotcha 👍

jvarho:
> It sounds as if one possible solution (if we feel the rate-limit will cause issues) could be to post all messages to a non-fifo queue (no rate limit), receive messages from that queue in another lambda with up to batches of 10, and then sending the entire batch of up to 10 messages to a fifo queue.

Sounds complex for "just" a tenfold increase.

If it's a single page that's a problem, adding some sloppiness to cache expiry would be simple. E.g. return s-max-age=1 for stale pages, so Cloudfront caches them for a bit.

Another possibility would be sharding. E.g. creating N queues and hashing the page + last modified time to pick the queue. That wouldn't help with a single page seeing all those requests, but would if a large site has many ISR pages. Complex, but at least arbitrarily scalable.

kirkness:
I've reverted to using SQS in the most recent couple of commits and added a check for if/when we hit a throttling exception, if we run into one then we return an `s-maxage=1`.

evangow:
I thought @jvarho's suggestion to simply call the async lambda only when the queue is throwing a throttle error made a lot of sense.

It scales without adding nearly as much complexity as having multiple queues would, and it would also handle the case where ISR gets called on a page that hasn't been generated before (so doesn't have a stale page to rely on).

To be honest though, I'm not sure I'm properly following the s-max-age solution being used.

Is the s-max-age used to determine whether the page needs to be regenerated? Thus, by setting it to 1, we're essentially delaying the regeneration until the next time the page is requested?

kirkness:
> I thought @jvarho's suggestion to simply call the async lambda only when the queue is throwing a throttle error made a lot of sense.
> 
> It scales without adding nearly as much complexity as having multiple queues would, and it would also handle the case where ISR gets called on a page that hasn't been generated before (so doesn't have a stale page to rely on).
> 
> To be honest though, I'm not sure I'm properly following the s-max-age solution being used.
> 
> Is the s-max-age used to determine whether the page needs to be regenerated? Thus, by setting it to 1, we're essentially delaying the regeneration until the next time the page is requested?

The reason I opted for this approach was the keep the behavior more in line with vercel (i.e. @jvarho's comment about rendering the page at most once per revalidation period). If a page on your site is highly busy, then you'll be sure that a new regeneration message will be added to the queue. Whilst the page is regenerating, if your page gets requested again, we'd attempt to send another message to the queue which would not be received as its a dupe. If this happens many-many times within a second and we hit a rate limit (which by the way would likely be many more than 300/second as the SDK will retry once after its been rejected), then this approach will set the current page to cache in the CDN edge for one second (this is what the s-maxage directive does). After this second the origin response lambda will be invoked again and we'll reattempt submitting a message to the revalidation queue - which should be successful at this point given we've not submitted a request for at least 1 second.

If we send an async invocation to the lambda, in the event of being hit with many requests, then we are only going to be sending more duplicate requests unnessecarily. The upside of the async invocation approach however is if the rate limit is being hit due to a site having many popular pages all being frequently requested all at once.

kirkness:
@dphang have you any idea what the e2e issue is [here](https://github.com/serverless-nextjs/serverless-next.js/runs/2513440417?check_suite_focus=true) - kind of sounds as if the component has not been built (and output to dist). Are there any known issues related to this, or is it worth me digging further?

jvarho:
> @dphang have you any idea what the e2e issue is [here](https://github.com/serverless-nextjs/serverless-next.js/runs/2513440417?check_suite_focus=true) - kind of sounds as if the component has not been built (and output to dist). Are there any known issues related to this, or is it worth me digging further?

The actual issue is already in the build step. Seems that a failing build does not stop the run. Stale lockfiles at least, not sure if all the errors are due to that.

kirkness:
> The actual issue is already in the build step. Seems that a failing build does not stop the run. Stale lockfiles at least, not sure if all the errors are due to that.

👍, looks like the AWS user used by Actions doesn't have the required access to create the SQS queue. Perhaps @danielcondemarin or @dphang have access to update the user?

mtannerfors:
> > The actual issue is already in the build step. Seems that a failing build does not stop the run. Stale lockfiles at least, not sure if all the errors are due to that.
> 
> 👍, looks like the AWS user used by Actions doesn't have the required access to create the SQS queue. Perhaps @danielcondemarin or @dphang have access to update the user?

Amazing work @kirkness et al.

dphang:
> > The actual issue is already in the build step. Seems that a failing build does not stop the run. Stale lockfiles at least, not sure if all the errors are due to that.
> 
> 
> 
> 👍, looks like the AWS user used by Actions doesn't have the required access to create the SQS queue. Perhaps @danielcondemarin or @dphang have access to update the user?
> 
> 
> 
> 

@danielcondemarin should have access to update the GitHub actions user's permissions

mtannerfors:
> > > The actual issue is already in the build step. Seems that a failing build does not stop the run. Stale lockfiles at least, not sure if all the errors are due to that.
> > 
> > 
> > 👍, looks like the AWS user used by Actions doesn't have the required access to create the SQS queue. Perhaps @danielcondemarin or @dphang have access to update the user?
> 
> @danielcondemarin should have access to update the GitHub actions user's permissions

@danielcondemarin @dphang @kirkness Anything I/we can assist with to help speed up the conclusion of this PR?

kirkness:
> @danielcondemarin @dphang @kirkness Anything I/we can assist with to help speed up the conclusion of this PR?

Cheers @mtannerfors but I think we are just waiting on AWS permissions being updated by @danielcondemarin so that we can run the e2e tests.

dphang:
> > @danielcondemarin @dphang @kirkness Anything I/we can assist with to help speed up the conclusion of this PR?
> 
> Cheers @mtannerfors but I think we are just waiting on AWS permissions being updated by @danielcondemarin so that we can run the e2e tests.

Thanks, @danielcondemarin hasn't been active lately but has given me access to the AWS account for e2e tests. Should be able to update the permissions later today.

packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -446,8 +451,14 @@ class NextjsComponent extends Component {
       return inputValue[lambdaType] || defaultValue;
     };
 
+    const queue = await sqs({
```
dphang, line 454: BTW, I guess ISR is a build time feature, so would you be able to detect if users are using it and only create the SQS queue and regeneration Lambda if so?

dphang:
> > > @danielcondemarin @dphang @kirkness Anything I/we can assist with to help speed up the conclusion of this PR?
> > 
> > 
> > Cheers @mtannerfors but I think we are just waiting on AWS permissions being updated by @danielcondemarin so that we can run the e2e tests.
> 
> Thanks, @danielcondemarin hasn't been active lately but has given me access to the AWS account for e2e tests. Should be able to update the permissions later today.
> 
> EDIT: looks like I still don't have the permissions, will ping again but he's in UK time so probably will have to wait until next morning :)
> 
> Meanwhile there are some conflicts in the lockfile due to dependency upgrades, @kirkness maybe you can update in the meantime

BTW, have added SQS permissions, let me know if anything else is needed.

packages/libs/lambda-at-edge/src/lib/cleanRequestUriForRouter.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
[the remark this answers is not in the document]
jvarho: I've yet to touch the response handler in by refactoring precisely to avoid conflicting with this pull request. Much of the remaining response handler routing logic can be refactored away, I believe. But it will be easier after both this and my currently open PR are merged to master.

packages/serverless-components/nextjs-component/src/component.ts
jvarho: Technically there's no way to know if a dynamic SSG page with fallback true/blocking will use ISR if it does not prerender any static paths that do.

But in practice it should be good enough to only enable regeneration when there are some revalidation times in the prerender manifest. It would be easy to work around.

packages/libs/lambda-at-edge/src/lib/cleanRequestUriForRouter.ts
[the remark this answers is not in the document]
kirkness: Cool, let's update this in a follow-up PR.

kirkness:
> > > > @danielcondemarin @dphang @kirkness Anything I/we can assist with to help speed up the conclusion of this PR?
> > > 
> > > 
> > > Cheers @mtannerfors but I think we are just waiting on AWS permissions being updated by @danielcondemarin so that we can run the e2e tests.
> > 
> > 
> > Thanks, @danielcondemarin hasn't been active lately but has given me access to the AWS account for e2e tests. Should be able to update the permissions later today.
> > EDIT: looks like I still don't have the permissions, will ping again but he's in UK time so probably will have to wait until next morning :)
> > Meanwhile there are some conflicts in the lockfile due to dependency upgrades, @kirkness maybe you can update in the meantime
> 
> BTW, have added SQS permissions, let me know if anything else is needed.
> 
> I also added: "lambda:ListEventSourceMappings" to the permissions and will rerun the tests. So looks like final list is:
> 
> ```
>                 "lambda:ListEventSourceMappings",
>                 "lambda:CreateEventSourceMapping",
>                 "iam:UpdateAssumeRolePolicy",
>                 "iam:DeleteRolePolicy",
>                 "sqs:*"
> ```
> 
> @kirkness looks like there are a couple things:
> 
> 1. Can you update readme with all the new permissions needed for ISR?
> 2. Also I think a lot of e2e tests are failing with the lockfile issue, probably need to update something there
> 3. The ISR tests are failing in the latest e2e test run

Cheers @dphang! I've checked off those first 2 points, although am wondering whether you can see or know of any reason why the 2 pages I've added to `next-app` e2e tests are not deploying successfully. I can see that before the e2e script has cleaned up s3, the page at https://d3dnttyy4tog6k.cloudfront.net/revalidated-ssg-page returns a 503 (along with the default CloudFront permissions error message), although can't do much more without logs and seeing what is making its way to S3, etc. I've deployed several separate instances of the `next-app` to my own account and every time its successful, so I'm wondering if there is something environment/setup specific going on.

dphang:
Sure, let me take a look at the logs and let you know what happened

dphang:
@kirkness updated the policy, I think you need to add sqs.SendMessage permission to the default policy here? https://github.com/serverless-nextjs/serverless-next.js/blob/07b8135b58226a33858eaf4e3a4bc5591009bf0d/packages/serverless-components/nextjs-component/src/component.ts#L450

kirkness:
> @kirkness updated the policy, I think you need to add sqs.SendMessage permission to the default policy here?
> 
> https://github.com/serverless-nextjs/serverless-next.js/blob/07b8135b58226a33858eaf4e3a4bc5591009bf0d/packages/serverless-components/nextjs-component/src/component.ts#L450
> 
> Also 1/3 of the new ISR tests passed, not sure if they are flaky but AFAIK there are no more permissions issues

Sweet, thanks for checking, odd that it worked locally! I'll add this today along with making sure these tests aren't too flakey!

kirkness:
> @kirkness updated the policy, I think you need to add sqs.SendMessage permission to the default policy here?
> 
> https://github.com/serverless-nextjs/serverless-next.js/blob/07b8135b58226a33858eaf4e3a4bc5591009bf0d/packages/serverless-components/nextjs-component/src/component.ts#L450
> 
> Also 1/3 of the new ISR tests passed, not sure if they are flaky but AFAIK there are no more permissions issues

That permission appears to already exist on the default lambda's policy: https://github.com/serverless-nextjs/serverless-next.js/blob/1f757d1b5f27e87a86aa218b3c59303e31de9b0a/packages/serverless-components/nextjs-component/src/component.ts#L478-L482

Perhaps the policy isn't getting updated?

packages/serverless-components/nextjs-component/src/component.ts
kirkness: I've added a check to the serverless components and the cdk construct for the presence of ISR pages, and create a queue based on that. I agree, I think it'd be easy enough to work around in case you do need the queue.

dphang:
> > @kirkness updated the policy, I think you need to add sqs.SendMessage permission to the default policy here?
> 
> > 
> 
> > https://github.com/serverless-nextjs/serverless-next.js/blob/07b8135b58226a33858eaf4e3a4bc5591009bf0d/packages/serverless-components/nextjs-component/src/component.ts#L450
> 
> > 
> 
> > Also 1/3 of the new ISR tests passed, not sure if they are flaky but AFAIK there are no more permissions issues
> 
> 
> 
> That permission appears to already exist on the default lambda's policy: https://github.com/serverless-nextjs/serverless-next.js/blob/1f757d1b5f27e87a86aa218b3c59303e31de9b0a/packages/serverless-components/nextjs-component/src/component.ts#L478-L482
> 
> 
> 
> Perhaps the policy isn't getting updated?

Yup, thanks! I think the policy might not be getting updated after it's already deployed...but I think it should be ok to ask users to update it

kirkness:
@dphang - the e2e's are killing me slowly 😂.... It looks like the issue is caused by the shared libs not reliably being built due to yarn failing to install (which I understand is a known issue). I've gone through each of the tests locally and each app/test suite passes.

I'd be fairly keen to work on this issue in a follow-up PR. My suspicion is that the issue is related to the `postinstall` script is running `yarn install` in parallel (perhaps trying to create the same file/dir at the same time).  So perhaps shifting to using `workspaces` would be beneficial? Has this been looked at before?

dphang:
> @dphang - the e2e's are killing me slowly 😂.... It looks like the issue is caused by the shared libs not reliably being built due to yarn failing to install (which I understand is a known issue). I've gone through each of the tests locally and each app/test suite passes.
> 
> I'd be fairly keen to work on this issue in a follow-up PR. My suspicion is that the issue is related to the `postinstall` script is running `yarn install` in parallel (perhaps trying to create the same file/dir at the same time). So perhaps shifting to using `workspaces` would be beneficial? Has this been looked at before?
> 
> **Edit:** Looks like @jvarho's [PR](https://github.com/serverless-nextjs/serverless-next.js/pull/1079) will help!

Yep hopefully it helps.

I do see that the code limit is reached (75 GB), I'll add a script to clean up old Lambdas which can run in the finalize step

raffij:
Looking at this I learnt about `versionFunctions: false` in the serverless.yml which is useful for our deploys, but possibly not for this instance.

README.md
```diff
@@ -385,6 +385,11 @@ The exhaustive list of AWS actions required for a deployment:
   "s3:PutAccelerateConfiguration",
   "s3:PutBucketPolicy",
   "s3:PutObject"
+  "lambda:ListEventSourceMappings",
+  "lambda:CreateEventSourceMapping",
+  "iam:UpdateAssumeRolePolicy",
+  "iam:DeleteRolePolicy",
+  "sqs:*"
```
dphang, line 392: For better security, we should update this to just the sqs permissions needed, but not blocking for this PR

dphang:
I think so far it looks good to me, really great work on this! If you have any improvements like updating the README on how to use this feature (e.g all the SQS permissions etc needed), improving tests, cleaning up more of the code etc. please feel free to add in future PRs.

dphang (approved):

amman20:
Awesome work guys,
when are we expecting this to be available in the latest stable version?

leerob:
Nice work, everyone! 👏

packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -475,6 +507,58 @@ class NextjsComponent extends Component {
...
+          "timeout",
+          "regenerationLambda",
+          10
+        ) as number,
+        runtime: readLambdaInputValue(
+          "runtime",
+          "regenerationLambda",
+          "nodejs12.x"
+        ) as string,
+        name: bucketOutputs.name
```
b3nThomas, line 551: Should this be trying to extract the value from my config, instead of using the s3 bucket name?:
```
        name:
            regenerationLambda: should-be-used <-----
```

PR #1029, opened 2021-04-29
perf(lambda-at-edge): fix rollup bundling for dynamic imports perf

dphang:
Note: testing rollup config to properly bundle dynamic imports like s3client which is not always needed. The cost is an additional wrapper on top of main code but this seems expected optimization from Rollup.js (preloading and parsing optimization). 

Previously, inlining dynamic imports into one file was still causing that code to be at least partially preloaded outside of the handler, e.g initializing some S3/AWS SDK related objects even if the S3Client itself was only initialized when needed. So it caused higher cold starts due to this for all paths. With this change, we split into multiple chunks. Total there should be a few chunks:

* Entry chunk (wrapper of ~20 lines on top of other code)
* Core chunk (all of the routing code bundled into one file)
* Dynamic chunks (e.g S3 client, GetObjectCommand, PutObjectCommand) which is heavy code only needed on fallback paths

Total size of all chunks remains around the same but slightly higher at 657 kB vs. 651 kB before. However, Lambda does not need to read (and preload some parts of it) ~166 kB of code for S3 unless it's needed. There is one extra file I/O due to the wrapper requiring the core code but from tests, it appears to be negligible as ~25 ms is saved in init time from my tests (195 -> 170 ms with the 512 MB Lambda function)
In the future, we can do further optimizations like getting rid of complexity of having the origin response handler, and even using Rollup again at build/deploy time to further get rid of code paths that are not used (e.g if user has no rewrites, we can get rid of that code)

### DIFF
--- a/packages/libs/lambda-at-edge/package.json
+++ b/packages/libs/lambda-at-edge/package.json
@@ -51,7 +51,7 @@
     "@types/send": "^0.14.5",
     "@types/sharp": "^0.26.1",
     "fetch-mock-jest": "^1.5.1",
-    "rollup": "^2.26.6",
+    "rollup": "^2.46.0",
     "rollup-plugin-node-externals": "^2.2.0",
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-typescript2": "^0.27.2",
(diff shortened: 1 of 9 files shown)


jvarho:
Yes, this looks better than my effort :+1:

I can confirm this reproduces the improvement I saw in cold starts. Here are five-run averages for before  and after this branch. I'm using 128 MB memory setting and a redirect to get the worst case.

Before:

Max Memory Used: 71 MB
Init Duration: 201.8 +/- 8.3 ms
Duration: 30.2 +/- 5.3 ms

After:

Max Memory Used: 67 MB
Init Duration: 161.2 +/- 6.9 ms
Duration: 31.3 +/- 6.7 ms

About 20% lower init duration and a bit less memory use.

I have not tested the fallback time, because it's dominated by S3 latency and rendering.

PR #1031, opened 2021-04-29
fix(sls-next/cdk-construct): defaultBehavior.edgeLambdas only applied to defaultBehavior

afenton90:
Passing `defaultBehavior.edgeLambdas` was being applied to `defaultBehavior` and  to `_next/data/*` behaviour. Could cause unintended side effects.


kirkness (approved):
LGTM!

dphang (approved):

PR #1033, opened 2021-05-01
Core library refactor

jvarho:
As a first step towards supporting apigw/lambda* as a backend, I'm trying to refactor some of the common routing code from under lambda-at-edge.

So far this only moves basic auth and redirects. I stopped here to ask for feedback regarding a few questions I have:

1. Does this make sense? Or is there some other location this code should go or a better way to share it?
2. The main change so far is new type definitions (libs/routing/src/types.ts). The idea is to only define what is needed, so these are compatible with cloudfront, but also other providers with minimal shimming. Is basing these on cloudfront request/response models reasonable or should I use some other model as a base? Or get rid of Request/Response and take more specific values and return more abstract objects that need to be transformed?
3. Test coverage does not seem to include cross-package tests. E.g. default handler tests should cover all the redirect functionality, but that is not reflected in the coverage report after moving the code. Any ideas why or how to solve?

The next steps would be:

* Handle rewrites (slightly less straightforward code moving)
* Handle route resolution (i.e. refactor router functions and path normalization)
* Build manifest generation logic? (or is that meant to be cloudfront-specific?)

---

*Actual apigw/lambda implementation would be a dependent change after this. The main reason I want to use that over lambda@edge is keeping backend traffic within a single region, perhaps even in VPC.


dphang:
Thanks for this! Yeah, I was going to work on this after 1.19 to refactor the core logic out of the Lambda@Edge related code, but looks like you got started first.

I would say CloudFront logic should stay in lambda-at-edge package and the core code should only return a generic response. I believe we should follow what Vercel does which is just using the Node.js req/res models. Then Lambda-at-edge package can turn into just a wrapper on the core code; it will just use the Lambda@Edge compatibility layer to convert Node.js req/res to/from CloudFront requests/responses (https://github.com/serverless-nextjs/serverless-next.js/blob/3a11699f92df41ce9086fbc2f368ae21b2ac7252/packages/compat-layers/lambda-at-edge-compat/next-aws-cloudfront.d.ts)

packages/libs/lambda-at-edge/package.json
```diff
@@ -63,6 +63,7 @@
   "dependencies": {
     "@aws-sdk/client-s3": "1.0.0-rc.3",
     "@hapi/accept": "5.0.1",
+    "@sls-next/routing": "link:../routing",
```
dphang, line 66: Maybe call this @sls-next/core as it might have other logic/utilities besides routing logic

packages/libs/routing/src/auth.ts
```diff
@@ -1,17 +1,17 @@
...
   if (authentication && authentication.username && authentication.password) {
     const validAuth =
       "Basic " +
       Buffer.from(
         authentication.username + ":" + authentication.password
       ).toString("base64");
 
-    if (authorizationHeader !== validAuth) {
+    if (!authorizationHeaders || authorizationHeaders[0]?.value !== validAuth) {
       return {
```
dphang, line 15: For this, this was a quick way to return a response directly from origin request handler in CloudFront. But as I mentioned, it might be better returned as Node.js req/res instead to be more standard and converted with the lambda@edge compatibility code (converting Node.js req/res -> CloudFront req/res)

dphang:
> 3\. Test coverage does not seem to include cross-package tests. E.g. default handler tests should cover all the redirect functionality, but that is not reflected in the coverage report after moving the code. Any ideas why or how to solve?

Not sure if it's a Codecov issue or just the coverage is not being collected (feeling toward the latter), maybe you can check the reports manually and see? But since routing is moved out into a shared package and then depended on via `link`, I don't think jest's code coverage will count dependencies? Maybe there is some option to include specific internal dependencies as part of the report

packages/libs/routing/src/auth.ts
jvarho: You mean using IncomingMessage and ServerResponse?

That would pretty much mean implementing a state machine that manipulates a pair of them passed from the lambda-at-edge (or other) implementation? I was thinking this could be just a nice and easily testable library of pure functions. 🙂 Returning a new ServerResponse or IncomingMessage would get annoying fast (typing, streaming), I think.

Then again, moving as much logic to shared code as possible would, of course, make other implementations simpler. I'll play around with it a bit when I have time. But if you have strong opinions about how this should work, please share.

jvarho:
> Not sure if it's a Codecov issue or just the coverage is not being collected (feeling toward the latter), maybe you can check the reports manually and see?

Yeah, I did try to run coverage locally and saw before posting that it didn't collect them correctly. All I found was this through some bug report: https://github.com/pveyes/jest-multi-project-monorepo-coverage-bug But that does not seem to help – there is no projects field in jest config.

packages/libs/routing/src/auth.ts
dphang: Yeah, that's what I meant. I think for now it might be ok to keep it simpler first and do the conversion as needed later. Also Cloudflare workers that doesn't support Node.js yet so anyway they might need to be shimmed for Cloudflare workers? (they are on a restricted V8 environment)

dphang:
> > Not sure if it's a Codecov issue or just the coverage is not being collected (feeling toward the latter), maybe you can check the reports manually and see?
> 
> Yeah, I did try to run coverage locally and saw before posting that it didn't collect them correctly. All I found was this through some bug report: https://github.com/pveyes/jest-multi-project-monorepo-coverage-bug But that does not seem to apply here – there is no projects field in jest config. Suggests that it _should_ work, though.

Yeah, I am not quite sure then...

Though this problem will go away once there's a clear separation between core code and lambda@edge/AWS code, because most of those handler code / tests should be genericized and moved to the core package...

* Core code is tested via the default/api/image handler tests (but needs to be made generic and remove CF/AWS related stuff)
* Lambda-at-edge package will turn into just a lightweight wrapper/plugin package that wraps Lambda@Edge handler on top of the core code. Tests can just mock the core code and test the wrapper in a few example scenarios (returning response directly from origin request handler. Repeat for other platforms.
* e2e tests will cover full integration anyway, so with this there should be good coverage.

If no luck on the coverage problem, one idea is to go the other way: rename/move everything in lambda-at-edge -> core and then slowly genericize it and move lambda-at-edge specific code out into a lambda-at-edge-plugin package. So instead of refactoring the core code out into its own package, we are refactoring the lambda-at-edge code out into its own package. Thoughts? I realize it might negate some of the value of this PR...

packages/libs/routing/src/auth.ts
jvarho: Apparently Cloudflare has Node.js support now: https://blog.cloudflare.com/node-js-support-cloudflare-workers/ The current 1 MB limit is a problem for Next.js though. And I'm not sure what their time limits are.

Anyway, I changed the approach slightly, returning "Route" objects instead, which lambda-at-edge handles. The idea is that eventually the only logic in lambda-at-edge request handler would be going through the different cases and either returning a redirect, rendering and returning a page or making an S3 request.

jvarho:
> Though this problem will go away once there's a clear separation between core code and lambda@edge/AWS code, because most of those handler code / tests should be genericized and moved to the core package...

Agreed. But at least for now I found that using moduleNameMapper in jest to use the uncompiled entry point makes coverage collection work.

dphang:
> > Though this problem will go away once there's a clear separation between core code and lambda@edge/AWS code, because most of those handler code / tests should be genericized and moved to the core package...
> 
> Agreed. But at least for now I found that using moduleNameMapper in jest to point to the uncompiled entry point makes coverage collection work.
> 
> I'd prefer to work this way rather than moving lambda-at-edge logic out so there's an opportunity to untangle the interleaved conditionals.

Sure, glad you found a way to make it work!

packages/libs/routing/src/auth.ts
dphang: Yeah, that makes sense. The 1MB limit may be ok for small projects but for most, even mine which is medium-sized, each page is like 950 kB compressed. So to fit all pages into one bundle would be hard. Although static pages-only apps can be supported

packages/libs/core/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@sls-next/routing",
```
dphang, line 2: Update name to @sls-next/core

package.json
```diff
@@ -75,7 +75,8 @@
       "<rootDir>/packages/**/*.{js,ts}"
     ],
     "moduleNameMapper": {
-      "fs-extra": "<rootDir>/node_modules/fs-extra"
+      "fs-extra": "<rootDir>/node_modules/fs-extra",
+      "@sls-next/core": "<rootDir>/packages/libs/core/src/index.ts"
```
dphang, line 79: Great! Maybe add a note in readme as well for future reference :)

packages/libs/core/src/locale.ts
```diff
@@ -0,0 +1,57 @@
...
+) => {
+  if (routesManifest.i18n) {
+    const locales = routesManifest.i18n.locales;
+    const defaultLocale = routesManifest.i18n.defaultLocale;
+
+    const preferredLanguage = Accept.language(acceptLanguage);
+
+    // Find language in locale that matches preferred language
+    for (const locale of locales) {
+      if (preferredLanguage.toLowerCase() === locale.toLowerCase()) {
```
dphang, line 47: I know I wrote this code before but maybe you can to `toLowerCase()` in line 43 instead to help with perf a bit

packages/libs/core/tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "sourceMap": false,
+    "removeComments": true
```
dphang, line 5: Would this mean the standard Rollup bundle (unminified) will remove comments as well?

dphang:
LGTM save minor comments, there are a few things I noticed to fix but they are out of scope for your PR (more like some minor improvements to existing code I wrote) so I can fix it later.

Since this is a refactor, could you also post what the new handlers might look like, e.g the handler size vs. performance differences (expect to be negligible) vs. before the refactor

packages/libs/core/tsconfig.build.json
jvarho: Apparently yes, changed to false. Minification will remove them.

packages/libs/core/src/locale.ts
jvarho: I would hope that V8 would to that anyway, but might as well.

However, I meant to ask about that code. Is there any reason it uses `Accept.language(header)` and matches manually, rather than letting the library do the matching: https://hapi.dev/module/accept/api/?v=5.0.2#languagelanguageheader-preferences

I think that would allow a better change of matching the user's preferences. Totally orthogonal to this PR though.

jvarho:
> LGTM save minor comments, there are a few things I noticed to fix but they are out of scope for your PR (more like some minor improvements to existing code I wrote) so I can fix it later.
> 
> Since this is a larger refactor, could you also post what the new handlers might look like, e.g the handler size vs. performance differences (expect to be negligible) vs. before the refactor

Currently this will increase handler sizes slightly, because there are some copied functions that are still directly used within lambda-at-edge. I was going to try to get some more route matching moved before this would be merged. And/or export those functions from core. But if you prefer to merge early that's ok too.

Trying to decipher the routing logic I've found at least a few bugs already.

E.g. currently dynamic SSR routes can override more specific dynamic HTML routes. And SSG fallback routes seem to override more specific SSR routes in some cases. Next.js routing logic sorts them all by segment depth: https://github.com/vercel/next.js/blob/3e6ce9071b849974d15a616529af6120b2449cb6/packages/next/next-server/lib/router/utils/sorted-routes.ts#L198

I think the best way to do things would be to similarly pre-sort routes correctly in the build step, so that the first matched route has priority.

packages/libs/core/src/locale.ts
dphang: Yea, I think there was no reason - I might have just missed that code. Feel free to update it!

dphang:
> > LGTM save minor comments, there are a few things I noticed to fix but they are out of scope for your PR (more like some minor improvements to existing code I wrote) so I can fix it later.
> > Since this is a larger refactor, could you also post what the new handlers might look like, e.g the handler size vs. performance differences (expect to be negligible) vs. before the refactor
> 
> Currently this will increase handler sizes slightly, because there are some copied functions that are still directly used within lambda-at-edge. I was going to try to get some more route matching moved before this would be merged. And/or export those functions from core. But if you prefer to merge early that's ok too.
> 
> Trying to decipher the routing logic I've found at least a few bugs already.
> 
> E.g. currently dynamic SSR routes can override more specific dynamic HTML routes. And SSG fallback routes seem to override more specific SSR routes in some cases. Next.js routing logic sorts them all by segment depth: https://github.com/vercel/next.js/blob/3e6ce9071b849974d15a616529af6120b2449cb6/packages/next/next-server/lib/router/utils/sorted-routes.ts#L198
> 
> I think the best way to do things would be to similarly pre-sort routes correctly in the build step, so that the first matched route has priority.

Thanks! Yeah, we definitely need to fix the routing priority especially if there is a mix of many types. I tried to fix some of these issues but there is not enough test coverage, the sorting by segment depth makes sense though

dphang:
> > LGTM save minor comments, there are a few things I noticed to fix but they are out of scope for your PR (more like some minor improvements to existing code I wrote) so I can fix it later.
> > Since this is a larger refactor, could you also post what the new handlers might look like, e.g the handler size vs. performance differences (expect to be negligible) vs. before the refactor
> 
> Currently this will increase handler sizes slightly, because there are some copied functions that are still directly used within lambda-at-edge. I was going to try to get some more route matching moved before this would be merged. And/or export those functions from core. But if you prefer to merge early that's ok too.
> 
> Trying to decipher the routing logic I've found at least a few bugs already.
> 
> E.g. currently dynamic SSR routes can override more specific dynamic HTML routes. And SSG fallback routes seem to override more specific SSR routes in some cases. Next.js routing logic sorts them all by segment depth: https://github.com/vercel/next.js/blob/3e6ce9071b849974d15a616529af6120b2449cb6/packages/next/next-server/lib/router/utils/sorted-routes.ts#L198
> 
> I think the best way to do things would be to similarly pre-sort routes correctly in the build step, so that the first matched route has priority.

Thanks! Yeah, we definitely need to fix the routing priority especially if there is a mix of many types. I tried to fix some of these issues but there is not enough test coverage, the sorting by segment depth makes sense though

dphang:
I think we can merge it now and merge additional changes later (anyway, I have released 1.19 already so this will be part of 1.20 alpha)

jvarho:
> I think we can merge it now and merge additional changes later (anyway, I have released 1.19 already so this will be part of 1.20 alpha)

Ok, makes sense to avoid needless conflicts. I'll probably try to move more route handling over tomorrow.

dphang (approved):

PR #1034, opened 2021-05-01
Delete old files from dist on build

jvarho:
Since #1029 rebuilding locally keeps accumulating rollup modules, making local builds difficult to deploy.

Each time you run `yarn` in root or under lambda-at-edge a new set of modules is created. They all get included in any builds that link to the local repository, like e2e-tests.

Not sure this is the best solution, but AFAICT rollup itself doesn't have an option to clean up the build directory.

### DIFF
--- a/packages/libs/lambda-at-edge/rollup.config.js
+++ b/packages/libs/lambda-at-edge/rollup.config.js
@@ -4,6 +4,7 @@ import { nodeResolve } from "@rollup/plugin-node-resolve";
 import externals from "rollup-plugin-node-externals";
 import json from "@rollup/plugin-json";
 import { terser } from "rollup-plugin-terser";
+import del from "rollup-plugin-delete";
 
 const LOCAL_EXTERNALS = [
   "./manifest.json",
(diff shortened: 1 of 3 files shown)


packages/libs/lambda-at-edge/rollup.config.js
```diff
@@ -21,6 +22,10 @@ const generateConfig = (input) => ({
     format: "cjs"
   },
   plugins: [
+    del({
+      targets:
+        './dist/${input.filename}/${input.minify ? "minified" : "standard"}'
```
dphang, line 27: ' should be `?

jvarho: Indeed, I managed to break it after testing 🤦🏽 fixed

dphang:
Yes this should be fine to ensure no extra chunks are generated. Actually I'll add another script to also clean entire dist directory first when calling yarn (e.g for fresh install, rebuilding entire project or publishing). But this is still useful in case you just want to just build the handlers

dphang (approved):

PR #1035, opened 2021-05-01
chore(lambda-at-edge): add command to clean dist directory

dphang:


dphang:
e2e test failure looks like transient error, this command isn't run yet, so I'm going to merge it.

PR #1040, opened 2021-05-05
Extend locale matching to secondary languages

jvarho:
This allows matching a user's language settings even if the first preference does not match the locales available.

E.g. someone whose browser is set to prefer German, but accept French, can be served a French version after this change if German is not an option.

### DIFF
--- a/packages/libs/core/src/locale.ts
+++ b/packages/libs/core/src/locale.ts
@@ -37,20 +37,22 @@ export const getAcceptLanguageLocale = (
   routesManifest: RoutesManifest
 ) => {
   if (routesManifest.i18n) {
-    const locales = routesManifest.i18n.locales;
     const defaultLocale = routesManifest.i18n.defaultLocale;
+    const locales = new Set(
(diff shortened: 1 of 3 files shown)


dphang (approved):

packages/libs/core/src/locale.ts
```diff
@@ -37,21 +37,23 @@ export const getAcceptLanguageLocale = (
   routesManifest: RoutesManifest
 ) => {
   if (routesManifest.i18n) {
-    const locales = routesManifest.i18n.locales;
     const defaultLocale = routesManifest.i18n.defaultLocale;
+    const locales = new Set(
+      routesManifest.i18n.locales.map((locale) => locale.toLowerCase())
```
dphang, line 42: It would be nice if this is already in lowercase but it's Next.js that generates the routes manifest, so we should try to preserve it as much as possible

jvarho: The case-correct versions are needed for route matching. But it would of course be possible to add lowercase versions to build manifest if desired.

PR #1044, opened 2021-05-07
Cold start improvements

jvarho:
These try to reduce the cold start penalty by trimming default lambda size, especially the unconditionally imported part.

The benefit is minor. Without extensive testing it looks like maybe a few percent off the init duration and ~2 MB off memory use when not needing any of the conditional imports (no locale redirect or preview headers).

Total size was reduced only slightly but the unconditionally imported chunks are smaller.

Before:

dist/default-handler/minified/default-handler-eb183eff.js 94760
dist/default-handler/minified/index.js 364

After:

dist/default-handler/minified/index.js 30339


dphang:
> Since e2e test run failure due to some build error that probably exist in master already?

I have seen that but it seems transient, sometimes GH actions can't the component.js file. I'm not sure why as I never get that locally. After retrying it usually works.

dphang (approved):

Issue #733, opened 2020-11-01
Dynamic rewrites 404ing


patricktyndall:
@dphang this is a continuation of the use case / repro from #705. I'm actually surprised to be just finding this now. In all of the discussion this specific setup must have been missed, but it's really the central issue for my use case.

### Describe the bug
With a page defined at `pages/content/[...all ]` and the following rewrite, visiting `/dynamic-1` gives a 404.
```
{
	source: '/:path*',
	destination: '/:path*',
},{
	source: '/:path*',
	destination: '/content/:path*',
}
```
### Actual behavior
I get a 404 visiting `/dynamic-1`, where next server gives a 200.

### Expected behavior
200

### Steps to reproduce
- Clone and deploy reproduction repo: https://github.com/patricktyndall/sls-next-bug-reproduction
- Visit `/dynamic-1`

### Versions
- OS/Environment: Mac
- @sls-next/serverless-component version: `1.18.0-alpha.27`
- Next.js version: `9.5.5`

dphang:
Yes, I think it should be because of the no-op rewrite not being implemented, I had thought you did not need it anymore based on your latest comments, so I didn't prioritize it. I'll work on it this upcoming week.

PR #1057, opened 2021-05-09
Core refactor part 2

jvarho:
Moves most of the request routing over to core.

I had to make test changes as well due to:

1. There were simply tests where the test data was wrong, but the old logic worked differently and happened to work.

2. The 404 logic was tested two ways (render and serve) with the same manifests. I dropped the non-static 404 tests. I hope that's ok, AFAIU Next.js will always generate a static 404 since at least 9.5. (Note: static 500 logic should be implemented at some point.)

3. Rewrites had "no-op rewrite" support (https://github.com/serverless-nextjs/serverless-next.js/issues/733) which I was not able to reproduce with Next.js server and dropped. Those add complexity, but if we want to diverge from Next.js I can try to reimplement them. However, with the new types of rewrites added in 10.2 that may not be possible consistently.

4. Rewriting query parameters worked inconsistently when locale support was/not used.

I did not yet fix the routing priority issues I found, but they are flagged in a couple of comments. This branch reproduces the same (buggy) logic of going through routes by type rather than specificity. Before fixing that the build logic should be refactored, as doing the correct thing efficiently will require manifest changes.

Next steps:

* Refactor build logic to core.
* Refactor custom headers to core.
* Refactor response handler routing.
* Refactor lambda and s3 logic to separate libraries (for sharing between apigw/cf).
* Refactor image handler to separate(?) library.
* Implement apigw/lambda backend.

Unrelated TODO:

* Core library could use more direct tests.
* 10.2 rewrites should now be easier to implement.
* Locale and basepath logic could do with a once-over. It works, but there's probably room for simplification.
* I noticed redirects have no cache-control, 308 should probably be cached by cloudfront.


dphang:
> After these latest commits all the tests pass and lib/core has >99% test coverage. Coverage only falls because this removes lines overall. Some e2e tests fail to build, but I can't see how it's because of any changes I've made when the other jobs go fine.
> 
> I will continue with the next steps, but this can be merged unless one of the changes mentioned above is a problem.

The yarn build step failed here but the workflow still continued, not sure why. The actual problem looks to be with the yarn cache:

```
error https://registry.yarnpkg.com/graceful-fs/-/graceful-fs-4.2.6.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, stat '/home/runner/.cache/yarn/v6/npm-graceful-fs-4.2.6-ff040b2b0853b23c3d31027523706f1885d76bee-integrity/node_modules/graceful-fs/clone.js'"
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
info fsevents@2.3.2: The platform "linux" is incompatible with this module.
info "fsevents@2.3.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning "@aws-sdk/client-s3 > @aws-sdk/middleware-retry > react-native-get-random-values@1.5.0" has unmet peer dependency "react-native@>=0.56".
warning "@vercel/nft > acorn-export-ns-from@0.1.0" has incorrect peer dependency "acorn@^6.0.1".
warning " > rollup-plugin-node-externals@2.2.0" has unmet peer dependency "builtin-modules@^3.1.0".
[4/4] Building fresh packages...
lerna ERR! yarn install --frozen-lockfile exited 1 in '@sls-next/serverless-cli'
lerna ERR! yarn install --frozen-lockfile exited 1 in '@sls-next/serverless-cli'
lerna WARN complete Waiting for 1 child process to exit. CTRL-C to exit immediately.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Usually rerun will fix it but GH Actions doesn't support just re-running a single job right now =(

packages/libs/core/src/rewrite.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Get the rewrite of the given path, if it exists. Otherwise return null.
+ * @param path
+ * @param routesManifest
+ * @param router
+ * @param normalisedPath
+ */
+export function getRewritePath(
+  uri: string,
+  routesManifest: RoutesManifest
+): string | null {
```
dphang, line 15: Maybe better if we are consistent and return `undefined` always instead of `null`? I guess not in scope of this PR but something I noticed. See: https://palantir.github.io/tslint/rules/no-null-keyword/. Probably I can add an eslint rule for it later on.

packages/libs/core/src/data.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (pages.ssg.nonDynamic[normalisedUri] && !isPreview) {
+    return {
+      isData: true,
+      isStatic: true,
+      file: uri
+    };
+  }
+  if (
+    pages.ssr.nonDynamic[normalisedUri] ||
+    pages.ssg.nonDynamic[normalisedUri]
```
dphang, line 48: Is line 48 needed?

packages/libs/core/src/rewrite.ts
jvarho: Yes, I agree and did make the same change to some other functions. I can do that here.

jvarho:
Ok, so upgrading the next.js versions fixed one issue, but uncovered another. There is a front end change that drops the default locale under some situations. I've reported the inconsistent behavior as a bug in next.js, but there are many locale related open issues, so I guess that functionality is still a bit in flux. I modified the tests to pass with the current behavior, not sure what the best solution is.

An api rewrite test fail was an actual bug in my code, which I fixed. So even though the e2e tests are annoying to get to run, they are useful.

dphang:
> Ok, so upgrading the next.js versions fixed one issue, but uncovered another. There is a front end change that drops the default locale under some situations. I've reported the inconsistent behavior as a bug in next.js, but there are many locale related open issues, so I guess that functionality is still a bit in flux. I modified the tests to pass with the current behavior, not sure what the best solution is.
> 
> An api rewrite test fail was an actual bug in my code, which I fixed. So even though the e2e tests are annoying to get to run, they are useful.

Thanks for fixing the tests. Yeah, I don't personally use the locales feature too much, so not sure what is the best either. If you can make it work with the latest version, that is probably the best compromise for now. If users have issues with the latest of this component, they should probably try to upgrade Next.js first.

dphang (approved):

PR #1058, opened 2021-05-09
Node 14

jvarho:
This would move from supporting node versions 10 and 12 to supporting 12 and 14. AWS added support for 14 [recently](https://aws.amazon.com/blogs/compute/node-js-14-x-runtime-now-available-in-aws-lambda/) and will drop support for 10 over the summer.

The second commit can wait if there's a need to support 10 for another few months. It doesn't make a huge difference at the moment. (Dropping 12 eventually will be more of a help for code generation.)


dphang:
I know plain Lambda supports Node 14 now, but does Lambda@Edge support it as well now?

jvarho:
> I know plain Lambda supports Node 14 now, but does Lambda@Edge support it as well now?

Yes: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-requirements-limits.html#lambda-requirements-lambda-function-configuration

jvarho:
Other possible things to add would be changing the default at some point and changing what e2e tests use.

dphang:
> > I know plain Lambda supports Node 14 now, but does Lambda@Edge support it as well now?
> 
> Yes: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-requirements-limits.html#lambda-requirements-lambda-function-configuration

Great! Looks like the doc was recently updated on April 29.

dphang:
I'll merge this after e2e tests pass and I can update the policy to ignore the 10.x build

dphang:
e2e tests failed on one instance but it's due to throttling by AWS - will check with @danielcondemarin if we can get a quota increase or maybe just add a random 0-15 second sleep before each deploy to decrease the chance of deploying at same time

dphang (approved):

Issue #1030, opened 2021-04-29
Not compatible with nextJs > 10


Friis1978:
### Describe the bug
When updating to Next.js > v.10.0.0, we could not deploy the code in github actions.
If someone else have seen something similar, we really want your comments how to resolve this issue. 

### Actual behavior
error:
  AccessDenied: User: arn:aws:iam::***:user/github-cli is not authorized to perform: iam:GetRole on resource: role 7kv1mz9-eng6t1i
    at Request.extractError (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/protocol/query.js:50:29)
    at Request.callListeners (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/request.js:688:14)
    at Request.transition (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/request.js:690:12)
    at Request.callListeners (/home/runner/.serverless/components/registry/npm/@sls-next/serverless-component@1.19.0-alpha.30/node_modules/aws-sdk/lib/sequential_executor.js:116:18) {
[...]

PR #1065, opened 2021-05-10
Check if image component is used

jvarho:
The past couple of versions of Next.js have an indication of whether the image component is imported in .next/export-marker.json so use it to avoid building image handler when not used.

Should prevent issues like #1030 for people upgrading with a restricted role.


dphang:
Woops, looks like too many workflows were run at the same time, so we hit the GitHub API rate limit. Will try the workflow again after a few hours.

dphang (approved):

PR #1072, opened 2021-05-16
Core build refactor

jvarho:
Moves manifest build logic and response handler routing to core and fixes bugs.

Most of the diff is tests.

I added a script for generating the test case manifests using a new Next.js version because they were inconsistent in many ways. Doing that uncovered a bug with SSG routing, which I fixed.

Having a way to rebuild the test manifests also made it easier to change build logic. I added a dynamic route array that determines sorted route precedence and changed the format that routes are stored in. That fixed an issue with e.g. SSR routes trumping more specific HTML routes.

As a future step, special handling of catchall routes can be dropped as can the duplication of regex data and maybe locale routes. Also, it should now be easy to add performance tweaks like adding a prefix match step to avoid compiling regexes for routes that cannot match. (Requires benchmarking.)

Next steps:

* More build/response handler refactor?
* Refactor lambda and s3 logic to separate libraries (for sharing between apigw/cf).
* Refactor image handler to separate(?) library.
* Implement apigw/lambda backend.

Other possible changes this allows:

* Passing original request.uri as a header could simplify response handling and fix rewrite issues.
* ISR could probably use StaticRoute directly. (I suspect there are ISR bugs with basepath and locales.)
* Non-fallback dynamic SSG could be turned to 404 directly in the request handler.
* To get rid of response handler or not?


jvarho:
Fixing the remaining dynamic route e2e fails properly requires changes to the response handler, so I'll wait for ISR to land first.

dphang:
> Fixing the remaining dynamic route e2e fails properly requires changes to the response handler, so I'll wait for ISR to land first.

Are you planning to get rid of the origin response handler? I think it would simplify to just have one handler, and existing logic can be done using the AWS s3 client. Anyway I think you would need to use S3 when you want to serve all static pages through APIGW/Lambda, right?

jvarho:
> > Fixing the remaining dynamic route e2e fails properly requires changes to the response handler, so I'll wait for ISR to land first.
> 
> Are you planning to get rid of the origin response handler? I think it would simplify and improve perf to just have one handler, and existing logic can be done using the AWS s3 client. You would need to use S3 when you want to serve all static pages through APIGW/Lambda, right?

Not in this PR at least. It would simplify some things, but there are a lot of tests to be rewritten and it may hurt performance...

jvarho:
> You would need to use S3 when you want to serve all static pages through APIGW/Lambda, right?

Regarding this... The tradeoffs are different in (non-edge) lambda, where it is easy to avoid the cold start penalty and there's more space to work with. A large number of static pages could even be bundled in the lambda.

(I'm also wondering if there's a way to use S3 object lambda instead of apigw somehow. But I haven't figured out a good way.)

dphang:
> > > Fixing the remaining dynamic route e2e fails properly requires changes to the response handler, so I'll wait for ISR to land first.
> > 
> > 
> > Are you planning to get rid of the origin response handler? I think it would simplify and improve perf to just have one handler, and existing logic can be done using the AWS s3 client. You would need to use S3 when you want to serve all static pages through APIGW/Lambda, right?
> 
> Not in this PR at least. It would simplify some things, but there are a lot of tests to be rewritten and it may hurt performance for static sites...

Hm, that was my plan at least to simplify things and I believe it should also improve perf. At least for me, I don't actually need the functionality in the origin response handler, so it's just adding to cache misses on my SSR/SSG pages.

It should get rid of the overhead of initializing the response handler (still takes a bit of time even if warm). The AWS S3 client actually utilizes AWS's private networks if run in Lambda. From my tests a while back I remember it was still faster (i.e init + using S3 client in origin request handler < overhead of origin response handler + extra network hops from origin request handler -> S3 via CloudFront -> origin response handler). The only downside I guess is you could pay a little bit more for bandwidth/time since Lambda calls S3 and there is that 1 MB Lambda response limit (though pages, in practice should almost always be < 1 MB?).

> > You would need to use S3 when you want to serve all static pages through APIGW/Lambda, right?
> 
> Regarding this... The tradeoffs are different in (non-edge) lambda, where it is easy to avoid the cold start penalty and there's more space to work with. A large number of static pages could even be bundled in the lambda.
> 
> (I'm also wondering if there's a way to use S3 object lambda instead of apigw somehow. But I haven't figured out a good way.)

Yes definitely cold starts are more avoidable on regular Lambdas. Actually I wanted to bundle pages into even the Lambda@Edge, so as to avoid the network call to S3 (whether via CloudFront or the AWS client). It would work unless there are apps with 1000s of pages or heavy routes, so it could potentially be a configuration / we could detect if the user's pages + JS files can fit into the Lambda. I think the edge vs. regular Lambdas have the same size limits IIRC (50 MB compressed?).

jvarho:
> It should get rid of the overhead of initializing the response handler (still takes a bit of time even if warm). The AWS S3 client actually utilizes AWS's private networks if run in Lambda. From my tests a while back I remember it was still faster (i.e init + using S3 client in origin request handler < overhead of origin response handler + extra network hops from origin request handler -> S3 via CloudFront -> origin response handler).

That's promising, although I wonder if that's the case with all kinds of pages?

> The only downside I guess is you could pay a little bit more for bandwidth/time since Lambda calls S3 and there is that 1 MB Lambda response limit (though pages, in practice should almost always be < 1 MB?).

That reminds me that I was wondering if SSR pages (and api pages) should use S3 put and origin requests if they go over 1MB. Does not necessarily mean that a response handler is required of course.

> I think the edge vs. regular Lambdas have the same size limits IIRC (50 MB compressed?).

For some reason I thought it was larger, but you are correct, though there's of course containers.

gmontanola:
> Other possible changes this allows:
>
>     * ISR could probably use StaticRoute directly. (I suspect there are ISR bugs with basepath and locales.)

This is affecting us here. The locale path is used and leads to this error on the ISR Lambda:

```
Cannot find module './pages/pt-BR/perfil/[slug]'\nRequire stack:\n- /var/task/index.js\n- /var/runtime/UserFunction.js\n- /var/runtime/index.js
```

Would you happen to know of a workaround? And thanks for spotting this right away :)

dphang:
> > Other possible changes this allows:
> > ```
> > * ISR could probably use StaticRoute directly. (I suspect there are ISR bugs with basepath and locales.)
> > ```
> 
> This is affecting us here. The locale path is used and leads to this error on the ISR Lambda:
> 
> ```
> Cannot find module './pages/pt-BR/perfil/[slug]'\nRequire stack:\n- /var/task/index.js\n- /var/runtime/UserFunction.js\n- /var/runtime/index.js
> ```
> 
> Would you happen to know of a workaround? And thanks for spotting this right away :)

Thanks for reporting - @kirkness FYI, looks like we need to add the coverage for basepath/locales in e2e tests and fix this too

dphang (approved):

dphang:
anyway, I think this LGTM, thanks for improving this! For other comments we can address in the next PRs

PR #1079, opened 2021-05-18
Build and stash packages only once in e2e tests

jvarho:
Moves building packages and utils for e2e tests out of the matrix into the initialize step. This reduces the chance of failures due to one or more of the builds failing.

This may make it slower to run e2e tests (when there's no queue) because stashing the build takes an extra minute or so.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -20,6 +20,12 @@ jobs:
       pr_git_sha: ${{ steps.set_pr_git_sha.outputs.pr_git_sha }}
 
     steps:
+      # For non-fork PRs
+      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


jvarho:
Won't help with flaky tests, of course. There's a single failure from a test that passed in master, which suggests the test is somehow wrong.

jvarho:
Ah, it's the same issue that I already fixed for several other tests in the same file. I.e. Next.js changes the location after the page has loaded. It sometimes passes if cypress checks the location before the transition has happened.

kirkness (approved):
Nice one!

dphang:
Great! This was on the TODO but just hadn't had a chance.

Still not sure why yarn sometimes continues even though one of the packages had an error code, though that can looked at separately

PR #1091, opened 2021-05-20
Trim the build manifest and reuse routing logic for ISR

jvarho:
Catchall routes are now handled as other dynamic routes. Regex duplication removed as well as some unused SSG fields.

And because I touched it, I reused routing logic for ISR, hopefully fixing locale support.


dphang:
One e2e test suite failed due to CF invalidation timing out, but I think there is enough coverage otherwise. I think it LGTM.

dphang (approved):

PR #1092, opened 2021-05-20
chore: use npmjs registry for lockfiles

dphang:
For some reason, GitHub actions has some flakiness with the yarn registry in lockfiles, so choosing to use the npm registry. This should help with the random build failures due to registry issues. Seems to be a known issue from some research.


dphang:
Flaky ISR e2e tests.

PR #1095, opened 2021-05-21
fix(lambda-at-edge): copy js chunks from .next/serverless/chunks to api/default lambdas if present

dphang:
This is two changes:

1. Copy chunks folder contents into the api/default lambdas. It seems this is generated in newer versions (e.g 10.0.6 and above at least) so we should copy it if present.

2. Update e2e tests to latest 10.2.2.

Going forward, it may be better to drop the previous tests since it makes the codebase less maintainable (have to maintain backwards compatibility) due to Next.js's various model changes (e.g for locales). Also no point doing the version range since they will be locked in the lockfile, we can manually update this as needed when newer Next.js versions come out.

### DIFF
--- a/packages/e2e-tests/next-app-dynamic-routes/package.json
+++ b/packages/e2e-tests/next-app-dynamic-routes/package.json
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/serverless-next.js/serverless-next.js",
   "dependencies": {
     "isomorphic-fetch": "2.2.1",
-    "next": "~10.0.1",
+    "next": "10.2.2",
     "react": "17.0.1",
     "react-dom": "17.0.1"
   },
--- a/packages/libs/lambda-at-edge/src/build.ts
+++ b/packages/libs/lambda-at-edge/src/build.ts
@@ -321,6 +321,12 @@ class Builder {
           }
         }
       ),
+      fse.existsSync(join(this.serverlessDir, "chunks"))
+        ? fse.copy(
+            join(this.serverlessDir, "chunks"),
+            join(this.outputDir, DEFAULT_LAMBDA_CODE_DIR, "chunks")
+          )
+        : Promise.resolve(),
       fse.copy(
         join(this.dotNextDir, "prerender-manifest.json"),
         join(this.outputDir, DEFAULT_LAMBDA_CODE_DIR, "prerender-manifest.json")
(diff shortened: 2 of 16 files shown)


jvarho:
Looks like between Next.js 10.2.0 and 10.2.2 the behavior of serverless build has changed a lot.

A small project I have goes from a 21MB default lambda to 3MB after I manually copy the chunks over. That pretty much makes useServerlessTraceTarget obsolete I guess? Unless I'm just not using enough libraries for it to matter anymore.

Looks good, but I wish this sort of thing didn't change in minor versions...

packages/e2e-tests/next-app-dynamic-routes/package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/serverless-next.js/serverless-next.js",
   "dependencies": {
     "isomorphic-fetch": "2.2.1",
-    "next": "~10.0.1",
+    "next": "10.2.2",
```
jvarho, line 23: I wonder if it would be better to `yarn add next@latest` in the e2e-test workflow? That would test changes in nextjs automatically.

E.g. have a matrix with next@latest and next@^10.0.0 instead of the prev-* tests, then skip any tests that rely on newer features.

packages/libs/lambda-at-edge/src/build.ts
```diff
@@ -321,6 +321,14 @@ class Builder {
           }
         }
       ),
+      // copy chunks if present and not using serverless trace
```
jvarho, line 324: Seems the build produces chunks even with serverless trace target. Are you sure they can be left out? Or are they added elsewhere?

jvarho: Never mind, they get added with the other traced files.

packages/e2e-tests/next-app-dynamic-routes/package.json
dphang: Yeah, good idea. So we don't need to keep updating the versions in multiple places

dphang:
> Looks like between Next.js 10.2.0 and 10.2.2 the behavior of serverless build has changed a lot.
> 
> A small project I have goes from a 21MB default lambda to 3MB after I manually copy the chunks over. That pretty much makes useServerlessTraceTarget obsolete I guess? Unless I'm just not using enough libraries for it to matter anymore.
> 
> Looks good, but I wish this sort of thing didn't change in minor versions...

Yes, the chunks are copied by serverless trace target

packages/e2e-tests/next-app-dynamic-routes/package.json
dphang: Although right now the test app states are stored in s3 based on the name of the app. Maybe we can just always run on the latest version

PR #1099, opened 2021-05-21
fix: also copy chunks for the revalidate lambda

kirkness:
Tests may still be flakey for ISR, not sure, however, I noticed they were failing legitimately due to the `chunks` not being copied into the lambda dir.


dphang:
Thanks! missed that one

dphang (approved):

dphang (approved):

PR #1107, opened 2021-05-22
Core handler refactor

jvarho:
Moves most remaining api and page handling logic to core. Core handler code now uses IncomingMessage/ServerResponse as desired.

Functional changes:

* Custom headers no longer have to use the S3 route. I can also add support for that if desired for compatibility.
* Fixes bugs if root route is SSG. Depending on the situation it might either not work or be effectively treated as SSR.

What remains in lambda-at-edge is:

* All logic related to static files and S3
* Image handler (TODO)
* A lot of build logic (TODO?)
* External rewrites – I think they could probably be handled without node-fetch in cloudfront, but also I could not get the mocks to work across packages.
* Revalidate triggering logic

I changed eslint rules a bit, since dynamic `require` is used extensively for good reason and some "errors" in tests were too common that I'd want to fix them. I hope those changes are ok.


packages/libs/core/src/handle/unauthorized.ts
```diff
@@ -0,0 +1,9 @@
+import { setHeadersFromRoute } from "./headers";
+import { Event, UnauthorizedRoute } from "../types";
+
+export const unauthorized = (event: Event, route: UnauthorizedRoute) => {
+  setHeadersFromRoute(event, route);
+  event.res.statusCode = route.status;
+  event.res.statusMessage = route.statusDescription;
+  event.res.end("Unauthorized");
```
dphang, line 8: Maybe unauthorized can be empty (I think I might have put this before)? Then the browser can render a nicer page.

packages/libs/lambda-at-edge/tests/shared-fixtures/built-artifact/pages/fallback/[slug].js
```diff
@@ -3,7 +3,7 @@ module.exports = {
   render: (req, res) => {
     res.end("pages/fallback/[slug].js");
   },
-  renderReqToHTML: (req, res) => {
+  renderReqToHTML: jest.fn((req, res) => {
```
Comment (dphang) [inline packages/libs/lambda-at-edge/tests/shared-fixtures/built-artifact/pages/fallback/[slug].js:6]: Ah nice, so we can check if they were called properly

packages/libs/core/src/handle/unauthorized.ts
jvarho: Sure, I can change that. (Although the feature itself is supposed to be dev only.)

dphang (approved):

PR #1109, opened 2021-05-23
Try to fix build step return code for CI

jvarho:
Based on testing locally, it seems the exit code is lost in the `|| true` in postinstall which I think is meant to make `opencollective-postinstall` non-erroring.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "lint": "eslint .",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "integration": "jest --runInBand --config jest.integration.config.json --setupTestFrameworkScriptFile=./jest.integration.setup.js",
-    "postinstall": "yarn packages-install && opencollective-postinstall || true",
+    "postinstall": "yarn packages-install && (opencollective-postinstall || true)",
     "docs": "cd documentation && yarn && yarn build"
   },
   "repository": {
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -156,6 +156,24 @@ jobs:
           yarn e2e:ci
         timeout-minutes: 30 # In case something goes wrong
 
+  finalize:
+    if: always()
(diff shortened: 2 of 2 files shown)


jvarho:
Seems to work: https://github.com/serverless-nextjs/serverless-next.js/runs/2648836468?check_suite_focus=true

However, e2e failure in initialization leaves state pending.

dphang:
Great, thanks!

dphang (approved):

package.json
```diff
@@ -25,7 +25,7 @@
     "lint": "eslint .",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "integration": "jest --runInBand --config jest.integration.config.json --setupTestFrameworkScriptFile=./jest.integration.setup.js",
-    "postinstall": "yarn packages-install && opencollective-postinstall || true",
+    "postinstall": "yarn packages-install && (opencollective-postinstall || true)",
```
dphang, line 28: @jvarho actually I think this line was intentional as lerna publish step would fail otherwise - just had an error after trying to publish the new version (although the packages publishes successfully, I think somehow this applies to the root only). Basically it publishes dependencies and then updates the package.json to resolve the package links, and I think at the last step it seems to try to run postinstall again for some reason and fails because it thinks the lockfile should be updated (but actually not because the package.json updates should be reverted back to links).

I'll try to figure out how to resolve this more gracefully..

jvarho: Ugh, that sounds pretty convoluted. In that case maybe the fix is doing something other than simply `yarn` as the build command in CI. I tried things like `lerna bootstrap` locally, but I'm not sure how all the lifecycle phases connect because there are so many interlinked tools. 😬

dphang: Yeah, I am trying out something, it might be the `publish` command added at the root `package.json` is running again after all packages have already been published, it seems it might be unnecessary? I removed it and updated the `postinstall` command to remove the `true`, will see if it works for the next publish...

PR #1110, opened 2021-05-23
Fix accept language redirect condition

### LINKED ISSUE #1105: Infinite 307 redirects on automatic locale detection when upgrading to anything greater than serverless-component@1.19.0
Hey, there seems to be a bug with the latest serverless component with automatic locale detection.

Nextjs version: 10.2.2.
Machine: Mac OSX

This is my nextjs configuration:
```
{
        target: 'serverless',
        devIndicators: {
            autoPrerender: false,
        },
        i18n: {
            locales: ['en', 'es'],
            defaultLocale: 'en',
        },
    }
```

Here are my steps:

I updated to the latest alpha version of serverless component.
When my Language is 'es' (configured in google chrome settings) and I land on mywebsite.com/
I get redirected to `mywebsite.com/es`
and then I notice an infinite  307 redirects to `/es` after this.

After downgrading to `serverless-component@1.19.0` the issue stops happening.

Has anyone encountered this issue too? 

Just a note, I have configured my cloudfront to not pass any cookies. (most likely not the cause)

```
        cloudfront:
            distributionId: XXXXXX
            defaults:
                forward:
                    cookies: 'none'
```

Let me know if there is any other information I can provide

jvarho:
Fixes #1105 

This feature could use some negative-case unit tests and _any_ e2e tests.

### DIFF
--- a/packages/libs/core/src/redirect.ts
+++ b/packages/libs/core/src/redirect.ts
@@ -91,7 +91,7 @@ export async function getLanguageRedirectPath(
   const trailingSlash = manifest.trailingSlash;
   const rootUri = basePath ? `${basePath}${trailingSlash ? "/" : ""}` : "/";
 
-  if (req.uri === rootUri || acceptLanguage) {
+  if (req.uri === rootUri && acceptLanguage) {
     return await getAcceptLanguageLocale(
       acceptLanguage,
       manifest,


jvarho:
I'll see if I can add tests in #1107 where I still have other tests to fix and write. But this seems obvious enough to merge as is, if current tests pass.

dphang (approved):

PR #1126, opened 2021-05-25
Fix page inclusion logic in default-handler build

### LINKED ISSUE #1081: Using getStaticProps leads to RequestEntityTooLargeException
### Describe the bug
I recently introduced `next-i18next` into my app and it requires that I use `getStaticProps` on every page. Size of my lambda exploded and I hit the dreadful `RequestEntityTooLargeException`.

### Actual behavior
I took a look at what happens in `.serverless_nextjs/default-lambda/pages` and it turns out that if the page has `getStaticProps`, it is included in lambda, therefore size exploded. Is that the expected behavior? I tested and for size of lambda it does not matter if i use getServerSideProps or getStaticProps and that seems to go against the docs.

### Versions
- OS/Environment: ubuntu
- @sls-next/serverless-component version: 1.19.0-alpha.45 , 1.19.1-alpha.7
- Next.js version: 10.2.0

### Additional context
For what its worth, I have a single `getInitialProps` at `_document.ts` due to styling solution, but removing it did not change anything.

### Checklist

- [X] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
- [X] Please first try using the most recent `latest` or `alpha` `@sls-next/serverless-component` [release version](https://github.com/serverless-nextjs/serverless-next.js/releases), which may have already fixed your issue or implemented the feature you are trying to use. Note that the old `serverless-next.js` component and the `serverless-next.js` plugin are deprecated and no longer maintained.

jvarho:
With new Next.js versions using locales in prerender manifest routes, the SSG page dropping logic no longer works. This moves the logic to core by using the build manifest for determining what's included. Should fix:

* Pages always being included when locales are used
* Dynamic `fallback: false` pages always being included
* ISR pages **not** being included (if no locales or apis are in use)

This may fix #1081


dphang (approved):

Reiiya:
Can confirm, Its fixed :)

PR #1132, opened 2021-05-27
Fix locale rewrites to dynamic ssr

### LINKED ISSUE #1131: Rewritten SSR page with dynamic route missing query for default locale
### Describe the bug

For a localised site, the SSR can't resolve dynamic path parameters for the default locale if the page is a rewrite e.g. `/path-rewrite/slug` -> `/ssr-page/slug`. Adding the locale prefix to the URL fixes this. It also works when running Next.js without deploying it to AWS.

### Actual behavior

The dynamic path parameters e.g. slug is `undefined`.

### Expected behavior

The dynamic path parameter should be parsed from the URL.

### Steps to reproduce

1. Clone the reproduction case https://github.com/myrjola/serverless-next.js/commit/373986e92c65e11ca00fc1550c1058c297e36f12
2. Deploy the packages/e2e-tests/next-app-with-locales to AWS
3. /path-rewrite/slug and verify that `slug` does not propagate to the server-side renderer, i.e., `query.slug` is `undefined`

### Screenshots/Code/Logs

In the default locale, `query.slug` is `undefined`.

image

Adding the locale prefix fixes it.

image

### Versions

- OS/Environment: MacOS
- @sls-next/serverless-component version: 1.20.0-alpha.13
- Next.js version: 10.2.2

### Checklist

- [X] You have reviewed the README and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
- [X] Please first try using the most recent `latest` or `alpha` `@sls-next/serverless-component` [release version](https://github.com/serverless-nextjs/serverless-next.js/releases), which may have already fixed your issue or implemented the feature you are trying to use. Note that the old `serverless-next.js` component and the `serverless-next.js` plugin are deprecated and no longer maintained.

jvarho:
Fixes #1131

Next.js routing requirements are pretty odd here. It needs the "original" url (before rewrites) to support asPath, but cannot cope with locales.

SSG routes (fallback or ISR) already get the locale added for fetching the static page.

### DIFF
--- a/packages/e2e-tests/next-app-with-locales/cypress/integration/rewrites.test.ts
+++ b/packages/e2e-tests/next-app-with-locales/cypress/integration/rewrites.test.ts
@@ -26,6 +26,11 @@ describe("Rewrites Tests", () => {
         expectedStatus: 200
       },
       {
+        path: "/wildcard-rewrite-2/foo",
+        expectedRewrite: "/dynamic-ssr/foo",
+        expectedStatus: 200
+      },
+      {
         path: "/regex-rewrite-1/123",
         expectedRewrite: "/ssr-page?slug=123",
         expectedStatus: 200
--- a/packages/e2e-tests/next-app-with-locales/next.config.js
+++ b/packages/e2e-tests/next-app-with-locales/next.config.js
@@ -81,6 +81,10 @@ module.exports = {
         destination: "/ssr-page"
       },
       {
+        source: "/wildcard-rewrite-2/:slug",
+        destination: "/dynamic-ssr/:slug"
+      },
+      {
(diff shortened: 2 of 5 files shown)


jvarho:
Can't seem to have an e2e test run go through... Each run something fails. This time it's github api returning 403, but the actual change is ok.

dphang:
> Can't seem to have an e2e test run go through... Each run something fails. This time it's github api returning 403, but the actual change is ok.

Yea, that's the problem with having so many e2e tests (and GH Actions doesn't support re-running just a single job yet). But the redirect tests that failed has coverage in other jobs, so it should be ok.

dphang (approved):

PR #1135, opened 2021-05-29
fix(core, lambda-at-edge): support static 500 page and localized stat…

dphang:
…ic error pages

* Also fix retrieval of localized static 404/500 pages from the locale-specific directory e.g `en/404.html` or `en/500.html`

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -247,6 +247,13 @@ const handleOriginResponse = async ({
       return response;
     }
 
+    // Set 500 status code for 500.html page. We do not need normalised URI as it will always be "/404.html"
+    if (s3Uri.endsWith("/500.html")) {
+      response.status = "500";
+      response.statusDescription = "Internal Server Error";
+      return response;
+    }
+
     const staticRegenerationResponse = getStaticRegenerationResponse({
       expiresHeader: response.headers?.expires?.[0]?.value || "",
       lastModifiedHeader: response.headers?.["last-modified"]?.[0]?.value || "",
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -365,9 +365,14 @@ const handleOriginResponse = async ({
     const bodyString = await getStream.default(s3Response.Body as Readable);
 
     const is404 = file.endsWith("/404.html");
+    const is500 = file.endsWith("/500.html");
     return {
-      status: is404 ? "404" : "200",
-      statusDescription: is404 ? "Not Found" : "OK",
+      status: is500 ? "500" : is404 ? "404" : "200",
+      statusDescription: is500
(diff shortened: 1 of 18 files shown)


dphang:
Test fails which seems to be due to incorrect static default 500.html generated when using locales. Upon investigation, it looks like Next.js is not generating it correctly. Filed a bug in there: https://github.com/vercel/next.js/issues/25574

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -247,6 +247,13 @@ const handleOriginResponse = async ({
       return response;
     }
 
+    // Set 500 status code for 500.html page. We do not need normalised URI as it will always be "/404.html"
+    if (s3Uri.endsWith("/500.html")) {
```
dphang, line 251: Will add a test for this. In the future if/when origin response handler is removed, it could be moved to CloudFront error page or we could use aws-sdk s3-client to retrieve the page from S3 while in the origin request handler (and thus be able to set the status code there).

jvarho:
I think you need to add some cache-control logic?

If SSR results in 500, the static 500 will have normal static page cache control in S3 and be cached in cloudfront, AFAICT. For 404 that path is fine, since it will only happen when the page is unroutable, so it will stay 404.

dphang:
> I think you need to add some cache-control logic?
> 
> If SSR results in 500, the static 500 will have normal static page cache control in S3 and be cached in cloudfront, AFAICT. For 404 that path is fine, since it will only happen when the page is unroutable, so it will stay 404.

Yea, good catch, I made the origin response handler override the cache-control header

jvarho:
> > I think you need to add some cache-control logic?
> > If SSR results in 500, the static 500 will have normal static page cache control in S3 and be cached in cloudfront, AFAICT. For 404 that path is fine, since it will only happen when the page is unroutable, so it will stay 404.
> 
> Yea, good catch, I made the origin response handler override the cache-control header

I guess you need the same in the fallback path since your core change applies to both? https://github.com/serverless-nextjs/serverless-next.js/pull/1135/files#diff-f4709477772e5badfaaf543f6104de4b4ab3bf8300a056c088f84c184d800288R367

That one also needs the correct status, if I'm correct in that it can get the static 500 now.

dphang:
> > > I think you need to add some cache-control logic?
> > > If SSR results in 500, the static 500 will have normal static page cache control in S3 and be cached in cloudfront, AFAICT. For 404 that path is fine, since it will only happen when the page is unroutable, so it will stay 404.
> > 
> > 
> > Yea, good catch, I made the origin response handler override the cache-control header
> 
> I guess you need the same in the fallback path since your core change applies to both? https://github.com/serverless-nextjs/serverless-next.js/pull/1135/files#diff-f4709477772e5badfaaf543f6104de4b4ab3bf8300a056c088f84c184d800288R367
> 
> That one also needs the correct status, if I'm correct in that it can get the static 500 now.

I thought fallback path is either a fallback page or the static 404 page, so it should not be able to error ever, right?

Issue #1098, opened 2021-05-21
Tracking ISR issues


dphang:
Just creating this to track any bugs or issues with the newly-released Incremental Static Regeneration feature -

* Improving e2e tests (flaky tests, more test coverage for locales, basepath etc?)

dphang:
@kirkness the ISR tests have been fairly flaky (mostly failing but sometimes 1/3 or 2/3 pass), if you have a chance please take a look otherwise I can check it out later on the weekend: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/863012949

kirkness:
Hi @dphang, keen to have a look at this, however, I won't be able to get much time on it until early next week, likely Tuesday.

kirkness:
Related: https://github.com/serverless-nextjs/serverless-next.js/pull/1099

dphang:
No worries, yeah it would be good to add more test coverage. The last few runs seem to not be flaky, though.

Miroko:
Not sure if I should open an issue about this or write straight to this thread...

We are getting the following debug log information when trying to deploy the `"@sls-next/serverless-component": "1.20.0-alpha.16"` (at the time of writing the latest alpha version) to AWS.

Our deploy script uses an existing serverless state made with previous deployment. That successfully deployed version uses SSR only, otherwise the configuration and component version are exactly the same. We are just trying to switch to use ISR.

Any help this this would be appreciated and thanks for the great work you have been doing for the ISR support!

```
  DEBUG ─ The QueueUrl has changed
  DEBUG ─ Deleting previous queue
  error:
  MissingRequiredParameter: Missing required key 'QueueUrl' in params
    at ParamValidator.fail (/builds/sok-digikehitys/raflaamo/raflaamo/frontend/node_modules/aws-sdk/lib/param_validator.js:50:37)
    at ParamValidator.validateStructure (/builds/sok-digikehitys/raflaamo/raflaamo/frontend/node_modules/aws-sdk/lib/param_validator.js:61:14)
    at ParamValidator.validateMember (/builds/sok-digikehitys/raflaamo/raflaamo/frontend/node_modules/aws-sdk/lib/param_validator.js:88:21)
    at ParamValidator.validate (/builds/sok-digikehitys/raflaamo/raflaamo/frontend/node_modules/aws-sdk/lib/param_validator.js:34:10)
[...]

PR #1147, opened 2021-06-02
fix: only delete the sqs queue if it already exists in the state

kirkness:
Initial issue: https://github.com/serverless-nextjs/serverless-next.js/issues/1098#issuecomment-852075948


dphang (approved):

dphang:
Thanks! Looks fine, I can merge it now but I'm on vacation until next weekend, so won't be able to publish until then

PR #1148, opened 2021-06-03
Use high throughput mode for regeneration SQS

jvarho:
Make use of the recently GA [high throughput mode](https://aws.amazon.com/about-aws/whats-new/2021/05/amazon-sqs-now-supports-a-high-throughput-mode-for-fifo-queues/) for FIFO queues.

Relatively big bump in aws-sdk versions, hopefully nothing breaks. I tried making the CDK change, but it seems even upgrading the libraries to latest would not yet support these parameters.

This is pretty niche, but shouldn't hurt anything even if you don't actually come close to the limits.

### DIFF
--- a/packages/serverless-components/aws-sqs/package.json
+++ b/packages/serverless-components/aws-sqs/package.json
@@ -19,6 +19,6 @@
   "homepage": "https://github.com/serverless-nextjs/serverless-next.js#readme",
   "peerDependencies": {
     "@serverless/core": "^1.1.2",
-    "aws-sdk": "^2.702.0"
+    "aws-sdk": "^2.920.0"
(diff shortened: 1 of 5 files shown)


dphang:
Looks like some attribute validation error causing failed deployment in the e2e test (next-app)

jvarho:
> Looks like some attribute validation error causing failed deployment in the e2e test (next-app)

Yeah, I'll have a look tomorrow.

jvarho:
I messed things up when I moved setting them from the sqs library (which I tested) to the serverless component.

Fixed now and seems to have passed the tests.

dphang (approved):

Issue #1111, opened 2021-05-23
1.20.0-alpha4 Version has S3 bucket region bug when bucket region is specfied in config file.


katakatataan:
### Describe the bug

When I specify bucket region.
InvaliadParameterValueException error is occured.
like below.
```
next-serverless:
  component: "@sls-next/serverless-component@1.20.0-alpha.4"
  inputs:
    roleArn: "{custom role}"
    bucketRegion: 'ap-northeast-1'
```

### Actual behavior

Actual behavior is written in Screenshots/Code/Logs section.

### Expected behavior

ap-northeast-1 region's S3 bucket is created and ISR works.

### Screenshots/Code/Logs

```
  error:
  InvalidParameterValueException: Function does not exist
    at Object.extractError (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/protocol/json.js:52:27)
    at Request.extractError (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/protocol/rest_json.js:55:8)
    at Request.callListeners (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/request.js:688:14)
    at Request.transition (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/{username}/.serverless/components/registry/npm/@sls-next/serverless-component@1.20.0-alpha.4/node_modules/aws-sdk/lib/state_machine.js:26:10
[...]

PR #1155, opened 2021-06-07
fix(nextjs-component): sqs region should be same as regeneration lamb…

dphang:
…da region (using s3 bucket region)

Should fix: https://github.com/serverless-nextjs/serverless-next.js/issues/1111


dphang:
Manually tested, seems like transient issue with CloudFront invalidations

PR #1219, opened 2021-06-13
Core static request refactor, getting rid of response handler

jvarho:
This implements my initial version of the changes discussed in #1144 – moving more of the logic to core and getting rid of the response handler.

Most default handler tests remain unported (I disabled them to run e2e tests) and I'll look into moving them to under libs/core where possible. The unlocalized, no basepath tests I updated while developing this.

A couple of questions:
1. Is the complexity of some pages using S3 client and some origin requests worth it?
2. What to do with public files and custom headers? The alternatives I can think of are to say "unsupported", to use S3 client and break any large files with headers or to add a response handler conditionally.


jvarho:
There are a lot of e2e test failures, which wasn't unexpected. Some are clear enough – e.g. some methods have to be disallowed in the request handler. Others less so. I'll have a look in the next few days.

dphang:
> This implements my initial version of the changes discussed in #1144 – moving more of the logic to core and getting rid of the response handler.
> 
> Most default handler tests remain unported (I disabled them to run e2e tests) and I'll look into moving them to under libs/core where possible. The unlocalized, no basepath tests I updated while developing this.
> 
> A couple of questions:
> 
> 1. Is the complexity of some pages using S3 client and some origin requests worth it?
> 2. What to do with public files and custom headers? The alternatives I can think of are to say "unsupported", to use S3 client and break any large files with headers or to add a response handler conditionally.

1. Yes I think pages should always use S3 client? Since they should almost always be < 1 MB.
2. Public files is in the grey area, I suppose most people may not be using custom headers on public files, but a few might have large files, so probably optimize for that? If they are using custom headers, probably we should say they are only supported if < 1 MB (we could add new cache behaviors specifically for public files with an origin response handler, but it might add too much complexity)

jvarho:
@dphang There are still many test failures, but this works well enough that I benchmarked it.

I've compared this branch to current master. I'm connecting to eu-central-1 with the S3 bucket in us-east-1, to try to include that latency in the picture. I'm requesting a HTML page, the same page each time, but with a query parameter to prevent cloudfront caching.

**Before:**

1st request (cold start)

0.51s END - START
68 + 7 ms Billed Duration
44.6 + 1.14 ms handler execution time

2nd request (1 minute later)

0.54s END - START
79 + 23 ms Billed Duration
0.39 + 0.29 ms handler execution time

3rd request (immediate after 2nd)

0.40s END - START
2 + 2 ms Billed Duration
0.29 + 0.19 ms handler execution time


**After:**


1st request (cold start)

2.26s END - START
2259 ms Billed Duration
2201.7 ms handler execution time

2nd request (1 minute later)

0.82s END - START
819 Billed Duration
688.5 ms handler execution time

3rd request (immediate after 2nd)

0.12s END - START
117 Billed Duration
106.7 ms handler execution time


I'm not sure how reproducible these results are, but it looks like the cold start penalty is much worse and the billed duration always significantly larger. In the best case it can be faster, but I suspect there is some caching going on that may not happen normally – normally the same file would not be requested soon (since it should get cached in cloudfront) unless it is e.g. a 404 page.

Maybe I've done something stupid in the implementation or am missing something, but this does not look very encouraging to me.

dphang:
Strange, let me try to test your changes too. I remember any AWS service called in Lambda is supposed to go over AWS's private networks as well, so using s3 client should be as fast as using an s3 origin.

When I tested it was a while back and with us-east-1 only, maybe that is affecting it?

dphang:
@jvarho maybe one thing you can check is whether it is using the right regional S3. From S3 docs https://docs.aws.amazon.com/general/latest/gr/s3.html. Basically it should be `s3.us-east-1.amazonaws.com` (or other bucket). I think that it should be using the right regional endpoints, but maybe it's not?

Recently with some AWS permanent redirect failure causing s3 uploads at deploy-time, gave me this idea. Since that somehow was not using the right endpoint and I had to set it manually, although that was in AWS SDK JS v2

dphang:
Nevermind, tested and it does seem to be using the right regional endpoints in AWS SDK JS v3. Though according to here it should go over AWS networks? https://aws.amazon.com/blogs/networking-and-content-delivery/leveraging-external-data-in-lambdaedge/

```
Leverage the AWS network by using AWS-based endpoints like Amazon S3, DynamoDB, API Gateway, ELB, and so on. When you connect to AWS resources, your application traffic travels on the high-quality network that Amazon monitors for both availability and low latency. This monitoring has the additional benefit of keeping error rates low and window sizes high.
```

jvarho:
I'm not pursuing this, so closing. The branch will be there in case someone wants to work on it.

PR #1258, opened 2021-06-18
Install latest next version in e2e tests

jvarho:
Now that next 11 has been released, test on it by installing the latest version in e2e tests.

Leaves prev- tests' next version alone. Those can be set to the oldest supported in package files.


dphang:
I would say to remove the existing `prev-*` tests and rename the current tests with `prev-*` prefix (so they are locked at 10.2.2 - though we should also bump that to 10.2.3 to capture the last version in Next.js 10). We could do it later though

.github/workflows/e2e-tests.yml
```diff
@@ -154,6 +154,13 @@ jobs:
       - name: "Uncompress build"
         run: tar -xf build-${{ matrix.os }}.tar.gz
 
+      - name: Choose latest next.js version
+        if: ${{ !startsWith(matrix.app, 'prev') }}
+        working-directory: packages/e2e-tests/${{ matrix.app }}
+        run: |
+          cd packages/e2e-tests/$APP_NAME
```
dphang, line 161: probably not needed if you already set the working directory?

jvarho: Yeah, left over from my first try, fixed.

jvarho:
> I would say to remove the existing `prev-*` tests and rename the current tests with `prev-*` prefix (so they are locked at 10.2.2 - though we should also bump that to 10.2.3 to capture the last version in Next.js 10). We could do it later though

So basically replace the current prev- tests with the code from the non-prev counterparts? Or should I also add the serverless-trace ones?

dphang:
> > I would say to remove the existing `prev-*` tests and rename the current tests with `prev-*` prefix (so they are locked at 10.2.2 - though we should also bump that to 10.2.3 to capture the last version in Next.js 10). We could do it later though
> 
> So basically replace the current prev- tests with the code from the non-prev counterparts? Or should I also add the serverless-trace ones?

Yeah I think serverless trace is kind of deprecated in the later versions, we can keep the code to build it but I think no need to test it, so probably the existing tests should be copied (except serverless trace) and fixed to version 10.2.3.

I guess also one issue with this change is it will force us to keep updating our changes (e.g e2e tests might suddenly fail if Next.js has a breaking change), but it might be a good thing so we catch issues in e2e tests before they are reported. I think it should be fine, though

jvarho:
> I guess also one issue with this change is it will force us to keep updating our changes (e.g e2e tests might suddenly fail if Next.js has a breaking change), but it might be a good thing so we catch issues in e2e tests before they are reported. I think it should be fine, though

Yeah, I think it's good to uncover compatibility problems sooner rather than later.

dphang (approved):

Issue #1257, opened 2021-06-18
Deployment to AWS results in Permanent Redirect error


ekasprzyk:
### Describe the bug

Serverless deploy results in:

```
PermanentRedirect: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
    at Request.extractError (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/services/s3.js:837:35)
    at Request.callListeners (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/request.js:688:14)
    at Request.transition (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/root/.serverless/components/registry/npm/@sls-next/serverless-component@1.17.0/node_modules/@sls-next/s3-static-assets/node_modules/aws-sdk/lib/request.js:690:12)
[...]

PR #1262, opened 2021-06-18
fix(s3-static-assets): when initializing s3 client, pass in bucket re…

dphang:
…gion to ensure correct s3 endpoints

Fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/1257


dphang:
GH Actions having some issues, but tested manually and everything seems fine

PR #1263, opened 2021-06-18
fix(s3-static-assets): use regional endpoint for s3 buckets

dphang:


dphang:
Some transient test issues but seems good so far, will re-run the e2e tests once more to ensure no other issues.

PR #1273, opened 2021-06-22
fix(lambda-at-edge, nextjs-component): add option to disable cleaning up .next directory before builds

dphang:


jvarho:
In next.js 11 running `next build` will clean up .next by default: https://nextjs.org/docs/upgrading#cleaning-the-distdir-is-now-a-default

Wouldn't it make more sense to drop the code for cleaning it up completely and let the user just configure it through next.config.js?

dphang:
@jvarho yes, I did see that, but we are still supporting 10.2.3, so for now I have added it. I think there are still quite a few people using Next.js 10.

PR #1287, opened 2021-06-23
fix(lambda-at-edge): improve error logging when s3 fails in image opt…

dphang:
…imizer.

Related: https://github.com/serverless-nextjs/serverless-next.js/issues/1285


dphang:
Flaky SQS tests.

PR #1290, opened 2021-06-24
Fix ISR, make tests retryable

jvarho:
Fixes prerendered dynamic ISR broken in #1238

Also, makes the ISR tests retryable so that failures can be trusted.


jvarho:
Retry works: https://github.com/serverless-nextjs/serverless-next.js/runs/2902210469?check_suite_focus=true#step:10:5550

I'll take a look at the one that fails 5 times. I can't see why it would keep failing there when warm unless there's an actual bug.

Issue #1283, opened 2021-06-23
Implement Domain based locale detection


cmaerz:
**Is your feature request related to a problem? Please describe.**
As mentioned in #1127 domain based locales are missing as a supported feature. In my current project I need the functionality. https://nextjs.org/docs/advanced-features/i18n-routing#domain-routing

**Describe the solution you'd like**
As my research shows the real hostname is forwarded till the `_app.js` file, but the domains confgured in `next.config.js` are not matched properly. I'm not really sure why it's not working. 
Does anyone have hints for me? I would try to solve it myself, but I have no idea where to start.

![image](https://user-images.githubusercontent.com/1479186/123094178-e45ece00-d42c-11eb-80c3-18ecb0679310.png)

dphang:
Yes sorry, this one is not implemented yet. If there is interested I can do it.

I think we will have to update the router around here to add the correct locale based on domain: https://github.com/serverless-nextjs/serverless-next.js/blob/21ac575d9b1a4a4d3244bd08629db78edacaf4a7/packages/libs/core/src/route/index.ts#L153 (similar to how we check hostname in the domain redirect).

dphang:
This was added in the latest alpha versions, I think it should be working now?

PR #1296, opened 2021-06-25
feat: Add domain locale detection (#1283)

cmaerz:
This implements basic domain locale detection now the `config.i18n.domains` section is respected.

I'm open for refactoring suggestions, as I'm not really deep in this project. Just wanted to get it in a pulll request, as I'm on vacation next week.

### DIFF
--- a/packages/libs/core/src/route/locale.ts
+++ b/packages/libs/core/src/route/locale.ts
@@ -1,11 +1,37 @@
-import { Manifest, RoutesManifest } from "../types";
+import { DomainData, Manifest, RoutesManifest } from "../types";
+import { IncomingMessage } from "http";
+
+export const findDomainLocale = (
+  req: IncomingMessage,
+  manifest: RoutesManifest
+): string | boolean => {
+  const domains =
+    manifest.i18n && manifest.i18n.domains ? manifest.i18n.domains : null;
+  if (domains) {
+    const hostHeaders = req.headers.host?.split(",");
+    if (hostHeaders && hostHeaders.length > 0) {
+      const host = hostHeaders[0];
+      const matchedDomain = domains.find(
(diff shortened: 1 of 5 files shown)


dphang (commented):
Great, some unit tests would be good for this (can’t add to e2e tests yet since I don’t have a domain yet)

and also next we should support the domain locales (not just default locale)

packages/libs/core/src/route/locale.ts
```diff
@@ -1,11 +1,37 @@
...
 export function addDefaultLocaleToPath(
   path: string,
-  routesManifest: RoutesManifest
+  routesManifest: RoutesManifest,
+  forceLocale: string | boolean = false
 ): string {
   if (routesManifest.i18n) {
-    const defaultLocale = routesManifest.i18n.defaultLocale;
+    const defaultLocale = forceLocale
+      ? forceLocale
```
dphang, line 33: Use ?? instead?

cmaerz: Changed this. Also rewrote some of that `addDefaultLocaleToPath` function.

cmaerz:
I added some tests for the old `addDefaultLocaleToPath` function and the new `findDomainLocale` (bit hacky here, but tbh i had no better idea to mock that  `IncomingMessage` Type.

dphang:
> I added some tests for the old `addDefaultLocaleToPath` function and the new `findDomainLocale` (bit hacky here, but tbh i had no better idea to mock that `IncomingMessage` Type).

Yeah I think that is ok for now, maybe I can try to see if there are some mocking libraries to help with it.

dphang:
e2e tests passed: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/972240984

dphang (approved):

PR #1310, opened 2021-06-27
Core test refactor

jvarho:
Moves handler tests from lambda-at-edge to core, as discussed in #1144 

Leaves those tests where the behavior in lambda-at-edge differs, e.g. basepath which is used in S3 requests and all the different cases of ISR and fallback.

This should make it easier to refactor things in core and mean less work for apigw tests.


dphang:
Thanks for this! Great work! I think it LGTM, will merge soon

dphang (approved):

PR #1311, opened 2021-06-27
Update README on lambda runtime and ISR

jvarho:
ISR is no longer alpha-only. Also, default runtime has changed and there's a fourth lambda function.

I do wonder about the image lambda: the readme says "Note that the sharp image library is built for NodeJS 12.x" – is that still true and does it mean the default for that should be reverted?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Is there a feature that you want but is not yet supported? Please open a [new is
 - [x] [Custom Headers](https://nextjs.org/docs/api-reference/next.config.js/headers). Caveats: every route should be able to have custom headers except `_next/static/*` and `static/*`, since those cache behaviors do not have Lambda handlers attached to them. You also need to specify the S3 key as the source when redirecting any path mapped to an S3 file (see [PR](https://github.com/serverless-nextjs/serverless-next.js/pull/662) for more details).
 - [x] [Image Optimization](https://nextjs.org/docs/basic-features/image-optimization) Available in the latest 1.19 alpha version. Please try it out and let us know if there are any bugs.
 - [x] [Next.js 10 Localization](https://nextjs.org/blog/next-10). See: https://github.com/serverless-nextjs/serverless-next.js/issues/721 for more details and tips.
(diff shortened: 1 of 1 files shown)


jvarho:
Also, regarding the sharp libraries, the version under sharp_node_modules differs from the package.json dependency.

I'm not sure that's a good idea? Maybe the dependency should be locked if the embedded library continues to be used anyway.

dphang:
> Also, regarding the sharp libraries, the version under sharp_node_modules differs from the package.json dependency.
> 
> I'm not sure that's a good idea? Maybe the dependency should be locked if the embedded library continues to be used anyway.

Yes, right now it's because it is manually pre-generated using the Lambci/lambda docker image to compile the binaries, it works on Node.js 10/12/14 just fine though (the image library is C++ I think so the Node.js runtime doesn't really matter?)

But yes, I will go and prevent dependabot from updating this dependency and manually update/built it again for Node.js 14

dphang (approved):

dphang:
@jvarho actually I forgot that the `sharp` dependency listed in lambda-at-edge package as a dev dependency for unit tests only, since the actual sharp_node_modules included in the lambda are manually generated. Following the commands here: https://sharp.pixelplumbing.com/install#aws-lambda. But anyway I will try to keep it in sync.

dphang:
Looks like the Node14.x Lambci/lambda docker image is not yet created: https://hub.docker.com/r/lambci/lambda/tags?page=1&ordering=last_updated. I'll see if I can build it with the official AWS image instead: https://hub.docker.com/r/amazon/aws-lambda-nodejs/tags?page=1&ordering=last_updated

jvarho:
> @jvarho actually I forgot that the `sharp` dependency listed in lambda-at-edge package as a dev dependency for unit tests only, since the actual sharp_node_modules included in the lambda are manually generated. Following the commands here: https://sharp.pixelplumbing.com/install#aws-lambda. But anyway I will try to keep it in sync.

Ah, in that case I suppose it doesn't really matter.

Something like https://github.com/woss/aws-lambda-layer-sharp would be tidier than manually included packages. But that's a whole different issue.

dphang:
@jvarho yes a layer would be really nice, though sadly Lambda@Edge doesn't support layers yet =(

jvarho:
> @jvarho yes a layer would be really nice, though sadly Lambda@Edge doesn't support layers yet =(

😑 Of course...

dphang:
> > @jvarho yes a layer would be really nice, though sadly Lambda@Edge doesn't support layers yet =(
> 
> 
> 
> 😑 Of course...

On the plus side, you could add it for APIGW-Lambda 😉

PR #1312, opened 2021-06-27
chore(lambda-at-edge): update sharp to 0.28.3 and build on Lambda Nod…

dphang:
…e.js 14.x

* Uses https://hub.docker.com/r/amazon/aws-lambda-nodejs to build `sharp` and embed it into `sharp_node_modules`


dphang:
CodeQL failed but those are dependency issues / not impacting us, so merging it.

Issue #1328, opened 2021-06-29
Rewrites issue resurfacing


patricktyndall:
Hi @dphang after upgrading versions, an old issue (#733) has resurfaced. I was previously on `@sls-next/serverless-component@1.19.0-alpha.45` (no issue), and upgraded to `@sls-next/serverless-component@1.20.0-alpha.28` (issue re-appears).

### Describe the bug
When using a "no-op rewrite" as in the above issue, direct navigation to a content page 404s.

Example:
- we have our SSG page in the folder `pages/[[...content]]/index.js`
- we have the below rewrite config to allow us to access the URLs directly ie `website.com/my-pretty-route`
- when visiting `website.com/my-pretty-route`, this 404s.

Rewrite config:
```
{
	source: '/:path*',
	destination: '/:path*',
},{
	source: '/:path*',
	destination: '/content/:path*',
}
```

### Actual behavior
404 when visiting a route. Client-side navigation to the route works, but navigating directly doesn't.

### Expected behavior
200

### Steps to reproduce
Load a page as in the above

### Versions

- OS/Environment:
- @sls-next/serverless-component version: `@sls-next/serverless-component@1.19.0-alpha.45`
- Next.js version: `^10.2.4-canary.2` (this should be reproducible on a release version -- I need to use this for a bug fix within next)

dphang:
@patricktyndall yes it was removed during refactoring, I forgot why exactly but we probably should add it back...

patricktyndall:
@dphang ok, yes I definitely need this back -- important next compatibility break, and we both went through some trouble to get it fixed.

Do you know of a version number where this was removed? I had to upgrade to support `fallback: blocking`, but it broke my existing pages.

dphang:
I think somewhere in 1.19.1-alpha.6 according to https://github.com/serverless-nextjs/serverless-next.js/commit/dd6576e6276ad85cda0aabe016eb3e3b06460873

patricktyndall:
Ok, thanks for the quick help. For now I'll be using `1.19.1-alpha.5` as it appears to support `fallback: blocking` and also doesn't have the rewrites issue (phew, glad there was a version that works).

Would be grateful if we could fix this in the future!

jvarho:
I did flag that behavioral change in the pull request in question – #1057 – it was incompatible with what Next.js does when given the same configuration. The compatible way to add it back would be to implement #1003

dphang:
[...]

PR #1329, opened 2021-06-29
fix(core): support no-op rewrites

dphang:
Fixes: https://github.com/serverless-nextjs/serverless-next.js/issues/1328

### DIFF
--- a/packages/libs/core/src/route/rewrite.ts
+++ b/packages/libs/core/src/route/rewrite.ts
@@ -1,15 +1,18 @@
 import { addDefaultLocaleToPath } from "./locale";
 import { compileDestination, matchPath } from "../match";
-import { RewriteData, RoutesManifest } from "../types";
+import { PageManifest, RewriteData, RoutesManifest } from "../types";
+import { handlePageReq } from "../route/page";
 
 /**
  * Get the rewrite of the given path, if it exists.
- * @param path
+ * @param uri
+ * @param pageManifest
  * @param routesManifest
  */
 export function getRewritePath(
   uri: string,
-  routesManifest: RoutesManifest
+  routesManifest: RoutesManifest,
+  pageManifest?: PageManifest
 ): string | undefined {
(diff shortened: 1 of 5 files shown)


packages/libs/core/src/route/rewrite.ts
```diff
@@ -27,6 +30,21 @@ export function getRewritePath(
       return;
     }
 
+    // No-op rewrite support for pages: skip to next rewrite if path does not map to existing non-dynamic and dynamic routes
+    if (pageManifest && path === destination) {
+      const url = handlePageReq(
```
jvarho, line 35: What about public files?

jvarho: I suppose those just work because they get checked earlier?

dphang: Sure, we can add check for static files too.

dphang: Ah yes, I believe public files cannot be rewritten unless you use the new objects format:

> Rewrites are applied after checking the filesystem (pages and /public files) and before dynamic routes by default. This behavior can be changed by instead returning an object instead of an array from the rewrites function since v10.1 of Next.js:

dphang: Also now that the code chunking logic is much better from Next.js, we could also start thinking about combining default/API lambdas to simplify rewrites and also support cross-rewrites (between default an API lambdas)

jvarho: Yeah, I was about to mention api routes. But this is surprisingly simple 👍🏽

dphang:
CF distributions seem to be stuck in progress - seems to be AWS issue. Will retry later.

dphang:
e2e passed for most part (failed on external API rewrite due to GH api limits - we could probably try to move to another to avoid this)

PR #1351, opened 2021-07-02
configurable trustedSigners

Qubica:
This little adjustment will allow us to enable and configure trustedSigners per route.


dphang (commented):
LGTM but I think you accidentally added the `packages/libs/lambda-at-edge/tests/integration/with-empty-next-config/fixture/.DS_Store` file?

Qubica:
Ah, my mistake. I've removed the file.

dphang (approved):

PR #1366, opened 2021-07-06
Not found support in SSG fallback

### LINKED ISSUE #1286: CloudFront errors when `notFound: true` returned in `getStaticProps`
I have a page that's using dynamic SSR to generate pages on demand.  When a client requests a path/query that does not have a corresponding api entry, I'd like to return a 404.

I have tried using either the `notFound` or `redirect` return options within `getStaticProps`.  Both of these options work as expected locally, however neither work when deployed to AWS.

For example, given the following page:

`pages/items/[name].tsx`

```typescript
import React from 'react';
import { GetStaticProps, GetStaticPaths } from 'next';

interface Props {
  name: string | undefined;
}

export const getStaticProps: GetStaticProps = async (context) => {

  //Paths/conditions provided for example purposes
  const notFound = context.params?.name === 'not-found';
  const redirect = context.params?.name === 'redirect';

  if (notFound) {
    return {
      notFound: true
    };
  }

  if (redirect) {
    return {
      redirect: {
        destination: '/404',
        permanent: false
      }
    };
  }

  const result = {
    props: {
      name: context.params?.name
    }
  };

  return result;
};

export const getStaticPaths: GetStaticPaths = async () => ({
  paths: [],
  fallback: 'blocking'
});

const Item = (props: Props) => {
  const { name } = props;
  return (
    <h1>
      {`Item: ${name}`}
    </h1>
  );
};

export default Item;
```

Locally, both `/item/not-found` and `/item/redirect`, respond with a 404.  However when deployed to AWS the following occurs:

-  `/item/not-found` - CloudFront responds with a 200, but with the content `AccessDenied`.  In addition to this, a zero byte `not-found.html` object is created in S3.
- `item/redirect` - CloudFront responds with a 503

I'd like to be able to handle this "not found" scenario, where the client receives a 404, and no object is generated in S3.

Versions used:
- @sls-next/serverless-component alpha/latest
- NextJs: 10.0.8

jvarho:
Fixes #1286 and adds support for prerendered notFoundRoutes.

Does not cache notFound: true results in S3. It seems that next.js server does not cache them either.

### DIFF
--- a/packages/e2e-tests/next-app-dynamic-routes/pages/optional-catch-all-ssg-with-fallback/[[...catch]].tsx
+++ b/packages/e2e-tests/next-app-dynamic-routes/pages/optional-catch-all-ssg-with-fallback/[[...catch]].tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { GetServerSidePropsContext } from "next";
+import { GetServerSidePropsContext, GetStaticPropsResult } from "next";
 
 type OptionalCatchAllPageProps = {
   name: string;
@@ -22,8 +22,14 @@ export default function OptionalCatchAllPage(
 
 export async function getStaticProps(
   ctx: GetServerSidePropsContext
-): Promise<{ props: OptionalCatchAllPageProps }> {
+): Promise<GetStaticPropsResult<OptionalCatchAllPageProps>> {
   const catchAll = ((ctx.params?.catch as string[]) ?? []).join("/");
+
+  if (catchAll === "not-found") {
+    return {
+      notFound: true
+    }
+  }
 
   return {
     props: { name: "serverless-next.js", catch: catchAll }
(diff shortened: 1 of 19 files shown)


packages/e2e-tests/next-app-dynamic-routes/pages/optional-catch-all-ssg-with-fallback/[[...catch]].tsx
```diff
@@ -22,9 +22,15 @@ export default function OptionalCatchAllPage(
...
 export async function getStaticProps(
   ctx: GetServerSidePropsContext
-): Promise<{ props: OptionalCatchAllPageProps }> {
+): Promise<GetStaticPropsResult<OptionalCatchAllPageProps>> {
   const catchAll = ((ctx.params?.catch as string[]) ?? []).join("/");
 
+  if (catchAll === "not-found") {
+    return {
+      notFound: true
+    }
```
Comment (dphang) [inline packages/e2e-tests/next-app-dynamic-routes/pages/optional-catch-all-ssg-with-fallback/[[...catch]].tsx:31]: Codacy has a fix for the issue: [Insert `;`](https://app.codacy.com/gh/serverless-nextjs/serverless-next.js/pullRequest?prid=7710201)
```suggestion
    };
```

jvarho:
This was slightly more involved than I hoped.

I had to change the cloudfront compat code to retain origin response headers (specifically for transfer-encoding, but potentially others). It may have already been buggy with at least non-static _error in response handler, potentially causing 502 errors.

Also, it seems next.js really wants a json response for the data route on the fallback page. I wonder if the 404 response should be changed more generally...

dphang (approved):

PR #1448, opened 2021-07-26
Add support for custom headers in custom origins

Xaroth:
This mostly mimics https://github.com/serverless-components/aws-cloudfront/pull/20 , but then converted to ts.

This adds the 'headers' option for origins, which allows one to set custom headers to be set with these origin requests.


dphang:
Thanks, can you just make sure you are formatting the file according to prettier/eslint rules, so the codebase is consistent in style? Looks like the codacy analysis failed because of those?

Xaroth:
Will work on this when I get into the office. I'm not very familiar with the test suite being used (I normally don't develop much js/ts), so this might take me a bit 🙂

Xaroth:
@dphang Both points should be fixed. Ironically, the linting issues I already had staged, just not committed, which was a bad on my part. The tests failing was a failure at my end when converting it to ts.

Xaroth:
Hm, the test I see failing seems to be related to the fix you just merged in master, so is not related to this patch.

dphang:
> Hm, the test I see failing seems to be related to the fix you just merged in master, so is not related to this patch.

Yup, I was trying to upgrade a dependency but it failed the build, will revert the change.

Xaroth:
I see the e2e runs generated some errors, but I don't think those errors are related to the code, more to the service being used for that test.

dphang:
> I see the e2e runs generated some errors, but I don't think those errors are related to the code, more to the service being used for that test.

Yea, it's because it was hitting the homepage of https://jsonplaceholder.typicode.com/ for some tests and doing some comparison for rewrites, but because it's HTML and they recently added Cloudflare browser insights, there is some UUID generated that differs on each page load. Fixed it in another PR.

dphang (approved):

Xaroth:
Ah, that does explain a lot then, yeah 🙂

PR #1487, opened 2021-08-02
fix(lambda-at-edge): handle 404 status code returned by S3 in origin …

dphang:
…response handler

Related: https://github.com/serverless-nextjs/serverless-next.js/issues/1423


stephanie-pavemint:
Really appreciate your effort here!

PR #1509, opened 2021-08-06
feat(aws-sqs): support sqs tags/naming, convert aws-sqs to typescript

dphang:


dphang:
Will recreate without the TS conversion for now as it seems to be causing issues

PR #1522, opened 2021-08-09
Add example of nextjs-cdk-construct

hideokamoto:
I've made an example project of the `@sls-next/cdk-construct`.

## How to use (npm)

```bash
$ cd /PATH/TO/EXAMPLE/basic
$ npm install
$ npm run dev
```

### Deploy

```bash
$ npm run build
$ npm run cdk deploy
```

## How to use (npm)

```bash
$ cd /PATH/TO/EXAMPLE/basic
$ yarn
$ yarn dev
```

### Deploy

```bash
$ yarn build
$ yarn cdk deploy
```


dphang:
Thanks! You might want to check the codacy static analysis for some suggested fixes to styling etc, but since this is an example it might be ok

hideokamoto:
Thanks.
Is it possible to check the Static Code Analytics in my local env?
Or should I create a new account for this service?

dphang:
@hideokamoto I mean the codacy check that was run on the PR, I think it should be public and accessible without an account, you can see details here: https://app.codacy.com/gh/serverless-nextjs/serverless-next.js/pullRequest?prid=7883792. It did detect a few issues which I think you an try to fix if they make sense. Thanks!

PR #1543, opened 2021-08-13
feat(aws-sqs): support sqs tags/naming

dphang:


dphang:
problem with the existing test apps, need to take a look later - tested locally and things seem fine (and this doesn't affect runtime stuff anyway), so I'm merging this.

PR #1562, opened 2021-08-19
fix(aws-cloudfront): set tags when creating distribution

dphang:


dphang:
Tested locally, ignoring the codacy issues for now since they are minor and will take a bit of work to refactor the aws-cloudfront package (originally was in JS)

PR #1563, opened 2021-08-19
Fix ISR issue with `index.ts`

J3tto:
For the root path `/` the `s3StorePage` was not correctly storing to the `index.html` file and instead was storing as `.html`

Added an extra regex `replace()` to deal with the root path. 

#1098

### DIFF
--- a/packages/libs/lambda-at-edge/src/s3/s3StorePage.ts
+++ b/packages/libs/lambda-at-edge/src/s3/s3StorePage.ts
@@ -30,6 +30,7 @@ export const s3StorePage = async (
     ? `${options.basePath.replace(/^\//, "")}/`
     : "";
   const baseKey = options.uri
+    .replace(/^\/$/, "index")
     .replace(/^\//, "")
     .replace(/\.(json|html)$/, "")
     .replace(/^_next\/data\/[^\/]*\//, "");


dphang:
Thanks! I'll add a test for this as well so it doesn't break in the future.

dphang:
Hm, for some reason CI is failing on `ENOENT: no such file or directory, lstat '/home/runner/.cache/yarn/v6/npm-@aws-sdk-client-sqs-3.27.0-a1d982019362cdd1c39901092c7a944278966627-integrity/node_modules/@aws-sdk/client-sqs/CHANGELOG.md'`. Seems like a spurious error as it is working locally. Maybe try to rebase to the latest master (I don't have permissions to do that for you) so I can rerun the CI?

J3tto:
Looks like the CI might be already broken/erroring for  for master. Tried a rebase of the branch against master but already all up to date. 

There are however a few uncommitted changes to the `yarn.lock` (following a yarn install) I have added this to the branch in a bit to get it to pass CI build.

J3tto:
Updated, Hopefully that works

dphang (approved):

dphang:
Thanks!

Issue #605, opened 2020-09-15
next build fails silently 


michaelwindell26:
**Describe the bug**
When running `npx serverless --debug` if the `next build` process fails with an error it is not displayed in the output and the deployment hangs forever.

**To Reproduce**

1. Create a reproducible error in `pages/index.js` (in my case an `import` statement was incorrect)
2. Run `npx serverless --debug`
3. Output displays "appName Deploying" without the error message from `next build` displayed
4. Run `next build`
5. Output displays error message

**Expected behavior**
Output from `next build` should be forwarded to output from `npx serverless --debug`

**Desktop (please complete the following information):**

- OS: [e.g. WSL2 on Windows 10]

michaelwindell26:
For now I will setup our CI / CD pipeline to manually run `next build` so we can at least see an error if it occurs...

dphang:
Couldn't reproduce on Mac OS Catalina and latest alpha (1.17 alpha-13):

```bash
/usr/local/bin/node /usr/local/Cellar/yarn/1.22.4/libexec/bin/yarn.js run serverless
yarn run v1.22.4
$ serverless

  error:
  Error: Command failed with exit code 1: node_modules/.bin/next build
Failed to compile.

./pages/ssr-page.tsx:2:33
Type error: Cannot find module 'next2' or its corresponding type declarations.

  1 | import React from "react";
> 2 | import { NextPageContext } from "next2";
    |                                 ^
  3 | 
  4 | type SSRPageProps = {
  5 |   name: string;
info  - Creating an optimized production build...
    at makeError (/Users/danielphang/IdeaProjects/serverless-next.js/packages/libs/lambda-at-edge/node_modules/execa/lib/error.js:59:11)
    at handlePromise (/Users/danielphang/IdeaProjects/serverless-next.js/packages/libs/lambda-at-edge/node_modules/execa/index.js:114:26)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  shortMessage: 'Command failed with exit code 1: node_modules/.bin/next build',
  command: 'node_modules/.bin/next build',
  exitCode: 1,
  signal: undefined,
  signalDescription: undefined,
  stdout: 'info  - Creating an optimized production build...',
  stderr: 'Failed to compile.\n' +
    '\n' +
    './pages/ssr-page.tsx:2:33\n' +
    "Type error: Cannot find module 'next2' or its corresponding type declarations.\n" +
    '\n' +
[...]

PR #1658, opened 2021-09-09
feat(nextjs-component): introduce serverless-patched binary

dphang:
* Recognizes CI environments so that multiple "Deploying ..." messages are not displayed
* Fixes next silent build failure from here https://github.com/serverless-nextjs/serverless-next.js/issues/605

Unfortunately `patch-package` doesn't support multiple patches (so all of it is in one file), but hopefully we don't have many other major fixes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@types/fs-extra": "^9.0.1",
     "@types/jest": "^27.0.1",
     "@types/lodash": "^4.14.172",
-    "@types/node": "^16.7.13",
+    "@types/node": "^16.9.0",
     "@types/node-fetch": "^2.5.10",
     "@types/webpack": "^5.28.0",
     "@typescript-eslint/eslint-plugin": "^4.31.0",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "@types/webpack": "^5.28.0",
     "@typescript-eslint/eslint-plugin": "^4.31.0",
     "@typescript-eslint/parser": "^4.31.0",
-    "aws-sdk": "^2.984.0",
+    "aws-sdk": "^2.985.0",
     "eslint": "^7.3.1",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -95,7 +95,8 @@
       "/scripts/",
       "babel.config.js",
       "jest.config.js",
-      "<rootDir>/packages/serverless-components/aws-s3"
+      "<rootDir>/packages/serverless-components/aws-s3",
+      "<rootDir>/packages/serverless-components/nextjs-component/bin"
     ],
     "watchPathIgnorePatterns": [
       "/fixture/",
--- a/packages/e2e-tests/next-app/package.json
+++ b/packages/e2e-tests/next-app/package.json
@@ -6,7 +6,7 @@
     "dev": "next",
     "build": "next build",
     "start": "next start -p $PORT",
-    "serverless": "serverless",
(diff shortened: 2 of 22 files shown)


dphang:
Hm, so the patching should happen in `postinstall` step which is after the package is pulled but before `serverless` is actually executed. 

Are you running it with `npx serverless-patched` or did you install `@sls-next/serverless-component` and try to run the local `serverless-patched` binary? While testing, the local `serverless-patched` was a bit wonky because of how `node_modules` are linked...sometimes found that `node_modules/@serverless/cli` are not there. Maybe I need to add `@serverless/cli` as a hard dependency instead of just `serverless`?

You could also try `npx --package @sls-next/serverless-component serverless-patched` which will force `npx` to use a temp package, it might take a couple seconds extra to download the component though.

csandman:
I saw your original issue after commenting, and the need to run the command with `serverless-patched` instead.  The way I have it, `@sls-next/serverless-component` is not installed anywhere manually, so it is only pulled during the run of the `serverless` command.  This is my codebuild config:

```yml
phases:
  install:
    runtime-versions:
      nodejs: 14.x
    commands:
      - "touch .npmignore"
      - "npm install -g serverless"
  pre_build:
    commands:
      - "npm install"
  build:
    commands:
      - "serverless --debug"
```

And I have nothing to do with serverless in my `package.json`, relying on having `component: "@sls-next/serverless-component@3.4.0-alpha.3" in my `serverless.yml`. Would I need to have `@serverless/cli` and/or `@sls-next/serverless-component` in my dependencies to make it work?

dphang:
I have fixed the directory resolution and moved it into a separate package. The new command is: `npx @sls-next/serverless-patched --debug`. Just tested in GitHub Actions (with the e2e tests) and it seems to work fine, let me know if there are issues.

stan-sack:
still getting exit code 0 for failed deployment with `3.4.0-alpha.9`. Does that have the change?

stan-sack:
oh right i have to use patched serverless

stan-sack:
it might work. it exits with an error code but the debug logs dont work with the patched version:
```
$ npx @sls-next/serverless-patched --debug --stage=staging
npm WARN exec The following package was not found and will be installed: @sls-next/serverless-patched
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated request-promise-native@1.0.9: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142
npm WARN deprecated querystring@0.2.1: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated uuid@3.3.2: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
Cleaning up file based variables
00:00
ERROR: Job failed: exit code 1
```

stan-sack:
yeah i wasnt able to get this running. when i swapped the following in my CI script it broke with the above error
```
-    - npx @sls-next/serverless-patched --cache .npm --debug --stage=staging
+   - npx serverless --cache .npm --debug --stage=staging
```

I have custom `serverless.js` file in my root thats checking the stage flag (as recommended in your documentation). maybe thats breaking it?

PR #1786, opened 2021-10-04
feat: Handle Redirects in uncached SSG Requests

cmaerz:


packages/libs/core/src/defaultHandler.ts
```diff
@@ -258,6 +255,21 @@ const staticRequest = async (
 
   // This is a fallback route that should be stored in object store before returning it
   const { renderOpts, html } = fallbackRoute;
+  // Check if response is a redirect
+  if (
+    typeof renderOpts.pageData !== "undefined" &&
+    typeof renderOpts.pageData.pageProps !== "undefined" &&
+    typeof renderOpts.pageData.pageProps.__N_REDIRECT !== "undefined"
+  ) {
+    const statusCode = renderOpts.pageData.pageProps.__N_REDIRECT_STATUS;
```
dphang, line 264: Thanks! Can you please reuse the code in here to generate the response: https://github.com/serverless-nextjs/serverless-next.js/blob/14d410d4a71c2ea4f5b4bbea4bed9f1c0627b8fb/packages/libs/core/src/handle/redirect.ts#L4-L4

and https://github.com/serverless-nextjs/serverless-next.js/blob/14d410d4a71c2ea4f5b4bbea4bed9f1c0627b8fb/packages/libs/core/src/route/redirect.ts#L17-L17

It will be more reusable and would prevent duplicate code in both the v1 and v2 handlers.

dphang: Also you could probably move this into a common function in `core` (maybe in `route/redirect.ts` file)

cmaerz: Maybe you can check my solution. I'm not 100% sure if it's the right way.
I didn't put it into a common function, as in perspective the v2 Handlers are the way to go (duplicate code) and I'm not that big of a fan of moving 2 Lines in a 4-6 Line function. :D

PR #1844, opened 2021-10-15
fix(aws-cloudfront): handle bucket names with dots correctly

tomi:
As per [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html), bucket names can contain dots. This was addressed in [this PR](https://github.com/serverless-nextjs/serverless-next.js/pull/596), but it was never merged. Hopefully this can be merged.


dphang:
Thanks!

Issue #1620, opened 2021-09-02
CDK Construct - S3 Deployment stage missing `basePath`


ghsyeung:
If `basePath` is set in `next.config.js`, the build would put static assets within `${sererlessBuildOutDir}/assets/${basePath}`.

However, in the CDK construct (`NextJSLambdaEdge`), it tries to read the assets from `${sererlessBuildOutDir}/assets` without including the `basePath`

https://github.com/serverless-nextjs/serverless-next.js/blob/145c16d9e163aca7b5f6a753f60404363c842b8f/packages/serverless-components/nextjs-cdk-construct/src/index.ts#L404

The function returns with an empty `assets` object leading to none of the assets being uploaded.

Similar code in s3-static-assets handle this correctly

https://github.com/serverless-nextjs/serverless-next.js/blob/145c16d9e163aca7b5f6a753f60404363c842b8f/packages/libs/s3-static-assets/src/index.ts#L67

The fix looks relatively straight forward and it works for my use case.

```typescript
const { basePath } = this.routesManifest || {};
const normalizedBasePath = basePath && basePath.length > 0? basePath.slice(1) : "";
const assets = readAssetsDirectory({ assetsDirectory: path.join(assetsDirectory, normalizedBasePath) }); 
```

dphang:
Thanks @ghsyeung. Feel free to raise a PR to fix it and I'll merge it.

PR #2015, opened 2021-11-05
Add URL normalise to image optimiser URL

mathvbarone:
If Basepath is set, it needs to be removed from the query

 Not normalised -> error 403
 `url: '<base-path>/assets/images/logo.svg'`
 
Normalised -> 200
url: `'/assets/images/logo.svg'`

### DIFF
--- a/packages/libs/core/src/images/imageOptimizer.ts
+++ b/packages/libs/core/src/images/imageOptimizer.ts
@@ -130,6 +130,7 @@ export async function imageOptimizer(
   if (url.startsWith(basePath)) {
     url = url.slice(basePath.length);
   }
+  
   if (url.startsWith("/")) {
     href = url;
     isAbsolute = false;
--- a/packages/libs/core/src/images/imageOptimizer.ts
+++ b/packages/libs/core/src/images/imageOptimizer.ts
@@ -116,7 +116,7 @@ export async function imageOptimizer(
 
   let isAbsolute: boolean;
 
-  // Ensure if Basepath is in the URL, otherwise, a 400 is triggered (same behaviour as Nextjs)
+  // Ensure that Basepath is in the URL, otherwise, a 400 is triggered (same behaviour as Nextjs)
   if (basePath !== "/" && !url.startsWith(basePath)) {
     res.statusCode = 400;
     res.end('"basepath" set but not added to the URL');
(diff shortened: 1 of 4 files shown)


packages/libs/core/src/images/imageOptimizer.ts
```diff
@@ -115,6 +116,20 @@ export async function imageOptimizer(
...
+   * /<base-path>/_next/image?url=%2F<base-path>%2Fassets%2Fimages%2Flogo.svg&w=256&q=75
+   * 
+   * Normalised -> 200
+   * /<base-path>/_next/image?url=%2Fassets%2Fimages%2Flogo.svg&w=256&q=75
+   * 
+   * 
+   */
+  if (url.startsWith(basePath)) {
+    url = url.slice(basePath.length);
+  }
```
ivanbabenko, line 132: Maybe add blank line after this condition for readability

mathvbarone: Done! 😛

dphang:
Thanks, though it looks like the actual Next.js behavior with local `next start` is that basepath should be supplied to the URL when it's set in config, otherwise if no basepath there is a 400 response returned. So I guess in another PR you should actually check if basepath is passed into the relative URLs as well and throw a 400 response if not passed (to be fully in parity with Next.js behavior).

Not sure how it is related to #1620? I think that one is a path issue in uploading assets via the CDK construct.

mathvbarone:
> Thanks, though it looks like the actual Next.js behavior with local `next start` is that basepath should be supplied to the URL when it's set in config, otherwise if no basepath there is a 400 response returned. So I guess in another PR you should actually check if basepath is passed into the relative URLs as well and throw a 400 response if not passed (to be fully in parity with Next.js behavior).
> 
> Not sure how it is related to #1620? I think that one is a path issue in uploading assets via the CDK construct.

Hey, @dphang I added the verification as requested, can you please take another look?

I thought #1620 was also related to the addition of basepath on the query, you can disregard that (I'll edit the description of the PR)

Thanks!

packages/libs/core/src/images/imageOptimizer.ts
```diff
@@ -116,7 +133,14 @@ export async function imageOptimizer(
   let isAbsolute: boolean;
 
   if (url.startsWith("/")) {
-    href = url;
+    // Ensure that Basepath is in the URL, otherwise, a 400 is triggered (same behaviour as Nextjs)
+    if (basePath !== "/" && !url.startsWith(basePath)) {
```
dphang, line 137: Ideally should add tests for basepath (right now) but you can add in another PR

dphang (approved):

dphang:
e2e tests passed: https://github.com/serverless-nextjs/serverless-next.js/runs/4145408350?check_suite_focus=true

PR #2024, opened 2021-11-08
fix(core): render static pages for all http methods, for options meth…

dphang:
…ods return allowed methods. This will be in parity with Next.js server as it also renders static pages no matter the request method. Note: only applies if S3 pages are rendered through the Lambda, as for CloudFront -> s3 origin, http method cannot be modified (and we don't want to do it in origin response handler as it will just be inefficient).


dphang:
Transient test failure

Issue #2041, opened 2021-11-15
The ap-east-1 location constraint is incompatible for the region specific endpoint this request was sent to.


hotlong:
### Issue Summary

No code changed, but deploy failed in this week, It's ok 10 days ago. I wonder if it is because of bucketRegion
```
bucketRegion: ap-east-1
```

### Actual behavior

deploy failed

### Expected behavior

deploy success 

### Steps to reproduce

- deploy by github action https://github.com/steedos/steedos.com/blob/master/.github/workflows/dev.yml
- serverless config yml: https://github.com/steedos/steedos.com/blob/dev/serverless-dev.yml

### Screenshots/Code/Configuration/Logs

Full log at https://github.com/steedos/steedos.com/runs/4208662248?check_suite_focus=true
```
DEBUG ─ Starting deployment of CloudFront distribution to the us-east-1 region.
DEBUG ─ Updating CloudFront distribution of ID E1ONY9MU8PSX1R.
error:
  IllegalLocationConstraintException: The ap-east-1 location constraint is incompatible for the region specific endpoint this request was sent to.
    at Request.extractError (/Users/steedos/.serverless/components/registry/npm/@sls-next/serverless-component@3.6.0/node_modules/@sls-next/aws-cloudfront/node_modules/aws-sdk/lib/services/s3.js:699:35)
    at Request.callListeners (/Users/steedos/.serverless/components/registry/npm/@sls-next/serverless-component@3.6.0/node_modules/@sls-next/aws-cloudfront/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/steedos/.serverless/components/registry/npm/@sls-next/serverless-component@3.6.0/node_modules/@sls-next/aws-cloudfront/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/steedos/.serverless/components/registry/npm/@sls-next/serverless-component@3.6.0/node_modules/@sls-next/aws-cloudfront/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/Users/steedos/.serverless/components/registry/npm/@sls-next/serverless-component@3.6.0/node_modules/@sls-next/aws-cloudfront/node_modules/aws-sdk/lib/request.js:22:10)
[...]

PR #2042, opened 2021-11-15
fix(aws-cloudfront, nextjs-component): specify s3 client's region in …

dphang:
…aws-cloudfront the same as the bucket region

Should fix: https://github.com/serverless-nextjs/serverless-next.js/issues/2041


dphang:
Failure due to accidental concurrent run of same e2e tests, but considering some passed and this is a deployment logic change, it should be fine.

Issue #2048, opened 2021-11-17
ISR Regeneration fails with cdk


mathisobadia:
### Issue Summary
When using ISR with static regeneration and the cdk construct, the pages are built the first time correctly, but they are never updated.
The problem seems to be that the regeneration function times out after 30 seconds.

### Actual behavior
Pages are not regenerated and regeneration function fails due to time out. Initial page creation is very quick so the issue should not be that the page takes more than 30 seconds to regenerate.

### Expected behavior
pages should be regenerated

### Steps to reproduce
I have a pretty specific setup and I haven't yet created a repro case. I'm wondering if that might be a known issue, and to not take too much time of the maintainers I would like to try to debug it myself but I don't know how. Is there a way to modify the regeneration function in my setup to add logs to it so I can find out at what step the regeneration function times out? I'm also wondering if there might be a permission issue that is not showing in the logs

### Screenshots/Code/Configuration/Logs
Currently the logs for the regeneration function look like this so it is not very helpfull (that's a snippet but this is repeated over and over again pretty much all the time). 
```
2021-11-17T00:28:54.973+01:00 | START RequestId: 10b72789-d8d6-559f-a3dd-7d87982216b6 Version: $LATEST
  | 2021-11-17T00:29:25.005+01:00 | END RequestId: 10b72789-d8d6-559f-a3dd-7d87982216b6
  | 2021-11-17T00:29:25.005+01:00 | REPORT RequestId: 10b72789-d8d6-559f-a3dd-7d87982216b6 Duration: 30028.59 ms Billed Duration: 30000 ms Memory Size: 128 MB Max Memory Used: 128 MB
  | 2021-11-17T00:29:25.005+01:00 | 2021-11-16T23:29:25.004Z 10b72789-d8d6-559f-a3dd-7d87982216b6 Task timed out after 30.03 seconds
  | 2021-11-17T00:29:25.831+01:00 | START RequestId: 8f927a92-4632-543f-8a90-209c07208843 Version: $LATEST
  | 2021-11-17T00:29:55.867+01:00 | END RequestId: 8f927a92-4632-543f-8a90-209c07208843
  | 2021-11-17T00:29:55.867+01:00 | REPORT RequestId: 8f927a92-4632-543f-8a90-209c07208843 Duration: 30031.70 ms Billed Duration: 30000 ms Memory Size: 128 MB Max Memory Used: 128 MB
[...]

PR #2050, opened 2021-11-17
feat(nextjs-cdk-construct): add runtime options to regeneration

mathisobadia:
fixes issue #2048  
The regeneration function was the only function where the runtime could not be modified through setttings prop
This can cause the regeneration lambda function to time out an fail
The pull request adds the option to configure the regeneration runtime (memory, timeout, role, and node version)

as a note I have not modified the nomenclature for this function but it is the only lambda handler that is named as Function instead of Lambda for the others (nextApiLambda, nextDefaultLambda, nextImageLambda) but the regeneration lambda is named RegenerationFunction. I can modify that also if needed.

### DIFF
--- a/packages/serverless-components/nextjs-cdk-construct/src/index.ts
+++ b/packages/serverless-components/nextjs-cdk-construct/src/index.ts
@@ -136,7 +136,7 @@ export class NextJSLambdaEdge extends cdk.Construct {
           role: this.edgeLambdaRole,
           runtime:
             toLambdaOption("regenerationLambda", props.runtime) ??
-            lambda.Runtime.NODEJS_12_X,
+            lambda.Runtime.NODEJS_14_X,
           memorySize: toLambdaOption("regenerationLambda", props.memory) ?? 512,
           timeout:
             toLambdaOption("regenerationLambda", props.timeout) ?? Duration.seconds(30)
--- a/packages/serverless-components/nextjs-cdk-construct/src/index.ts
+++ b/packages/serverless-components/nextjs-cdk-construct/src/index.ts
@@ -133,11 +133,10 @@ export class NextJSLambdaEdge extends cdk.Construct {
           code: lambda.Code.fromAsset(
             path.join(this.props.serverlessBuildOutDir, "regeneration-lambda")
           ),
-          role: this.edgeLambdaRole,
           runtime:
             toLambdaOption("regenerationLambda", props.runtime) ??
             lambda.Runtime.NODEJS_14_X,
-          memorySize: toLambdaOption("regenerationLambda", props.memory) ?? 512,
+          memorySize: toLambdaOption("regenerationLambda", props.memory) ?? undefined,
           timeout:
             toLambdaOption("regenerationLambda", props.timeout) ?? Duration.seconds(30)
         }
--- a/packages/serverless-components/nextjs-cdk-construct/src/index.ts
+++ b/packages/serverless-components/nextjs-cdk-construct/src/index.ts
@@ -89,6 +89,20 @@ export class NextJSLambdaEdge extends cdk.Construct {
       ...(props.s3Props || {})
     });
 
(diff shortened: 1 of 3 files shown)


packages/serverless-components/nextjs-cdk-construct/src/index.ts
```diff
@@ -116,11 +130,16 @@ export class NextJSLambdaEdge extends cdk.Construct {
...
-          runtime: lambda.Runtime.NODEJS_14_X,
-          timeout: Duration.seconds(30),
           code: lambda.Code.fromAsset(
             path.join(this.props.serverlessBuildOutDir, "regeneration-lambda")
-          )
+          ),
+          role: this.edgeLambdaRole,
+          runtime:
+            toLambdaOption("regenerationLambda", props.runtime) ??
+            lambda.Runtime.NODEJS_12_X,
```
dmsolutionz, line 139: This should remain NODEJS_14_X?

mathisobadia: I'm not 100% sure why it was hardcoded as NODEJS_14_X but I think it's unintuitive that the regeneration function is the only one with a fixed nodejs version. This might result in some weird bugs if it is the regeneration function uses a different runtime than the default function because they use the same js page bundles. This line I added is the same as for the other functions qnd it makes sure that all functions use the same nodejs version except if the user really wants to set something different throught props

afenton90 (commented):
Looks good aside from the change mentioned by @dmsolutionz

packages/serverless-components/nextjs-cdk-construct/src/index.ts
mathisobadia: Don't we want the regeneration lamda have the same role as the other functions? I know in my case I need this because the role grants access to dynamo to get the date necessary for generating a page

mathisobadia:
> Looks good aside from the change mentioned by @dmsolutionz

Thank you for the feedback! I can change my pull request but I'm wondering why this function should be the only one with fixed node 14 runtime while all the other ones can be set with props? it doesn't make sense to keep the props as they are with the possibility to set the runtime version of the regeneration function since we now have 

```js
  /**
   * Lambda runtimes(s)
   */
  runtime?: LambdaOption<Runtime>;
```

and 

```js
export type LambdaOption<T> =
  | T
  | {
      defaultLambda?: T;
      apiLambda?: T;
      imageLambda?: T;
      regenerationLambda?: T;
    };
```

this means that the user can use a prop like 

```js
runtime: {
   regenerationLamda: Runtime.NODEJS_12_X
}
```

this would be valid typescript prop but it would not change the runtime of the function. This would be highly deceptive.
In my opinion if we keep the line as NODEJS_14_X then we also need to update the props so that it reflects the fact that the runtime is the only option that cannot be changed for the regeneration function. 

As an alternative, so that we don't change the runtime of current users unexpectedly we could keep the option to set the runtime but keep as a default the NODEJS_14_X instead of the default NODEJS_12_X that we have for all the other lambdas

packages/serverless-components/nextjs-cdk-construct/src/index.ts
afenton90: Sorry @mathisobadia to be clear I think what @dmsolutionz is saying is that the "default should remain as NODEJS_14_X" rather than it shouldn't be editable. If you change to `NODEJS_12_X` as the default that's a breaking change.
Having it as editable like the others is exactly the desired outcome here.

mathisobadia: Oh yeah that makes perfect sense! I've updated my pull request accordingly.

mathisobadia:
I just noticed that the tests fail because of the snapshots I've updated the pull request to fix the put back the previous default memory (undefined) and remove the unnecessary role that I added by mistake

dphang (approved):

PR #2124, opened 2021-11-27
🚧 Next.js 12 Output File Tracing

iiroj:
This PR implements the Next.js 12 [Output File Tracing](https://nextjs.org/docs/advanced-features/output-file-tracing), meaning that `@sls-next/serverless-component` will no longer set any `target` while building Next.js, and will instead use the generated `*.nft.json` files to copy required files into the Lambda@Edge deployment.

To use the new feature, one should:

- update to Next.js 12
- make sure there is no `target` in the Next.js configuration (it's now deprecated)
- set the `experimentalOutputFileTracing: true` in the `serverless.yml` inputs for `@sls-next/serverless-component`


iiroj:
It is now working, and I successfully deployed [iiro.fi](https://iiro.fi) with this! 🚀

iiroj:
I'm probably copying too many files. Let me try to leave out the server-related ones since this is for Lambdas anyway.

dphang:
Nice, I'll take a look this week. Is there a lot of perf difference (e.g cold starts) with the new output file tracing?

BTW, you can also use `next-app-experimental` for testing new features, although adding a new e2e test app is also fine for now (but in the future I would like to figure out a way to reduce duplicate apps/test code, maybe a single app with multiple variants of tests, e.g windows, with output file tracing etc.)

iiroj:
@dphang I haven't really dug too deep with this yet, just wanted to get it working first. It seems to generate a bit larger lambdas than the current default (730KB vs 1.01 MB), so I assume there's unused files.

I also tried running everything through esbuild to get a single handler file, and got it down to 380 KB. Maybe that's another PR 🙌

dphang:
> @dphang I haven't really dug too deep with this yet, just wanted to get it working first. It seems to generate a bit larger lambdas than the current default (730KB vs 1.01 MB), so I assume there's unused files.
> 
> I also tried running everything through esbuild to get a single handler file, and got it down to 380 KB. Maybe that's another PR 🙌

No worries, yeah I was wondering because in the past in my experience, the `@vercel/nft` package didn't seem to work well with packages like material UI (e.g if you only needed one component it still traced everything else and it caused significantly higher cold start times), wondering if it works better now. Definitely using esbuild at build-time might also help (and it will help other optimizations like removing unused core code, e.g if users don't use redirects or rewrites at all), although I do want to investigate if we can replace rollup with esbuild for pre-bundling as well so we have more consistency (it only affects development build times for this package. I did try it and it works but somehow the bundle is bigger than rollup's.).

dphang:
Also could you push a no-op commit again to trigger the CircleCI builds? I'm recently trying out circleci for doing some builds so we can run more concurrent builds in preparation for adding more platforms in the near future (e.g Lambda).

packages/e2e-tests/next-app-using-experimental-output-file-tracing/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "next-app-using-serverless-trace",
```
dphang, line 2: Can you also update all instances of `next-app-using-serverless-trace` to `next-app-using-experimental-output-file-tracing` and add the app to here https://github.com/serverless-nextjs/serverless-next.js/blob/master/.github/workflows/e2e-tests.yml#L90

dphang:
Also seems the new files that were added are not covered by unit tests? please add coverage if possible

iiroj:
@dphang thanks, will do. Do you think there should be some validation or help messages to confirm the `target` option should not be set (or be `"server"`)?

dphang:
> @dphang thanks, will do. Do you think there should be some validation or help messages to confirm the `target` option should not be set (or be `"server"`)?

Yeah I think validation makes sense, otherwise there is no point in setting the option?

iiroj:
@dphang please note that I will close this PR and re-open it from a new under my company GitHub, because I will spend some company time polishing it up! I'll do this next week.

PR #2169, opened 2021-12-03
Experimental Next.js 12 Output File Tracing

iiroj:
This PR implements the Next.js 12 [Output File Tracing](https://nextjs.org/docs/advanced-features/output-file-tracing), meaning that `@sls-next/serverless-component` will no longer set any `target` while building Next.js, and will instead use the generated `*.nft.json` files to copy required files into the Lambda@Edge deployment.

To use the new feature, one should:

- update to Next.js 12
- make sure there is no `target` in the Next.js configuration (it's now deprecated)
- set the `build.outputFileTracing: true` in the `serverless.yml` inputs for `@sls-next/serverless-component`


iiroj:
Ping @dphang I replaced https://github.com/serverless-nextjs/serverless-next.js/pull/2124 with this.

jonahallibone:
Is this a no-go? I'd really love to use the new output format

iiroj:
@jonahallibone I haven't had time to write really exhaustive tests... In the meantime, it seems Next.js is already working on a way to do the output themselves: https://nextjs.org/docs/advanced-features/output-file-tracing#automatically-copying-traced-files-experimental

jest.config.js
```diff
@@ -21,6 +21,8 @@ module.exports = {
     "/e2e-tests/",
     "/tests/",
     "/scripts/",
+    ".test_sls_next_output",
+    "__tests__",
```
iiroj, line 25: These were included in the coverage. Shouldn't be too important since they're not available in CI, but they do mess up local coverage.

packages/libs/core/src/build/builder.ts
```diff
@@ -57,7 +57,12 @@ export default abstract class CoreBuilder {
       this.buildOptions.nextStaticDir ?? this.buildOptions.nextConfigDir
     );
     this.dotNextDir = path.join(this.nextConfigDir, ".next");
-    this.serverlessDir = path.join(this.dotNextDir, "serverless");
+    this.nextTargetDir = path.join(
+      this.dotNextDir,
+      buildOptions && buildOptions.outputFileTracing
+        ? "server"
+        : "serverless"
+    );
```
iiroj, line 65: Without the `target` option, it's implied to emit into `.next/server/`.

packages/libs/lambda-at-edge/src/lib/copyOutputFileTraces.ts
```diff
@@ -0,0 +1,66 @@
...
+  pages
+}: {
+  serverlessDir: string;
+  destination: string;
+  pages: string[];
+}): Promise<void> => {
+  const DOT_NEXT = path.join(serverlessDir, "../");
+  const NEXT_SERVER_JS_NFT_JSON = path.join(
+    DOT_NEXT,
+    "next-server.js.nft.json"
```
iiroj, line 20: This file has some generic default stuff in it, and should always be included.

packages/libs/lambda-at-edge/src/lib/copyOutputFileTraces.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  const nftJsonFiles = [NEXT_SERVER_JS_NFT_JSON].concat(
+    pages.map((f) => `${f}.nft.json`)
+  );
```
iiroj, line 25: Every built "page" has its own nft file.

packages/libs/lambda-at-edge/src/lib/copyOutputFileTraces.ts
```diff
@@ -0,0 +1,66 @@
...
+      return Promise.reject(
+        `Failed to read trace \`${nft}\`. Check that you're using the \`outputFileTracing\` option with Node.js 12.`
+      );
+    }
+  };
+
+  await Promise.all(nftJsonFiles.map((file) => readNft(file)));
+
+  await Promise.all(
+    Array.from(traces)
```
iiroj, line 47: After every file from the nft json is added to the `traces` set, we turn it back into an array. This way there are no duplicates!

packages/libs/lambda-at-edge/src/lib/copyOutputFileTraces.ts
```diff
@@ -0,0 +1,66 @@
...
+      .filter((file) => !file.endsWith("package.json"))
```
iiroj, line 48: There seem to be a lot of `package.json` files, I don't think we need those.

packages/libs/lambda-at-edge/src/lib/copyOutputFileTraces.ts
```diff
@@ -0,0 +1,66 @@
...
+            ? normalized
+            : path.relative(serverlessDir, src)
+        );
+
+        /**
+         * When Next.js is used with TypeScript files, the file trace
+         * will include source `.ts`/`.tsx` files. Filter these out
+         * because they're already bundled, and because `src === dest`.
+         */
+        return src === dest ? [] : [fse.copy(src, dest)];
```
iiroj, line 63: This is roughly the same logic as in the previous implementation, but I was thinking it would make more sense to generically check that only target paths inside the output directory are included.

iiroj: Returning an empty array from `Array.prototype.flatMap` is the same as filtering it out.

packages/libs/lambda-at-edge/src/lib/copyRequiredServerFiles.ts
```diff
@@ -0,0 +1,39 @@
...
+  nextConfigDir,
+  destination
+}: {
+  nextConfigDir: string;
+  destination: string;
+}): Promise<void> => {
+  const REQUIRED_SERVER_FILES = path.join(
+    nextConfigDir,
+    ".next/required-server-files.json"
+  );
```
iiroj, line 14: Not really sure what the point of this file is, and why it's not an `.nft.json`, but seems to have required content anyway.

Purii:
Awesome! We absolutely need this feature to include localization files.
Is there anything I can support with or is this PR only waiting for approval + merge by @dphang ?

iiroj:
@Purii I do want to write more tests because otherwise the codecoverage check doesn't pass, but it does seem to work at least for my small site!

packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -218,6 +218,11 @@ class NextjsComponent extends Component {
         ? nextConfigPath
         : resolve(inputs.build.baseDir);
 
+    const outputFileTracing =
+      typeof inputs.build !== "boolean" &&
+      typeof inputs.build !== "undefined" &&
+      !!inputs.build.outputFileTracing;
```
dphang, line 224: probably can cover it in custom inputs test? or if it's not too convenient, we could modify the existing e2e tests to cover this

dphang:
thanks! I think it looks fine for the most part, codecov drop is just 0.08%, I gave a minor comment, I can probably add a new app to test this later so we get coverage there

dphang (approved):

packages/serverless-components/nextjs-component/src/component.ts
iiroj: I added a custom inputs test and also updated the `next-app-experimental` to use this flag.

dphang (approved):

iiroj:
@dphang I'm not familiar with the e2e test setup; would it make sense to run those before merging?

dphang:
> @dphang I'm not familiar with the e2e test setup; would it make sense to run those before merging?

Ran it here https://github.com/serverless-nextjs/serverless-next.js/actions/runs/1744128397 (there are some issues with the experimental app which I will fix later but if it passes on the other apps it should be fine)

iiroj:
@dphang seems like another e2e app failed, how to debug? I'll try to run this branch locally against my own site if it catches any errors.

dphang:
Ya that app is because somehow the IAM policy got corrupted, I need to manually fix it. But it should be fine since actually your changes is primarily build-time (the apps tests are to sanity check it doesn't break existing things) and experimental now so we can add more e2e tests later.

iiroj:
@dphang  did the tests fail because of this branch? Any way I can help debug them?

https://github.com/serverless-nextjs/serverless-next.js/actions/runs/1755195463

dphang:
Hm I think a new minor version of nextjs yesterday is causing failures since we pick the latest one, need to look into logs and fix it but your  changes hadn't failed earlier when I ran tests. I will check why it is failing, they may have introduced some new required files.

PR #2178, opened 2021-12-04
🚧 Add bundledLambdas option for bundling them with esbuild

iiroj:
This PR adds a new `inputs.build.bundledLambdas` option:

| Name            | Type     | Default Value | Description                                                                                                                                                                                                                                                                                                                                                                      |
| --------------- | -------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| build.bundledLambdas | `boolean` | `false` | **Experimental** Set this to true to use [esbuild](https://esbuild.github.io) to minify and bundle your Lambdas into a single file. This greatly decreases the Lambda file size and speeds up cold-boot time. |


iiroj:
What do you think about this, @dphang? I didn't write any unit tests yet because I wanted to get some input first on the idea itself.

dphang:
Seems promising, but hm, does this fully work with the `serverless` target? I thought one problem was the generated server-side chunks are dynamically imported by the corresponding pages and I think esbuild can't handle those properly?

iiroj:
Thanks @dphang, I will investigate. If it doesn't work we need to keep the chunks around.

iiroj:
@dphang you seem to be correct, and building the integration test directories with chunks does imply the  bundled Lambda would be broken. This is relevant: https://github.com/evanw/esbuild/pull/1155.

Let me close this PR for now...

PR #2262, opened 2022-01-13
feat: Handle Redirects that are rendered after ISR is made #2054

### LINKED ISSUE #2054: ISR Regeneration broken when a former 200 page results in a Redirect on later regeneration
### Issue Summary
We have a Page that moved and now results in a redirect. The Redirect isn't resolved

### Actual behavior

The response is 200 and the Redirect JSON is shown.
image

### Expected behavior

Redirect is resolved.

### Steps to reproduce

1. Have a page that is 200 Status code and let it Render to S3
2. Move that page somehow (in our case it's a change to the DB)
3. Wait for ISR of that page
4. JSON is shown instead of a redirect happening

### Versions

- OS/Environment: Cloudfront / Lambda@Edge
- @sls-next/serverless-component version: 3.5.0-alpha.8
- Next.js version: Newest 11

cmaerz:
Should fix #2054 please review, if my idea is the right direction

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -54,6 +54,8 @@ import { createRedirectResponse } from "@sls-next/core/dist/module/route/redirec
 import { redirect } from "@sls-next/core/dist/module/handle/redirect";
 import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
 import getStream from "get-stream";
+import { isJson } from "@sls-next/core";
+import { s3RemovePage } from "./s3/s3RemovePage";
 
 const basePath = RoutesManifestJson.basePath;
 
@@ -108,6 +110,7 @@ export const handler = async (
   const { now, log } = perfLogger(manifest.logLambdaExecutionTimes);
 
   const tHandlerBegin = now();
+  console.log("we are in v1 handler");
 
   if (isOriginResponse(event)) {
     response = await handleOriginResponse({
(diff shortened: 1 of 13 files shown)


cmaerz:
@dphang Any updates here? We really need this for our project :/

dphang:
Yes I'll look at it this week - was out for a trip. Hopefully we get more maintainers as well soon :)

packages/libs/lambda-at-edge/src/default-handler.ts
```diff
@@ -108,6 +110,7 @@ export const handler = async (
   const { now, log } = perfLogger(manifest.logLambdaExecutionTimes);
 
   const tHandlerBegin = now();
+  console.log("we are in v1 handler");
```
dphang, line 113: Remove, seems for debugging purposes? (yes the platform client only applies to v2 handlers which is used for lambda and the generic handlers)

packages/serverless-components/nextjs-component/src/component.ts
```diff
@@ -524,7 +524,7 @@ class NextjsComponent extends Component {
         {
           Effect: "Allow",
           Resource: `arn:aws:s3:::${bucketOutputs.name}/*`,
-          Action: ["s3:GetObject", "s3:PutObject"]
+          Action: ["s3:GetObject", "s3:PutObject", "s3:DeleteObject"]
```
dphang, line 527: I think delete object permissions should be added to the nextjs-cdk-construct too

dphang:
Ran e2e tests here: https://github.com/serverless-nextjs/serverless-next.js/actions/runs/1721272630. Gave some comments and I think you could add a little more test coverage but otherwise no major issues

packages/libs/lambda-at-edge/src/default-handler.ts
cmaerz: Well the story behind this is longer.. i wrote in my test config file `useV2Handlers` instead `useV2Handler` and that bugged me for about an hour :D I'll remove it...

dphang (approved):

PR #2265, opened 2022-01-17
Use basepath in s3 assets

DoNormal:
I ran into the problem described in #1620: when using a `basePath` in the Next.js application, the S3 asset bucket remains empty. There was a fix mentioned in the issue, but no PR yet, so I've taken the liberty to make that change and open this PR. I tried the change locally and it worked.

@dphang or any other maintainer: I'm of course willing to add tests, but I couldn't figure out how to integrate such a basePath test with the existing CDK tests.

### DIFF
--- a/packages/serverless-components/nextjs-cdk-construct/src/index.ts
+++ b/packages/serverless-components/nextjs-cdk-construct/src/index.ts
@@ -406,7 +406,9 @@ export class NextJSLambdaEdge extends Construct {
     );
 
     const assetsDirectory = path.join(props.serverlessBuildOutDir, "assets");
-    const assets = readAssetsDirectory({ assetsDirectory });
+    const { basePath } = this.routesManifest || {};
+    const normalizedBasePath = basePath && basePath.length > 0 ? basePath.slice(1) : "";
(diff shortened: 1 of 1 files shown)


dphang:
I think this is fine as it's a small enough change. Thx!

dphang (approved):

moshfrid:
i dont know why, but using this new alpha version seems to break isr. 
image

dphang:
Hm I think from the stacktrace it looks like the error is coming from the
Next.js chunks? can you actually find the code where it's failing and see
if it's coming from the library? maybe at
/var/task/chunks/595.js:417276:13? I didn't change anything in the ISR code
but seems like might be some Next.js version issue..maybe there's some
compatibility issue with the latest version?

moshfrid:
To be honest Im very confused, here's the function that's failing
image

DoNormal:
I've had this error as well with Next, even before this change, so my guess is that it's not serverless-nextjs' fault. I haven't yet investigated what could be the cause.

moshfrid:
@DoNormal I'm not having any issues using the alpha.4 tho

PR #2285, opened 2022-01-21
feat(cloudfront): add ability to set ResponseHeadersPolicyId

chrisneal:
I have added the ability to set the ResponseHeadersPolicyId on the Cloudfront behaviours.

I had to update some snapshots to get the tests to pass, though I get failures when Jest gets to the `Running coverage on untested files...` section.

I haven't had a chance to run an end-to-end test with it yet though.


dphang:
Thx! Let me run E2E tests but I think it should be good

dphang (approved):

PR #2323, opened 2022-01-29
fix: copy page traces when using build.outputFileTracing

iiroj:
In https://github.com/serverless-nextjs/serverless-next.js/pull/2169 with the new `inputs.build.outputFileTracing` option we copy all files included in the`*.nft.json` trace files, as well as those from `required-server-files.json`. I noticed the latter get accidentally copied to `<lambda>/.next/` instead  of the root, and from there did a bit of grepping and concluded the files aren't actually necessary.

Not sure what they are supposed be used for, but this PR removes them anyhow. If we find a case where `inputs.build.outputFileTracing` fails because they're missing, maybe we can restore them but for now IMO it's best to leave them out.

Thoughts @dphang?


iiroj:
I'll re-use this PR to report a finding: it seems when bundling React, the `nft.json` files do not include the `react/jsx-runtime` dependency anywhere, and this will lead to a 503 error in the deployed lambda:

```
{
  "errorType": "Error",
  "errorMessage": "Cannot find module 'react/jsx-runtime'\nRequire stack:\n- /var/task/pages/index.js\n- /var/task/default-handler-v2-536898eb.js\n- /var/task/index.js\n- /var/runtime/UserFunction.js\n- /var/runtime/index.js",
  "trace": [
    "Error: Cannot find module 'react/jsx-runtime'",
    "Require stack:",
    "- /var/task/pages/index.js",
    "- /var/task/default-handler-v2-536898eb.js",
    "- /var/task/index.js",
    "- /var/runtime/UserFunction.js",
    "- /var/runtime/index.js",
    "    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)",
    "    at Function.Module._load (internal/modules/cjs/loader.js:746:27)",
    "    at Module.require (internal/modules/cjs/loader.js:974:19)",
    "    at require (internal/modules/cjs/helpers.js:93:18)",
    "    at Object.997 (/var/task/pages/index.js:227:18)",
    "    at __webpack_require__ (/var/task/webpack-runtime.js:25:42)",
    "    at Object.457 (/var/task/pages/index.js:21:20)",
    "    at __webpack_require__ (/var/task/webpack-runtime.js:25:42)",
    "    at __webpack_exec__ (/var/task/pages/index.js:244:39)",
    "    at /var/task/pages/index.js:245:78"
  ]
}
```

iiroj:
Ok, so it looks like the generated `nft.json` files do not include any of the dependencies used by the pages, for example `styled-components`. Looks like we still need to use `@vercel/nft` in addition to them.

iiroj:
After copying all files it's failing to:

```
{
  "errorType": "TypeError",
  "errorMessage": "e.render is not a function",
  "trace": [
    "TypeError: e.render is not a function",
    "    at Se (/var/task/default-handler-v2-536898eb.js:1:21691)",
    "    at Ee (/var/task/default-handler-v2-536898eb.js:1:22137)",
    "    at Te (/var/task/default-handler-v2-536898eb.js:1:22345)",
    "    at async Ba (/var/task/default-handler-v2-536898eb.js:1:345152)",
    "    at async Runtime.exports.handler (/var/task/default-handler-v2-536898eb.js:1:613959)"
  ]
}

iiroj:
Looks like the implementation relies on this being present:
https://github.com/vercel/next.js/tree/canary/packages/next/build/webpack/loaders/next-serverless-loader

iiroj:
@dphang I now realize it's going to take some time to get the handler updated to support this feature...

I need to bundle the Next.js `BaseServer`, construct it from `required-server-files.json` (which contains the next app config), and then use the `server.getRequestHandler` so that I can finally replace the `page.render(req, res)` comparative to the current handler.

Let me open a PR to revert https://github.com/serverless-nextjs/serverless-next.js/commit/e01571fa347cfb71741eb092482b69dc5b54eb52 in the meantime...

PR #2324, opened 2022-01-29
Revert "Experimental Next.js 12 Output File Tracing (#2169)"

iiroj:
This reverts commit e01571fa347cfb71741eb092482b69dc5b54eb52.

See https://github.com/serverless-nextjs/serverless-next.js/pull/2323#issuecomment-1024973847, it looks like taking advantage of the output file tracing won't be as trivial as I thought, so let's revert the current implementation to not have it block other updates to this fantastic library!

Sorry about the hassle; I'll be back with a better-working solution.


iiroj:
@dphang for your information.

dphang (approved):

PR #2325, opened 2022-01-31
CDK: Allow to whitelist headers

Negan1911:
Hello, first of all thanks for this great library, and thanks for porting it to CDK.

Currently, I'm working on a Cloudfront app that has the domain as `*.mypage.com`

Where `*` is a brand selected by our customers, identifying it has become difficult for a couple of reasons:
- I don't know how to access the original event sent to the lambda (I also don't know how I can implement this one) so I went to the 2nd option which is reading the `Host` header.
- When reading the `Host` header on SSR, it works quite well, except when the request is to `_next/data`, in that case, the `Host` header is from CloudFront because it's not being forwarded, this PR adds the ability to forward headers.

I've tried to keep it up with what was needed, it was being tested by my end and I added a unit test, please let me know if I'm missing something, also, if you have some clues on how to read the original event on nextjs `getServerSideProps`. I would love to implement that on another PR.

Again, thank you!

### DIFF
--- a/documentation/docs/cdkconstruct.md
+++ b/documentation/docs/cdkconstruct.md
@@ -92,6 +92,8 @@ new NextJSLambdaEdge(this, "NextJsApp", {
 - `s3Props?: Partial<BucketProps>` - pass custom s3 props
 - `whiteListedCookies?: string[]` - provide a list of cookies to forward to the
   CloudFront origin.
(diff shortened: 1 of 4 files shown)


dphang (approved):
Thanks, it LGTM

PR #2328, opened 2022-01-31
Update cdktf and use prebuilt providers

DanielMSchmidt:
Hey, thanks for this awesome project, I'm really keen to try it out!

- Updates cdktf to 0.9 and adds every used provider as prebuilt provider `@cdktf/provider-...`
- Moves from `TerraformStack` to `Construct` for base class so that the class can be used within a Terraform Stack
- Adds keyword so this appears on constructs.dev

Feel free to let me know if I should drop any of the commits :)


dphang:
Sorry i didn't see it for a while. This looks great! I've been meaning to complete the reference cdktf implementation for lambda, so does it work better out of the box now?

dphang (approved):

DanielMSchmidt:
> So does it work better out of the box now? 

I haven't tested it to be honest, I just looked at the code and made the adjustments and checked it synthed. What works better now is dependency management, if someone already uses aws and the prebuilt provider they have one version installed, so the one used here is not conflicting e.g. in the provider version. Plus you can install multiple apps in one installation :)

PR #2333, opened 2022-02-05
🚧 feat: add `build.outputStandalone` option

iiroj:
This is an experimental Work-In-Progress PR for adding support for the new `experimental.outputStandalone` setting introduced in Next.js 12 https://github.com/vercel/next.js/pull/31003

There's a new minimal `next-handler` Lambda that currently doesn't support any of the existing `serverless-nextjs` features.

The build process is also very simple:

1. build Next.js with ` experimental.outputStandalone: true`
2. copy entire `.next/standalone` as the `default-lambda`
3. remove the included `.next/standalone/server.js`
4. emit new handler, which creates a Next.js `NextServer` and handles the lambda request

What do you think @dphang ? How should we integrate this into the existing stack? Should we make a completely new lambda, or maybe a setting into the `default-handler` that uses this new file instead of importing pages and using their `render` method?


packages/libs/lambda-at-edge/src/next-handler.ts
```diff
@@ -0,0 +1,41 @@
...
+    dir: join(__dirname),
+    // @ts-ignore
+    minimalMode: true
+  });
+
+  const requestHandler = nextServer.getRequestHandler();
+  await requestHandler(req, res);
+
+  return responsePromise;
+};
```
iiroj, line 41: Using the standalone Next.js export makes the lambda logic pretty simple. This is intentionally missing all the existing features because I'm not yet sure what the best to use this would be...

dphang:
> This is an experimental Work-In-Progress PR for adding support for the new `experimental.outputStandalone` setting introduced in Next.js 12 [vercel/next.js#31003](https://github.com/vercel/next.js/pull/31003)
> 
> There's a new minimal `next-handler` Lambda that currently doesn't support any of the existing `serverless-nextjs` features.
> 
> The build process is also very simple:
> 
> 1. build Next.js with ` experimental.outputStandalone: true`
> 2. copy entire `.next/standalone` as the `default-lambda`
> 3. remove the included `.next/standalone/server.js`
> 4. emit new handler, which creates a Next.js `NextServer` and handles the lambda request
> 
> What do you think @dphang ? How should we integrate this into the existing stack? Should we make a completely new lambda, or maybe a setting into the `default-handler` that uses this new file instead of importing pages and using their `render` method?

Maybe it can be a new lambda if it’s different? Not too familiar with the new handler yet, haven’t kept up with all new features unfortunately…

jinqian1998:
Looking forward to this feature, thanks to @iiroj works!
For now, Tim is removing `target: 'serverless'` in nextJS, so it's better we have another way deploying it!

nike1v:
> Looking forward to this feature, thanks to @iiroj works! For now, Tim is removing `target: 'serverless'` in nextJS, so it's better we have another way deploying it!

Hi!
Using the code provided here you managed to deploy the NextJS app using standalone output?

nike1v:
HI @iiroj !
Are you using this solution right now? Or you found a way to deploy Next 12/13 in other way?
Thanks.

iiroj:
Hello @nike1v , we have abandoned this effort and went with Vercel... would look into OpenNext maybe: https://open-next.js.org/

PR #2393, opened 2022-03-04
fix redirect in getStaticProps

### LINKED ISSUE #2346: Redirects not working in getStaticProps
### Issue Summary
Redirects are throwing errors in getStaticProps. I know this is not an issue with the code or anything related to the next js because I have checked the local build and it successfully redirects to the targeted page. 

### Actual behavior
While redirecting it throws an error that the client-side exception occurred.  I am unable to explore more about the error. Actual redirection doesn't occur.
image

### Expected behavior
getStaticProps should redirect to the destination path.

### Steps to reproduce
- Add redirect to getStaticProps with some external or internal URL. 
- Go to the path that should redirect to the targeted URL
- Client-side exception occurs

### Versions

- OS/Environment: Mac
- @sls-next/serverless-component version: 3.2.0
- Next.js version: ^11.0.1

### Checklist

- [x] You have reviewed the [README](https://github.com/serverless-nextjs/serverless-next.js/blob/master/README.md) and [FAQs](https://github.com/serverless-nextjs/serverless-next.js#faq), which answers several common questions.
- [x] You have reviewed our [DEBUGGING](https://github.com/serverless-nextjs/serverless-next.js/wiki/Debugging-Issues) wiki and have tried your best to include complete information and reproduction steps (including your configuration) as is possible. As there is only one maintainer (who maintains this in his free time) and thus very limited resources, if you have time, please try to debug the issue a bit yourself if possible.
- [x] You have first tried using the most recent `latest` or `alpha` `@sls-next/serverless-component` [release version](https://github.com/serverless-nextjs/serverless-next.js/releases), which may have already fixed your issue or implemented the feature you are trying to use. Note that the old `serverless-next.js` component and the `serverless-next.js` plugin are deprecated and no longer maintained.

KensukeTakahara:
This fixes #2346.

Added redirect to getStaticProps as shown in the code below.

```tsx
import type { NextPage, GetStaticProps } from "next";
import { useRouter } from "next/router";

export const getStaticProps: GetStaticProps = async () => {
  const data = await fetch("https://github.com/serverless-nextjs/serverless-next.js").then((res) =>
    res.text()
  );

  if (data.length > 0) {
    return {
      redirect: {
        permanent: false,
        destination: "/bar",
      },
    };
  }

  return {
    props: {},
  };
};

export async function getStaticPaths() {
  return {
    paths: [],
    fallback: true,
  };
}

const Foo: NextPage = () => {
  const router = useRouter();
  const { id } = router.query;

  return <p>Post: {id}</p>;
};

export default Foo;

```

And I deployed it to vercel, and the following response was returned.
```
Request URL: https://xxx.vercel.app/_next/data/xxx/foo/1.json
Request Method: GET
Status Code: 200 OK
{
  "pageProps": {
    "__N_REDIRECT": "/bar",
    "__N_REDIRECT_STATUS": 307
  },
  "__N_SSG": true
}
```

Therefore, I stopped returning 307 and return 200 & json body.

### DIFF
--- a/packages/libs/core/src/handle/redirect.ts
+++ b/packages/libs/core/src/handle/redirect.ts
@@ -8,20 +8,20 @@ export const redirect = (event: Event, route: RedirectRoute) => {
   event.res.end();
 };
 
-export const redirectSsg = (event: Event, route: RedirectRoute) => {
+export const redirectByPageProps = (event: Event, route: RedirectRoute) => {
   event.res.setHeader(
     "cache-control",
     route.headers?.cacheControl?.join(":") ?? ""
   );
   event.res.statusCode = 200;
-  event.res.write(
-    JSON.stringify({
-      pageProps: {
-        __N_REDIRECT: route.headers?.location[0].value ?? "",
-        __N_REDIRECT_STATUS: route.status
-      },
-      __N_SSG: true
-    })
-  );
+
+  const body = {
+    pageProps: {
+      __N_REDIRECT: route.headers?.location[0].value ?? "",
+      __N_REDIRECT_STATUS: route.status
+    },
+    __N_SSG: true
+  };
+  event.res.write(JSON.stringify(body));
   event.res.end();
 };
--- a/packages/libs/core/src/handle/redirect.ts
+++ b/packages/libs/core/src/handle/redirect.ts
@@ -7,3 +7,21 @@ export const redirect = (event: Event, route: RedirectRoute) => {
   event.res.statusMessage = route.statusDescription;
   event.res.end();
 };
+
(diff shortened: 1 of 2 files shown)


dphang:
thx, I think you can rebase since the lockfile is fixed now.

dphang (approved):

dphang:
Hm, actually seems this might have broke one test in https://github.com/serverless-nextjs/serverless-next.js/runs/5485904554?check_suite_focus=true, maybe you can check it?

cmaerz:
@dphang @KensukeTakahara  This breaks my App. I just get a JSON Response instead of a redirect. (when i directly access a page)
image

PR #2438, opened 2022-05-11
feat(nextjs-cdk-construct): add optional prop to override API Lambda role

dvejmz:
Add an optional prop to the `NextJsLambdaEdge` CDK construct to override the IAM role that is attached to it. This can be useful as sometimes it's desirable to give the Nextjs API backend the ability to interact with other AWS resources such as DynamoDB.


dvejmz:
Closing as there are severe limitations with this approach which I want to address I will raise a new PR once I've redesigned the role override approach to make it more robust.

Issue #2437, opened 2022-05-11
Rewrites don't work with special symbols in URL


sleukhin:
### Issue Summary
Rewrites don't work with special symbols in URL

1. no-op rewrites don't work with paths that include symbols like `@` and `=`
2. external rewrites will build incorrect destination routes for paths that include symbols like `@` and `=`

### No-op rewrites example:
**rewrite config:**
```
{
  source: '/:path*',
  destination: '/:path*',
},
{
  source: "/:path*",
  destination: `https://external.app/:path*`
}
```

**PageManifest:**
```
pageManifest = {
  pages: {
    dynamic: [],
    html: {
      dynamic: {},
      nonDynamic: {}
    },
    ssg: {
      dynamic: {},
      nonDynamic: {},
      notFound: {}
    },
    ssr: {
      dynamic: {},
      nonDynamic: {}
    }
  }
};
```

#### Actual behavior
input URL: `/@iamatest`
rewrite result: `/%40iamatest`

input URL: `/iamatest=`
rewrite result: `/iamatest%3D`

#### Expected behavior
according to how Next.js server rewrites work

input URL: `/@iamatest`
rewrite result: `https://external.app/@iamatest`

input URL: `/iamatest=`
rewrite result: `https://external.app/iamatest=`

### External rewrites example:
**rewrite config:**
```
{
  source: "/:path*",
  destination: `https://external.app/:path*`
}
```

#### Actual behavior
input URL: `/@iamatest`
rewrite result: `https://external.app/%40iamatest`

input URL: `/iamatest=`
rewrite result: `https://external.app/iamatest%3D`

#### Expected behavior
according to how Next.js server rewrites work

input URL: `/@iamatest`
rewrite result: `https://external.app/@iamatest`

input URL: `/iamatest=`
rewrite result: `https://external.app/iamatest=`

Results from above show that rewrites create incorrect URLs when symbols like `@` and `=` are used.

In my case, this causes an additional problem because the `external.app` redirects to the correct URL and I get an infinite loop of redirects:
- a user requests `https://next.js.app/@iamatest`
- next.js.app rewrites to `https://external.app/%40iamatest`
- external.app redirects to `https://next.js.app/@iamatest`
- ... repeat ...

### Versions
- @sls-next/serverless-component version: 3.7.0
- Next.js version: 11.1.4

### Checklist

[...]

PR #2440, opened 2022-05-13
fix: [Issue-2437] Rewrites should support special symbols in dynamic routes

savvaleukhin:
A proposal for fixing «Rewrites don't work with special symbols in a URL»
[Issue reference.](https://github.com/serverless-nextjs/serverless-next.js/issues/2437)

Summary of the issue: 
1. no-op rewrites don't work with paths that include symbols like `@` and `=`
2. external rewrites will build incorrect destination routes for paths that include symbols like `@` and `=`
all of the above works in the Next.js server

I followed the code and found no reason for using encoders in the rewrite mechanism.
the [next code](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/libs/core/src/route/rewrite.ts#L26):
```
const match = matchPath(path, rewrite.source);
if (!match) {
  continue;
}
```
should prevent wrong path matching from the beginning.

Also, the [`compile` function](https://github.com/serverless-nextjs/serverless-next.js/compare/master...sleukhin:rewrites-with-special-symbols-in-dynamic-routes?expand=1#diff-b7d18047cd9bd1f67c726c2ff914a4e2ce5f543533d8b7a9653e124c067cb7f9R35) should not produce an invalid path since `validation` is enabled by default

@dphang  Am I missing something?

### DIFF
--- a/packages/libs/core/src/match.ts
+++ b/packages/libs/core/src/match.ts
@@ -32,7 +32,7 @@ export function compileDestination(
     ) {
       // Handle external URL redirects
       const { origin, pathname, search } = new URL(destination);
-      const toPath = compile(pathname);
+      const toPath = compile(pathname, { encode: encodeURI });
       const compiledDestination = `${origin}${toPath(params)}${search}`;
 
       // Remove trailing slash if original destination didn't have it
@@ -44,7 +44,9 @@ export function compileDestination(
     } else {
       // Handle all other paths. Escape all ? in case of query parameters
       const escapedDestination = destination.replace(/\?/g, "\\?");
-      const toPath = compile(escapedDestination);
+      const toPath = compile(escapedDestination, {
+        encode: encodeURI
+      });
       return toPath(params);
     }
   } catch (error) {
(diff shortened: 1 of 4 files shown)


packages/libs/core/src/match.ts
```diff
@@ -32,7 +32,7 @@ export function compileDestination(
     ) {
       // Handle external URL redirects
       const { origin, pathname, search } = new URL(destination);
-      const toPath = compile(pathname, { encode: encodeURIComponent });
+      const toPath = compile(pathname);
```
lucasbaze, line 35: **Question:** Instead of removing the encoding entirely, would it make sense to use `encodeURI`? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

sleukhin: Hi, thanks for the suggestion.

I've checked how the next.js server works with path matching and the difference is quite significant.
However, the usage of `encodeURI` makes the behavior a little bit closer.
It will encode not-ASCII chars but will not encode symbols like @=.

Issue #2477, opened 2022-08-13
Cannot build with middleware and target: serverless


revmischa:
I don't know if this is a serverless-next issue but this warning has become an error for me in the latest nextjs

```
> Build error occurred
MiddlewareInServerlessTargetError: Next.js Middleware is not supported in the deprecated serverless target.
Please remove `target: "serverless" from your next.config.js to use Middleware.
```

And my next.config.mjs is empty

_Originally posted by @revmischa in https://github.com/serverless-nextjs/serverless-next.js/issues/2329#issuecomment-1213421647_

johnymanuelli:
Bump.

rkingonnw:
+1

revmischa:
I'm working on a new setup that will solve this: https://github.com/jetbridge/cdk-nextjs

PR #2478, opened 2022-08-13
Remove target config (#2477)

revmischa:
`target` is no longer a supported configuration option in nextjs.

If you try to use a recent version of Nextjs with middleware this becomes a build error: https://github.com/serverless-nextjs/serverless-next.js/issues/2477

Perhaps we should remove the `useServerlessTraceTarget` option or just [assume it is true](https://nextjs.org/docs/advanced-features/output-file-tracing)?

### DIFF
--- a/packages/libs/core/src/build/builder.ts
+++ b/packages/libs/core/src/build/builder.ts
@@ -105,8 +105,7 @@ export default abstract class CoreBuilder {
 
     const { restoreUserConfig } = await createServerlessConfig(
       cwd,
-      path.join(this.nextConfigDir),
-      false
+      path.join(this.nextConfigDir)
     );
 
     try {
--- a/packages/libs/core/src/build/lib/createServerlessConfig.ts
+++ b/packages/libs/core/src/build/lib/createServerlessConfig.ts
@@ -1,12 +1,11 @@
 import fs from "fs-extra";
 import path from "path";
 
-function getCustomData(importName: string, target: string): string {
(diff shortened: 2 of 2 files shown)


packages/libs/lambda-at-edge/src/build.ts
```diff
@@ -91,7 +90,7 @@ class Builder {
     this.nextConfigDir = path.resolve(nextConfigDir);
     this.nextStaticDir = path.resolve(nextStaticDir ?? nextConfigDir);
     this.dotNextDir = path.join(this.nextConfigDir, ".next");
-    this.serverlessDir = path.join(this.dotNextDir, "serverless");
+    this.serverlessDir = path.join(this.dotNextDir, "server");
```
revmischa, line 93: I think this changed

rntvicente:
Hi, thanks for this initiative. Do you happen to have a merge preview?

Currently I can't use Next12 middleware.

revmischa:
I don't know if these changes are worth pursuing for this module. I want to investigate the much simplified approach using the new `standalone` output option in NextJShttps://github.com/serverless-stack/sst/issues/1988 described here: https://github.com/vercel/next.js/discussions/33275#discussioncomment-3343424

I would also advise checking out https://github.com/sladg/nextjs-lambda

revmischa:
Anyone who wants to deploy NextJS on AWS should look at https://github.com/jetbridge/cdk-nextjs or https://docs.sst.dev/constructs/NextjsSite

PR #2486, opened 2022-09-01
Fix passing etag arg as last modified header, prefer last modified if present as message deduplication id

nskarakoc:
ISR enabled pages does not always get fresh content because of SQS message deduplication id set to ETag of the object (previously generated static page HTML file) stored in S3. Here is the scenario where `revalidate` option is set to `1`:

1. Open the page, see regeneration happened and page content updated in S3 (assuming previously generated content was different or 5-min elapsed)
2. Update page content (i.e. modify the DB records that are used to build the page)
3. Wait for `revalidate` seconds
4. Open the same page again, see regeneration is being triggered by the default Lambda, but is not picked by regeneration Lambda as SQS message deduplication id (ETag of page HTML file in S3) is exactly same as previous message. 

So in this case we have to wait for 5 minutes which is [SQS message deduplication interval](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagededuplicationid-property.html). This PR will attempt to use last modified header instead, so that even if the page content is same, it's last modified date will be different (i.e. newer) therefore regeneration will happen.

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -386,7 +386,7 @@ const handleOriginResponse = async ({
           request,
           pageS3Path: s3Uri,
           eTag: response.headers["etag"]?.[0].value,
-          lastModified: response.headers["etag"]?.[0].value,
+          lastModified: response.headers["last-modified"]?.[0].value,
           pagePath: staticRoute.page,
           queueName: regenerationQueueName
         });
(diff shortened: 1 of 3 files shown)


nskarakoc:
@dphang @J3tto could you please take a look at this PR? Thanks!

delisdeli (approved):

chrisneal:
@dphang @J3tto Have you had a chance to look into this?

PR #2495, opened 2022-10-14
fix: locale in homepage with query params

apollisa:
This is a PR to fix a weird bug on **localized** SSR root path, where adding query params to the root path redirected to the default locale.

Happy to have your thoughts & comments!


Misfits09:
Nice job, thanks !
That would help us a lot, we had to enable the `trailingSlash` in my project in order to make our homepage usable...

alexandrepernin (approved):

PR #2509, opened 2022-11-18
Do not forward cookies to S3 for public files

theodiablo:
# Why?
I am facing an issue with a Next.js site build with SST when requesting public files while accessing to the site with a lot of long cookies:
image

Here, getting a 400 error when requesting `/fonts/sofia-pro-bold.woff2` which is present in the `/public` directory in my site.

The issue seems to be that S3 has a hard limit of 8192 characters for the headers, and all the cookies are forwarded to S3 when requesting public files, which then makes the request explode in S3.

I have started a [discussion on the Discord #help channel](https://discord.com/channels/983865673656705025/983866416832864350/threads/1043132972162895892) to look for help on this.

# What?
Edit the Lambda@Edge handler to not forward cookies to S3 for public files.

# How to reproduce?
While accessing a Next.js website, in your browser create 3 or more cookies with 3000 characters (total should be more than 8192) and request a public resource. The request doesn't complete.

This is the simplest and safest solution I could come up with to not break anything else, happy to get other suggestions or proposals to solve this issue.

### DIFF
--- a/packages/libs/lambda-at-edge/src/default-handler.ts
+++ b/packages/libs/lambda-at-edge/src/default-handler.ts
@@ -253,6 +253,7 @@ const handleOriginRequest = async ({
 
   if (route.isPublicFile) {
     const { file } = route as PublicFileRoute;
+    request.headers["cookies"] = [];
     return await staticRequest(
       event,
       file,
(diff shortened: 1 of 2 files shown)


theodiablo:
@dphang, happy to get your feedback on this fix.
It would be nice to get this in ASAP to solve our problem.

